PR #1, opened 2018-09-08
Added a button component

Bogedal:


src/app/button/button.component.scss
```diff
@@ -0,0 +1,6 @@
+button {
+    border: solid 1px;
```
jakobe: color?

src/app/button/button.component.tns.html
```diff
@@ -0,0 +1 @@
+<Button text="button works!" class="btn btn-primary"></Button>
```
jakobe: Should bind label property to text attribute

src/app/button/button.component.scss
Bogedal: I went with default black border and made the whole button lightskyblue.

PR #15, opened 2018-09-25
Page structure native

neigaard:


package.json
```diff
@@ -33,6 +33,8 @@
     "@angular/router": "~6.1.0",
     "core-js": "^2.5.4",
     "nativescript-angular": "~6.1.0",
+    "nativescript-orientation": "^2.2.1",
+    "nativescript-platform-css": "^1.6.6",
```
jakobe: Do we need to check for platform? Or is it because that's the only way to get the screen size?

src/app/app-shell/home/home.component.tns.scss
```diff
@@ -0,0 +1,48 @@
+$prototype-color: rgb(0, 198, 224);
+$prototype-color-bg: rgb(232, 246, 255) ;
```
jakobe: `$prototype-color-bg` => `$prototype-bg-color` (it's easier if all color vars end with `-color`)

src/app/app-shell/home/home.component.tns.scss
```diff
@@ -0,0 +1,48 @@
...
+
+.main-container {
+  flex-direction: column-reverse;
```
jakobe: Why `flex-direction: column-reverse`?

src/app/app-shell/home/home.component.tns.scss
```diff
@@ -0,0 +1,48 @@
...
+
+.landscape.android740 :host .main-container {
+  flex-direction: row;
+}
+
+.menu-container {
+  flex-direction: row;
+  align-items: center;
+  justify-content: space-between;
+  height: 120 dp;
```
jakobe: Do we need to specify `dp`?

src/app/app.module.tns.ts
```diff
@@ -30,4 +32,9 @@ import {IntroComponent} from './intro/intro.component';
   bootstrap: [AppComponent],
   schemas: [NO_ERRORS_SCHEMA]
 })
-export class AppModule { }
+export class AppModule {
+  constructor() {
+    platformCss.sizeGroupings(true);
+    platformCss.sizeGroupings([9999,740,320]);
```
jakobe: Magic numbers ✨ 🎩  => put into named vars

package.json
neigaard: Yes with this plugin it is, we will make our own superduper plugin or possible just make a change request to that existing plugin.

src/app/app-shell/home/home.component.tns.scss
neigaard: We need to talk colors in general, I guess we will have a common set of colors? I stole these from Web-Ninja Tommyboy.

src/app/app-shell/home/home.component.tns.scss
neigaard: I think we do, but I will investigate :)

src/app/app.module.tns.ts
neigaard: Yes, this will change, only here as a demo and it is not at all working as we want.

src/app/app-shell/home/home.component.tns.scss
jakobe: This example doesn't: https://docs.nativescript.org/ui/styling#supported-measurment-units
`.myLabel {
    font-size: 28;
    width: 200;
    height: 30;
}`

src/app/app.module.tns.ts
jakobe: Ok check

src/app/app-shell/home/home.component.tns.scss
neigaard: I need to read about it before I beleive it :p

PR #17, opened 2018-10-08
Theming raogf 100

jb-trm:


src/scss/_common-functions.scss
```diff
@@ -9,4 +9,14 @@
 
 @function z($key) {
   @return _get-map-prop($z-layers, $key);
+}
+
+@function rems2dp($size) {
```
neigaard: I doubt it, rem is dependant on font size, we probably want it to be more a fixed size on native

src/scss/themes/_common-themes.scss
```diff
@@ -1,16 +1,34 @@
...
     --surface: rgb(64, 64, 64);
     --on-surface: rgba(255, 255, 255, 0.87);
   }
 }
-
-.domain--jb {
-  color: #fff;
-  
-}
+This wouldn't require the css to be build XX number of times.
```
neigaard: Add more commenting about this, make it very clear that this Tommy magic fixes a performance problem and that is why we do this and that isnt allowed...

PR #19, opened 2018-10-15
Theming raogf 100 v2

jb-trm:
Added colortype attribute to kirby-button - used to indicated which color the button should use from the theme.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
     "ios": "tns run ios --bundle --device 'iPhone X'",
     "pack-bump-patch": "cd src/kirby && npm version patch",
     "pack-bump-minor": "cd src/kirby && npm version minor",
-    "pack-copy-to-dist-folder": "cpx \"src\\kirby\\**\\!(*.spec.ts)\" dist-lib --watch",
+    "pack-copy-to-dist-folder": "cpx \"src\\kirby\\**\\!(*.spec.ts)\" dist-lib",
+    "pack-copy-to-dist-folder-watch": "cpx \"src\\kirby\\**\\!(*.spec.ts)\" dist-lib --watch",
     "pack-build": "npm run pack-copy-to-dist-folder",
     "pack-publish": "npm publish dist-lib",
     "pack-build-and-publish": "npm run pack-build && npm run pack-publish"
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "ios": "tns run ios --bundle --device 'iPhone X'",
     "pack-bump-patch": "cd src/kirby && npm version patch",
     "pack-bump-minor": "cd src/kirby && npm version minor",
-    "pack-copy-to-dist-folder": "cpx \"src\\kirby\\**\\!(*.spec.ts)\" dist-lib",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "ios": "tns run ios --bundle --device 'iPhone X'",
     "pack-bump-patch": "cd src/kirby && npm version patch",
     "pack-bump-minor": "cd src/kirby && npm version minor",
-    "pack-copy-to-dist-folder": "cpx \"src\\kirby\\**\\!(*.spec.ts)\" dist-lib",
+    "pack-copy-to-dist-folder": "cpx \"src\\kirby\\**\\!(*.spec.ts)\" dist-lib --watch",
```
jakobe: Should be separate script, e.g. `pack-copy-to-dist-folder-watch`

src/kirby/components/button/button.component.scss
```diff
@@ -13,18 +13,25 @@ button {
   background-color: var(--primary);
   border-color: var(--primary);
   border-radius: var(--kirby-button-border-radius);
-  border: 1px solid;
+  border-width: 1px;
```
jakobe: Why split on 2 lines?

src/kirby/components/button/button.component.scss
```diff
@@ -13,18 +13,25 @@ button {
...
+  border-style: solid;
   color: var(--on-primary);
```
jakobe: `--on-` is that a convention?

src/kirby/components/button/button.component.scss
```diff
@@ -13,18 +13,25 @@ button {
...
+  padding: 0.7rem 1.75rem;
   transition: opacity 200ms linear;
   width: 100%;
 
   &.round {
     --kirby-button-border-radius: 2rem;
   }
 
+  &.button--secondary {
+    background-color: var(--secondary);
```
jakobe: Maybe something more specific than just `--secondary`, to avoid clashing with other components/frameworks?
+ Add suffix (`-color`, `-background-color`, `-border-color`, `border-radius` etc)
E.g. `--kirby-secondary-color`

src/kirby/components/button/button.component.scss
jakobe: Maybe something more specific than just `--on-primary`, to avoid clashing with other components/frameworks?
+ Add suffix (`-color`, `-background-color`, `-border-color`, `border-radius` etc)
E.g. `--kirby-on-primary-color`

src/kirby/components/button/button.component.ts
```diff
@@ -17,11 +18,12 @@ export class ButtonComponent implements OnInit {
   constructor() { }
 
   ngOnInit() {
-    const { expand, shape, theme } = this;
+    const { expand, shape, theme, colortype } = this;
     this.buttonCssClasses = {
       'round': shape === 'round',
       'block': expand === 'block',
       'full': expand === 'full',
+      ['button--' + colortype]: colortype,
```
jakobe: What's with the double dashes? `--`?

PR #45, opened 2018-11-30
Image component

neigaard:
First draft of the image component. Size and image url is hardcoded and it should not be, but someone else will fix this.


src/kirby/components/image/image.component.tns.ts
```diff
@@ -0,0 +1,64 @@
...
+declare const CGSizeMake: any;
+declare const android: any;
+
+@Component({
+  selector: 'kirby-image',
+  templateUrl: './image.component.html',
+  styleUrls: ['./image.component.scss']
+})
+export class ImageComponent implements OnInit {
+  view: View;
```
kofoeddk: Mangler lige en @Input() width: number;

kofoeddk: Men det tænker jeg lige, at vi kigger på senere ;)

PR #61, opened 2018-12-13
Feature/vulnerability updates

langecode:
Updating dev dependencies with transitive dependencies which has vulnerabilities.


jonasjensen-brolstark:
@langecode is this PR still relevant or should it just be closed?

PR #77, opened 2019-01-22
Feature/card scaling

jakobe:
Added ResizeObserverService to track card scaling + implemented poly- and ponyfill 🐴 for all other browsers than Chrome


src/kirby/components/card/card.component.ts
```diff
@@ -8,10 +11,32 @@ import { Component, OnInit, Input } from '@angular/core';
 export class CardComponent implements OnInit {
   @Input() title: string;
   @Input() subtitle: string;
+  sizes = {
```
jonasjensen-brolstark: Is there a need for this to be public?

src/kirby/components/card/card.component.ts
```diff
@@ -8,10 +11,32 @@ import { Component, OnInit, Input } from '@angular/core';
...
+  }
 
   ngOnInit() {
+    this.resizeObserverService.observe(this.elementRef, (entry) => this.handleResize(entry));
+  }
+
+  private handleResize(entry: ResizeObserverEntry) {
+    Object.entries(this.sizes).forEach(([size, width]) => {
+      if (entry.contentRect.width >= width) {
+          // this.renderer.setAttribute(entry.target, size, '');
```
jonasjensen-brolstark: My internal linter tells me, that this code will never be run. :)
The same goes for the comment below.

src/kirby/components/shared/resize-observer/resize-observer.service.ts
```diff
@@ -0,0 +1,67 @@
...
+        this.zone.runOutsideAngular(() => {
+          this.observer.observe(element);
+        });
+      }
+      this.observedElements.set(element, action);
+    }
+    return element;
+  }
+
+  private handleResize(entries: ResizeObserverEntry[]) {
```
jonasjensen-brolstark: please follow the style guide [05-14](https://angular.io/guide/styleguide#style-05-14) on the member sequence.

src/kirby/components/shared/resize-observer/resize-observer.service.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  observe(elementOrRef: Element | ElementRef<Element>, action: (entry: ResizeObserverEntry) => void): Element {
+    if (this.observer === undefined) {
+      this.observer = this._resizeObserverFactory.create((entries) => this.handleResize(entries));
+    }
+
+    const element = elementOrRef instanceof ElementRef ? elementOrRef.nativeElement : elementOrRef;
+    if (!this.observedElements.has(element)) {
+      if (this.observer) {
+        // console.log(`Observing <${element.tagName}> through ResizeObserver...`);
```
jonasjensen-brolstark: Is this needed?

src/kirby/components/shared/resize-observer/resize-observer.service.ts
```diff
@@ -0,0 +1,67 @@
...
+      if (action) {
+        action(entry);
+      }
+    });
+  }
+
+  ngOnDestroy() {
+    if (this.observer) {
+        // IMPORTANT: Has to be run outside the Angular zone, for it to work with ResizeObserver polyfill:
+        this.zone.runOutsideAngular(() => {
```
jonasjensen-brolstark: The indentation is a bit off here.

src/polyfills/resize-observer-polyfill-loader.js
```diff
@@ -0,0 +1,39 @@
...
+      var script = document.createElement('script');
+      script.src = 'kirby/polyfills/resize-observer-polyfill.min.js';
+      script.onload = init;
+      document.head.appendChild(script);
+    }
+
+    function polyfillNeeded() {
+        return (typeof ResizeObserver === 'undefined') &&
+               (window.screen && window.screen.width >= 576);
+               //matchMedia('(min-width: 576px)').matches
```
jonasjensen-brolstark: Remove if not needed.

src/polyfills/resize-observer-polyfill-loader.js
```diff
@@ -0,0 +1,39 @@
...
+          return true;
+        }
+      
+        // include the 'heartz' as a way to have a non matching MQ to help terminate the join
+        // https://github.com/Modernizr/Modernizr/issues/1814
+        var query = ['(', prefixes.join('touch-enabled),('), 'heartz', ')'].join('');
+        return mq(query);
+      }
+
+    function init() {
```
jonasjensen-brolstark: Is the init function needed if we only do a console.log?

src/kirby/components/shared/resize-observer/resize-observer.service.ts
```diff
@@ -0,0 +1,67 @@
...
+  // IMPORTANT: Keeps track of the existing ResizeObserver so it can be reused for performance.
+  // See: https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/z6ienONUb5A/F5-VcUZtBAAJ
+  private observer: ResizeObserver | null;
+  private observedElements = new WeakMap<Element, (entry: ResizeObserverEntry) => void>();
+
+  constructor(private _resizeObserverFactory: ResizeObserverFactory, private zone: NgZone) {}
+
+  observe(elementOrRef: Element | ElementRef<Element>, action: (entry: ResizeObserverEntry) => void): Element {
+    if (this.observer === undefined) {
+      this.observer = this._resizeObserverFactory.create((entries) => this.handleResize(entries));
```
jonasjensen-brolstark: Could the assignment not be moved to the constructor. Then the observe function be changed to:
 
```suggestion
    const element = elementOrRef instanceof ElementRef ? elementOrRef.nativeElement : elementOrRef;

    if (!this.observer) {
      return element;
    }

    if (!this.observedElements.has(element)) {
      // IMPORTANT: Has to be run outside the Angular zone, for it to work with ResizeObserver polyfill:
      this.zone.runOutsideAngular(() => {
        this.observer.observe(element);
      });
      this.observedElements.set(element, action);
    }
    return element;
```

Would that not be a bit more readable?

src/kirby/components/card/card.component.ts
jakobe: True :) We need to talk about whether to use css classes or attributes... Or maybe it doesn't really matter - but let's decide together

src/kirby/components/shared/resize-observer/resize-observer.service.ts
jakobe: Not understood? To me the code does that already...

src/kirby/components/shared/resize-observer/resize-observer.service.ts
jakobe: Nope

src/kirby/components/shared/resize-observer/resize-observer.service.ts
jakobe: Not understood?

src/polyfills/resize-observer-polyfill-loader.js
jakobe: Done

src/polyfills/resize-observer-polyfill-loader.js
jakobe: Nope

src/kirby/components/shared/resize-observer/resize-observer.service.ts
jakobe: Yes, but then we create the `ResizeObserver` before it's actually needed (that is, when `observe` is called). All though that might be a bit theoretical since we only create the entire service if it's needed. I've changed it + removed the return value as we didn't use it anywhere.

byNikx: Cross-browser support is not very good for `ResizeObserver` even with polyfills, can we take another path?

jakobe: Not really sure what you mean by this? Have you looked at the polyfill implementation or are you basing your comment on caniuse.com alone? Have you compared it with the supported browserlist for DRB?

src/kirby/components/shared/resize-observer/resize-observer.service.ts
jonasjensen-brolstark: Your `private handleResize` functions comes before the public function `ngOnDestroy` and `unobserve`. 

If we should follow the style guid, private members should come after public members in the code :)

src/kirby/components/shared/resize-observer/resize-observer.service.ts
jonasjensen-brolstark: I am also unsure what i am talking about here. So just ignore.

src/kirby/components/card/card.component.ts
jonasjensen-brolstark: For me styling belongs in css classes and not attributes. Would that not also be be more "valid" HTML way?

Is there any pros with one over the other?

PR #99, opened 2019-02-07
Styled section header to fit the new cell layout.

jonasjensen-brolstark:
@PetarRing and @SilverGull would you look at this and check that i have not messed anything up in your cell implementation?


src/kirby/components/list/list.component.scss
```diff
@@ -5,19 +5,37 @@ ul {
   margin-top: 0;
   font-size: font-size('s');
   margin-block-end: 0;
+
   li {
-    background-color: white;
     background-color: var(--kirby-surface);
     border-bottom: 1px solid var(--kirby-background);
     padding: 16px;
```
PetarRing:
```suggestion
    padding:  size('m');
```

PetarRing: I think this was left out when Peter merged my branch onto his 😄

PetarRing: The hover color could use some transparancy, hence the thing I tried with rgba 😄

src/kirby/components/list/list.component.tns.scss
```diff
@@ -1,11 +1,8 @@
 @import "../../scss/native-imports";
-$list-item-height: 44;
-$list-item-padding: 16;
-$list-header-height: 38;
 .row {
   padding: 16 8 16 8;
   justify-content: space-between;
   flex-direction: row;
-  border-color: #d6d6d6;
+  border-color: #f5f5f5;
```
PetarRing: Shouldn't this be a theming color? 
```suggestion
  border-color: get-color('kirby-background');
```

src/kirby/components/list/list.component.tns.scss
```diff
@@ -1,11 +1,8 @@
 @import "../../scss/native-imports";
-$list-item-height: 44;
-$list-item-padding: 16;
-$list-header-height: 38;
 .row {
   padding: 16 8 16 8;
```
PetarRing: Should this be a theming size? 😄 I am not sure if you need to specify all padding or only need left/right and top..
```suggestion
  padding: size('gutter') size('gutter-small') size('gutter') size('gutter-small');
```

src/kirby/components/list/list.component.scss
jonasjensen-brolstark: Should that not be a themed variable if it should have transparancy? or a lighter version of `--kirby-secondary` ?

src/kirby/components/list/list.component.tns.scss
jonasjensen-brolstark: Thanks didn't know we had cutter sizes.

src/kirby/components/list/list.component.tns.scss
jonasjensen-brolstark: Awesome was looking for how to use variables in native.

PR #102, opened 2019-02-11
Enhancement/kirby list accesibility

PetarRing:


src/kirby/components/list/list.component.scss
```diff
@@ -38,4 +36,10 @@ ul {
       }
     }
   }
+}
+
+.clickable {
```
jonasjensen-brolstark: Can be moved up to line 20 under the `li` rule and add a `&`.
```suggestion
&.clickable {
...
}
```

I expect the class `clickable` is going to be used on anything else the `li` tag?

src/kirby/components/list/list.component.scss
```diff
@@ -35,11 +35,14 @@ ul {
         background-color: transparent;
       }
     }
+
+    &.clickable {
```
jonasjensen-brolstark: Maybe the class should also be renamed. Sorry missed that in the last review.

tsconfig.json
```diff
@@ -20,6 +20,10 @@
         "paths": {
             "~/*": [
                 "src/*"
+            ],
```
jonasjensen-brolstark: And you are sure on what you are doing here?

src/kirby/components/list/list.component.scss
PetarRing: We can do that. But this class is purely for web. If you think I will :)

PetarRing: *will rename it

src/kirby/components/list/list.component.html
```diff
@@ -11,7 +11,7 @@
         >
         </ng-container>
       </li>
-      <li class="row" *ngFor="let item of section.items" (click)="onItemClick(item)" >
+      <li class="row" [ngClass]="{'clickable': isSelectable}" *ngFor="let item of section.items" (click)="onItemClick(item)" tabindex="0">
```
jonasjensen-brolstark: Maybe we should use the [Aria button role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) on the `li` tags that are selectable.

src/kirby/components/list/list.component.scss
jonasjensen-brolstark: Do it 🚀 

But it is just to use the same  vocabulary in code and styling.

src/kirby/components/list/list.component.html
PetarRing: I belive I changed that and pushed it with a few other changes

PR #107, opened 2019-02-14
Upgrade to NativeScript 5.2.0 and Angular 7.2.0

MortenGregersen:
After upgrading NativeScript and Angular, Highcharts started failing when Kirby was used in a NativeScript app. This explains why the Highcharts upgrade is part of this pull request. With the new version of Highcharts some changes to the source was necessary.


src/kirby/components/chart/chart.component.ts
```diff
@@ -43,35 +43,50 @@ export class ChartComponent implements OnInit, OnChanges {
...
+            type: 'pie',
+            data: this.data as Array<Highcharts.SeriesPieDataOptions>
+          }];
+          break;
+        }
+        case ChartType.AREASPLINE: {
+          this.options.series = [{
+            type: 'areaspline',
+            data: this.data as Array<Highcharts.SeriesAreasplineDataOptions>
+          }];
```
askarby: Missing break-statement (for explicitly explain why it's required, in a comment)?

MortenGregersen: It was missing ;)

PR #108, opened 2019-02-14
Make donut chart background transparent

hafstad:


src/kirby/components/chart/chart-helper.tns.ts
```diff
@@ -1,9 +1,10 @@
-import { Options } from 'highcharts';
-import { ElementRef } from '@angular/core';
-import { WebView, LoadEventData } from 'tns-core-modules/ui/web-view/web-view';
-import { android, AndroidApplication, AndroidActivityEventData } from 'tns-core-modules/application';
-import { isAndroid, isIOS } from 'tns-core-modules/platform';
+import {Options} from 'highcharts';
```
jonasjensen-brolstark: There is no need to remove whitespaced here, would make your changes more obvious 😄 
 ```suggestion
import { Options } from 'highcharts';
```

hafstad:
Will close this for now

PR #117, opened 2019-02-21
Fixed the import of list-cell-line.component

MortenGregersen:
To match the other components


MortenGregersen:
> You shouldn't check in `package-lock.json` if you haven't changed any dependencies

PR #121, opened 2019-02-26
RTI-311: Prettier, TSLint and custom TSLint rules. 

askarby:
Introducing prettier and custom tslint rules.


.prettierrc.json
```diff
@@ -0,0 +1,9 @@
+{
+  "arrowParens": "always",
+  "bracketSpacing": true,
+  "parser": "typescript",
+  "printWidth": 100,
```
jonasjensen-brolstark: Is this width up for disquisition? I find it sometimes does more harm then good when splitting things after 100 chars.

src/app/examples/list-example/base-list.component.ts
```diff
@@ -1,12 +1,17 @@
 export class BaseListComponent {
-    myItems = [
-        {title: 'Vestas Wind Systems has a very long name', subTitle: '2000 stk', amount: '5.587.218.309 DKK', detail: 225},
-        {title: 'A.P. Møller-Mærsk', subTitle: '2 stk', amount: '14.019 DKK', detail: 0},
-        {title: 'Novo Nordisk A/S B', subTitle: '18 stk', amount: '7560 DKK', detail: 171},
-        {title: 'Danske Bank A/S', subTitle: '10 stk', amount: '1560 DKK', detail: -171},
-    ];
+  myItems = [
+    {
```
jonasjensen-brolstark: Why did prettier only format one of the json objects? or was it done manually.

tools/tslint-rules/disallowSpecificEnvironmentImportRule.ts
```diff
@@ -0,0 +1,45 @@
...
+    type: 'maintainability',
+    description: `Ensures that no imports contain specific environment.ts file (as it's handled by Angular/webpack).`,
+    options: null,
+    optionsDescription: 'Not configurable',
+    rationale: `Explicitly importing 'environment.mock.ts', 'environment.prod.ts' etc. is a no-go.`,
+    typescriptOnly: false,
+  };
+
+  static readonly FAILURE_STRING = `Do not import from specific environment file as it's handled by Angular/webpack. Remove the suffix ('.mock' or '.prod').
+*** Please note: Environment import can be simplified as: "import { environment } from '@drb-environment';"`;
```
jonasjensen-brolstark: Well we don't have @drd-environment in Kirby? 😄 

We should properly change this error string.

.prettierrc.json
askarby: I just copied this from relationsbanken, and really don't know if a (consious) decision has been made for 100 characters (I assume someone has).

Prettier actually has a strong opinion of not using more than 80 characters (have a look at this link: https://github.com/prettier/prettier/issues/2375)

Maybe ask @jakobe ?

src/app/examples/list-example/base-list.component.ts
askarby: I has actually formatted "all".

Prettier's philosophy is to "cram as much stuff into one line as possible", but if all properties of a JS-object can't be on a single line, it'll nicely format it (by having one property per line).

tools/tslint-rules/disallowSpecificEnvironmentImportRule.ts
askarby: I agree... will do!

src/app/examples/list-example/base-list.component.ts
jonasjensen-brolstark: Understood, so this is the Prettier way of doing. Let follow that 😄

.prettierrc.json
jonasjensen-brolstark: So the question is more how hard we want to follow Prettier. Should we do the recommended 80 characters then?

But lets hear the opinion of @jakobe and Alexander.

src/main.ns.ts
```diff
@@ -3,22 +3,6 @@ import { platformNativeScriptDynamic } from 'nativescript-angular/platform';
 
 import { AppModule } from './app/app.module';
 import './dev/livesync-navigation';
-const options: AppOptions = {};
```
jonasjensen-brolstark: Is this deletion on purpose or a merge error?

jonasjensen-brolstark: Doh is the change from master that deleted this. Ignore my comment.

PR #129, opened 2019-03-01
Update issue templates

jonasjensen-brolstark:


hafstad:
I'm not sure I like creating questions as issues at Github, I would prefer another channel for this. However if that's what we wanna do it's looking good to me.

PR #130, opened 2019-03-04
Feature/kirby progress bar

PetarRing:


src/app/intro/intro.component.tns.html
```diff
@@ -1,15 +1,15 @@
...
-        <Button kirby-button text="Grid" [nsRouterLink]="['/examples/grid']" class="btn btn-primary"></Button>
-        <Button kirby-button text="Fonts" [nsRouterLink]="['/examples/fonts']" ></Button>
-        <Button kirby-button text="Charts" [nsRouterLink]="['/examples/chart']" ></Button>
-        <Button kirby-button text="Spinner" [nsRouterLink]="['/examples/spinner']" ></Button>
-        <Label text="{N} Only Components:" style="font-size: 18; margin: 10"></Label>
-        <Button kirby-button text="Doughnut Chart" [nsRouterLink]="['/examples/nativescript-only/doughnut-chart']" class="btn btn-primary"></Button>
-        <Button kirby-button text="Line Chart" [nsRouterLink]="['/examples/nativescript-only/line-chart']" class="btn btn-primary"></Button>
-    </Stacklayout>
-</ScrollView>
+<Stacklayout>
```
jonasjensen-brolstark: Yaaaa.. lets not remove the scroll view here. On a iPhone 6 i can no longer scroll in the demo app and get to the bottom menu item.

image

src/kirby/components/progress-bar/progress-bar.component.scss
```diff
@@ -0,0 +1,33 @@
+@import '../../scss/web-imports';
```
jonasjensen-brolstark:
```suggestion
@import 'web-imports';
```

src/kirby/components/progress-bar/progress-bar.component.scss
```diff
@@ -0,0 +1,33 @@
...
+            width: 100%;
+            // grid-column-start: 1;
+            // grid-column-end: 101;
+        }
+
+        &.progress {
+            background-color: var(--kirby-primary);
+            // grid-column-start: 1;
+        }
+
```
jonasjensen-brolstark: Could we please save a bit on the whitespace? The UX's needs it. 😉

src/kirby/components/progress-bar/progress-bar.component.scss
```diff
@@ -0,0 +1,33 @@
...
+    position: relative;
+    width: 100%;
+    height: 18px;
+    overflow: hidden;
+    border-radius: 20px;
+    .bar {
+        height: 100%;
+        width: 100%;
+        position: absolute;
+        // display: grid;
```
jonasjensen-brolstark: If you don't need it, remove it.

src/kirby/components/progress-bar/progress-bar.component.spec.ts
```diff
@@ -0,0 +1,27 @@
+/* tslint:disable:no-unused-variable */
```
jonasjensen-brolstark: what is this line good for?

as far as i can see you don't have any unused variables.

src/kirby/components/progress-bar/progress-bar.component.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { DebugElement } from '@angular/core';
+
+import { ProgressBarComponent } from './progress-bar.component';
+
+describe('ProgressBarComponent', () => {
```
jonasjensen-brolstark: Maybe we should come up with a test or two like:

- Can it handle negative numbers?
- what if no color is defined
- etc

src/kirby/components/progress-bar/progress-bar.component.tns.html
```diff
@@ -0,0 +1,14 @@
+
+<AbsoluteLayout class="bars-container" backgroundColor="lightgray">
```
jonasjensen-brolstark: Do we need the `backgroundColor` on the `AbsoluteLayout` element?

src/kirby/components/progress-bar/progress-bar.component.tns.scss
```diff
@@ -0,0 +1,29 @@
+
+@import "../../scss/native-imports";
```
jonasjensen-brolstark:
```suggestion
@import "native-imports";
```

src/kirby/components/progress-bar/progress-bar.component.tns.scss
```diff
@@ -0,0 +1,29 @@
...
+            background-color: get-color('kirby-primary-lighter');
+        }
+
+        &.progress {
+            border-radius: 20 0 0 20;
+            height: $bar-height;
+            background-color: get-color('kirby-primary');
+        }
+    }
+
```
jonasjensen-brolstark: Ewww more whitespace.

jonasjensen-brolstark:
It is actual possible to do this with Highcharts. Should we maybe consider doing it that way? See [here](https://jsfiddle.net/0scvag6q/2/)

PetarRing:
> The implementation looks good. But i don't know about the name. `Progress bar` could make the user think that it is some kind of loading bar (as you see on youtube in top of the screen).
> 
> But what you have made is a chart. So would `progress chart` not be a better name? If it is a chart we should also move this into the charts area of the cookbook.

But isn't it also called progress bar in most other implementations? Such as the 'Highcharts' library? :)

jonasjensen-brolstark:
> > The implementation looks good. But i don't know about the name. `Progress bar` could make the user think that it is some kind of loading bar (as you see on youtube in top of the screen).
> > But what you have made is a chart. So would `progress chart` not be a better name? If it is a chart we should also move this into the charts area of the cookbook.
> 
> But isn't it also called progress bar in most other implementations? Such as the 'Highcharts' library? :)

Your right. Progress chart is a stupid name.

src/kirby/components/progress-bar/progress-bar.component.spec.ts
PetarRing: 1. I am a bit noobish in the testing, but.... How would we test if it could handle negative values? The value is used ina binding so to test if it could handle it, we would need to test the UI right? And see that the width is not disturbed in any way.
2. If no color is defined, a default color is provided. Don't know how to test that either since it again is a UI thing :D

jonasjensen-brolstark: 1. Well maybe you need to define if negative values are allowed. And the same goes what if progress is over 100%. Does that make sense to have negative and >100 progress numbers?
2.  You are right, you would have to test the view of the component to test the default values or you would have to test that the div elements styling. It can be done. Maybe have a look on the `list` component.

I would say that 1. tests are more important then 2. tests.

src/app/examples/progress-bar-example/progress-bar-example.component.tns.html
```diff
@@ -0,0 +1,12 @@
+<ScrollView>
```
jonasjensen-brolstark: There is some more indentation issues here 😄 

Sorry i missed that first time around.

src/kirby/components/progress-bar/progress-bar.component.tns.scss
```diff
@@ -0,0 +1,27 @@
...
+@import "../../scss/native-imports";
+
+$bar-height: size('m');
+
+.bars-container {
```
jonasjensen-brolstark: There is still some indentation work here 😄

PR #131, opened 2019-03-04
Feature/list cell align

SilverGull:
- space-between and space-around added to horisontalAlignment
- stretch and baseline added to verticalAlignment
- getAlignItems() => getHorisontalAlignment()
- getJustifyContent() => getVerticalAlignment()


src/app/examples/list-example/list-example.component.html
```diff
@@ -1,3 +1,16 @@
+<!--
```
askarby: I like that we're starting to explain functionality in the cookbook... looks good!

PR #137, opened 2019-03-07
Fix/list warnings

SilverGull:
- Fixed warnings in list when no input has been set, by adding default value
- stk => pcs in example data
- A couple of elements in progress-bar-example.tns.html was "self closed". That is not supported.
- Wrong methods calls in list-cell.component.tns.html fixed

### DIFF
--- a/src/app/examples/list-example/base-list.component.ts
+++ b/src/app/examples/list-example/base-list.component.ts
@@ -2,13 +2,13 @@ export class BaseListComponent {
   myItems = [
     {
       title: 'Vestas Wind Systems has a very long name',
-      subTitle: '2000 stk',
+      subTitle: '2000 pcs',
       amount: '5.587.218.309 DKK',
       detail: 225,
     },
-    { title: 'A.P. Møller-Mærsk', subTitle: '2 stk', amount: '14.019 DKK', detail: 0 },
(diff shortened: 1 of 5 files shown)


src/app/examples/progress-bar-example/progress-bar-example.component.tns.html
```diff
@@ -1,12 +1,12 @@
 <ScrollView>
   <StackLayout>
     <kirby-card>
-      <kirby-card-header title="Progress bar" subtitle="Default Example" />
-      <kirby-progress-bar progress="33" />
+      <kirby-card-header title="Progress bar" subtitle="Default Example"></kirby-card-header>
```
jonasjensen-brolstark: I would not touch this, as it has nothing to do with the fix. Could you revert it?

src/app/examples/list-example/base-list.component.ts
```diff
@@ -2,13 +2,13 @@ export class BaseListComponent {
   myItems = [
     {
       title: 'Vestas Wind Systems has a very long name',
-      subTitle: '2000 stk',
+      subTitle: '2000 pcs',
```
jonasjensen-brolstark: I would not touch this, as it has nothing to do with the fix. Could you revert it?

PR #152, opened 2019-03-14
Refactor/list

SilverGull:
- Unused styling removed
- Value mapping using Map instead of enum

### DIFF
--- a/src/kirby/components/list/list-cell/list-cell.component.ts
+++ b/src/kirby/components/list/list-cell/list-cell.component.ts
@@ -1,19 +1,17 @@
 import { Component, OnInit, Input, HostBinding } from '@angular/core';
 
-// TODO: refactor - use map instead of enum
-enum verticalAlignmentEnum {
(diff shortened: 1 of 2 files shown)


src/kirby/components/list/list-cell/list-cell.component.ts
```diff
@@ -57,35 +55,27 @@ export class ListCellComponent implements OnInit {
...
-      if (horisontalAlignmentEnum[this.horisontalAlignment]) {
-        return horisontalAlignmentEnum[this.horisontalAlignment];
-      } else if (
-        // '-' not supported in enum
-        this.horisontalAlignment === 'space-between' ||
-        this.horisontalAlignment === 'space-around'
-      ) {
-        return this.horisontalAlignment;
-      }
+    if (this.horisontalAlignment && alignmentMap.get(this.horisontalAlignment)) {
```
jonasjensen-brolstark: We should add tests for this. Same goes for the other function

PR #172, opened 2019-03-25
Feature/component status

jakobe:


src/app/component-status/component-status.component.html
```diff
@@ -0,0 +1,100 @@
+
+<div class="hero">
+    <h1>Component Status</h1>
+</div>
+<p>The latest component updates and changes. If you have a suggestion for a new
+    component not listed here,
+    <a href="https://github.com/kirbydesign/designsystem/issues/new/choose">create an issue on GitHub</a></p>
```
hafstad: Consider linking to the new component issue template, and maybe also use the github env variable.

PR #217, opened 2019-04-17
Enhancement/component status GitHub integration

jakobe:


src/app/component-status/component-status.component.ts
```diff
@@ -161,22 +282,29 @@ export class ComponentStatusComponent implements OnInit {
...
-          return <GithubIssue>{
-            number: githubIssueNo,
-            status,
-          };
-        });
-      })
+      catchError((_) => {
+        this.gitHubError = true;
+        this.isLoading = false;
+        return of([] as GithubCard[]);
```
mmmtrifork: isn't "as GithubCard[]" the React way of casting ? I would use <GithubCard[]>[] instead. It is a very minor thing, so you may ignore it, I am just curious

jakobe: https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#new-tsx-file-extension-and-as-operator

PR #230, opened 2019-04-30
Fix/modal issues

### LINKED ISSUE #221: [BUG] Modal tab issues
**Describe the bug**
1. The modal should take focus when its opened - in the cookbook the button that opens the modal keeps the focus and you can then scroll in the background.
2. Should be able to tab to the close button (X) - currently it only tabs through the two buttons.
3. When first tabbing it scrolls the background to the bottom (Might be related with no. 1)

**To Reproduce**
1. Go to https://cookbook.kirby.design/home/showcase/modal
2. Click open modal
3. Scroll with mouse - it scrolls in the background
4. Press tab and it scrolls the background to the bottom

**Kirby version**
 - v0.0.120

**Desktop (please complete the following information):**
 - OS: Mac
 - Browser Chrome
 - Version Newest

mmmtrifork:
fixes #221


src/kirby/components/modal/modal.component.ts
```diff
@@ -9,14 +9,34 @@ import { IModalController } from './services/modal.controller.interface';
   templateUrl: './modal.component.html',
   styleUrls: ['./modal.component.scss'],
 })
-export class ModalComponent {
+export class ModalComponent implements AfterViewInit {
+  @ViewChild('modalWrapper') modalWrapper: ElementRef;
+  scrollY: number = Math.abs(window.scrollY);
```
unnamjolner: Can't this be really expensive? I wonder how many of these events are emitted for a single scroll down the page.

src/kirby/components/modal/services/modal.controller.ts
```diff
@@ -29,7 +29,7 @@ export class ModalController implements IModalController {
     modalCloseEvent.then((data) => {
       this.forgetTopModal();
       if (onCloseModal) {
-        onCloseModal(data);
+        onCloseModal(typeof data === 'object' && 'data' in data ? data.data : data);
```
unnamjolner: Quite complex to read. Maybe some different naming or an explanation would help.
Of course, I happen to know the bug and why the fix is there... But still. :)

src/kirby/components/modal/modal.component.ts
mmmtrifork: I wouldn't say so. This is the position of the scroll at the time when we clicked on the modal-opening button. We compute it only once and then use it inside the class, more precisely the value is used every time the focus has changed (which happens when we press tab on the keyboard, or click on something with the mouse). With this said, I am confused about which events you refer to, maybe I am not seeing something.

src/kirby/components/modal/services/modal.controller.ts
mmmtrifork: agreed, i will add a comment to clarify :)

src/kirby/components/modal/modal.component.ts
unnamjolner: Wow, sorry. I'm tired today. I thought you subscribed to scroll events.

src/kirby/components/modal/services/modal.controller.ts
unnamjolner: Super :)

src/kirby/components/modal/modal.component.ts
```diff
@@ -34,7 +34,8 @@ export class ModalComponent implements AfterViewInit {
   onModalDismiss(e: any) {
     // Handle key press, due to:
     // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Required_JavaScript_Features
-    if (e && e.keyCode !== 32 && e.keyCode !== 13) {
+    if (e && e.keyCode && e.keyCode !== 32 && e.keyCode !== 13) {
+      console.log(`interrupting dismiss, keycode: ${e.keyCode}`);
```
unnamjolner: I'm thinking this console.log should probably not make it to master? Otherwise no comments. :)

PR #233, opened 2019-05-02
Enhancement/custom icons

szutrifork:


src/kirby/components/icon/custom-icon-name.directive.tns.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.customName) {
+      this.handleCustomIcon(this.customName);
+    }
+  }
+
+  handleCustomIcon(name: string) {
+    if (name && this.customIconSettings[0].icons) {
```
jonasjensen-brolstark: what is the `[0]` good for here? `customIconSettings` is not an array.

Or what am i missing here 😄

src/kirby/components/icon/custom-icon-name.directive.tns.ts
```diff
@@ -0,0 +1,65 @@
...
+        console.warn('Missing font-family.');
+      } else if (icon === undefined) {
+        console.warn('Icon with name', name, 'was not found in custom font.');
+      } else {
+        this.setCustomIcon(icon.unicode, fontFamily);
+      }
+    }
+  }
+
+  getCustomIcon(name: string): { name: string; svg: string; unicode: string } {
```
jonasjensen-brolstark: I find it nice to mark functions as private, just so the reader of the code clearly can see that they are not called anywhere beside the containing class. Not a major issue, but one you should consider.

src/kirby/components/icon/custom-icon-name.directive.tns.ts
```diff
@@ -0,0 +1,65 @@
...
+    return this.customIconSettings[0].icons.find((icon) => icon.name === name);
+  }
+
+  setCustomIcon(unicode: string, fontFamily: string) {
```
jonasjensen-brolstark: same as the comment for `getCustomIcon`.

src/kirby/components/icon/custom-icon-name.directive.tns.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+
+  handleCustomIcon(name: string) {
+    if (name && this.customIconSettings[0].icons) {
+      const fontFamily = this.customIconSettings[0].fontfamily;
+      const icon = this.getCustomIcon(name);
+
+      if (fontFamily === undefined) {
+        console.warn('Missing font-family.');
```
jonasjensen-brolstark: This is my personal opinion, but i find it better that the code blows up then warns me in a console, that i am using it wrong.

Point being, that if i remove the `fontFamily` or uses an `icon` that is not in the settings, there is something wrong and i may overlook a warning in the console.

src/kirby/components/icon/custom-icon-name.directive.tns.ts
```diff
@@ -0,0 +1,65 @@
...
+      this.handleCustomIcon(this.customName);
+    }
+  }
+
+  handleCustomIcon(name: string) {
+    if (name && this.customIconSettings[0].icons) {
+      const fontFamily = this.customIconSettings[0].fontfamily;
+      const icon = this.getCustomIcon(name);
+
+      if (fontFamily === undefined) {
```
jonasjensen-brolstark: Would a `if(!fontFamily)` not be better here, as if i set `fontFamily` to null it would fail? 

The same goes for the check of the icon.

src/kirby/components/icon/custom-icon-name.directive.ts
```diff
@@ -0,0 +1,43 @@
...
+
+      if (icon !== undefined) {
+        this.setCustomIcon(icon.svg);
+      } else {
+        console.warn('Icon with name', name, 'was not found in custom font.');
+      }
+    }
+  }
+
+  getCustomIcon(name: string): { name: string; svg: string; unicode: string } {
```
jonasjensen-brolstark: Same comment as in the `tns` version of the directive, you may what to mark the private functions as private, just to make it easier to read the code. 😄

src/kirby/components/icon/custom-icon-name.directive.tns.ts
szutrifork: It is an array because 'multi: true' under '{ provide: CUSTOM_FONT_SETTINGS, useValue: customIconSettings, multi: true },' if it's set to false it will not be an array.

src/kirby/components/icon/custom-icon-name.directive.tns.ts
jakobe: Agree - except, how do we catch this at compile time? Or will it be catched in tests?
If it only happens runtime we might not wanna fail hard?

src/kirby/components/icon/custom-icon-name.directive.tns.ts
```diff
@@ -0,0 +1,65 @@
...
+  public ngOnChanges(changes: SimpleChanges): void {
+    if (changes.customName) {
+      this.handleCustomIcon(this.customName);
+    }
+  }
+
+  private handleCustomIcon(name: string) {
+    if (name && this.customIconSettings[0].icons) {
+      const fontFamily = this.customIconSettings[0].fontfamily;
+      const icon = this.getCustomIcon(name);
```
jakobe: Probably need to have a way of finding the right customIconSettings and then sending that as a parameter to getCustomIcon...

src/kirby/components/icon/custom-icon-name.directive.tns.ts
szutrifork: Support for having multiple custom fonts is removed.

src/kirby/components/icon/custom-icon-name.directive.tns.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public ngOnChanges(changes: SimpleChanges): void {
+    if (changes.customName) {
+      this.handleCustomIcon(this.customName);
+    }
+  }
+
+  private handleCustomIcon(name: string) {
+    if (name && this.customIconSettings[0].icons) {
```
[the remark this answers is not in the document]
szutrifork: Support for multiple custom fonts is removed

src/kirby/components/icon/custom-icon-name.directive.tns.ts
szutrifork: Support for multiple custom fonts is removed

PR #240, opened 2019-05-07
Enhancement/fix for 239

### LINKED ISSUE #239: [Enhancement] Kirby list section header height on iOS
Right now the section headers in a list, is on iOS hardcoded to be 50 in height. This prevents the users of kirby of having any larger sections headers then that.

If the hardcoded height was removed from kirby and the responsibility, of defining a height, was moved over to the user of kirby, the user would have more options.

jonasjensen-brolstark:
Fixes #239 

I have removed the hardcode height on the wrapper for the section header in a list. I moved this height to the built-in `kirby-list-section-header` component.

I also updated the documentation for the kirby list to explain the user of kirby, that if they don't use the built-in component for sections they will have to set the height for iOS explicit.

### DIFF
--- a/src/app/showcase/list-showcase/list-showcase.component.html
+++ b/src/app/showcase/list-showcase/list-showcase.component.html
@@ -1,4 +1,10 @@
 <div class="example">
   <kirby-list-example></kirby-list-example>
   <kirby-html-viewer [html]="exampleHtml"></kirby-html-viewer>
+  <br />
+  <br />
+  <span>When <b><em>not</em></b> using the built-in <code>kirby-list-section-header</code> component you will need to natively to define a height for iOS</span>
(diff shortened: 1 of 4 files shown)


src/app/showcase/list-showcase/list-showcase.component.html
```diff
@@ -1,4 +1,10 @@
 <div class="example">
   <kirby-list-example></kirby-list-example>
   <kirby-html-viewer [html]="exampleHtml"></kirby-html-viewer>
+  <br />
+  <br />
+  <span>When <b><em>not</em></b> using the built-in <code>kirby-list-section-header</code> component you will need to natively to define a height for iOS</span>
```
MortenGregersen:
```suggestion
  <span>When <b><em>not</em></b> using the built-in <code>kirby-list-section-header</code> component you will need to natively define a height for iOS</span>
```

PR #258, opened 2019-05-13
clean up and card padding fix for ios/android

szutrifork:


src/kirby/components/card/card-header/card-header.component.tns.scss
```diff
@@ -1,7 +1,7 @@
 @import '../../../scss/native-imports';
 
 .header {
-  margin: 16;
+  margin: font-size("s");
```
jonasjensen-brolstark: Should we not use `size('s')` here instead?  size `s` is actual `16` where font-size `s` is `14`.

src/kirby/components/card/card.component.tns.scss
```diff
@@ -4,13 +4,14 @@ $color-card-bg: get-color('kirby-contrast-light');
 
 .background, 
 .container {
+    display: flex;
     flex-direction: column;
-    width: 100%;
+    text-align: center;
+    margin: 8;
```
jonasjensen-brolstark: Here you should use `size('s')`.
```suggestion
    margin: size('s');
```

src/kirby/components/modal/action-sheet/action-sheet.component.tns.scss
```diff
@@ -10,6 +10,7 @@ kirby-card {
   border-color: get-color("kirby-medium");
   width: 100%;
   height: 60;
+  padding: 20;
```
jonasjensen-brolstark: Could this be converted to a none hardcode value? aka. use the `size` function.

PR #269, opened 2019-05-15
card component using scss instead of color helper

szutrifork:


src/kirby/components/card/card.component.tns.ts
```diff
@@ -8,7 +8,7 @@ import { registerElement } from 'nativescript-angular';
 import { ContentView } from 'tns-core-modules/ui/content-view';
 
 import { ScssHelper } from '../../scss/scss-helper';
-import { ColorType, ColorHelper } from '../../helpers/color-helper';
+import { ColorType } from '~/kirby/helpers/color-type';
```
jakobe: Don't use `~` imports as it breaks symlinking (there's as reason why the line above uses relative path...)

src/kirby/components/card/card.component.ts
```diff
@@ -2,7 +2,7 @@ import { Component, OnInit, Input, ElementRef, Renderer2, OnDestroy } from '@ang
 
 import { ResizeObserverService } from '../shared/resize-observer/resize-observer.service';
 import { ResizeObserverEntry } from '../shared/resize-observer/types/resize-observer-entry';
-import { ColorHelper, ColorType } from '../../helpers/color-helper';
+import { ColorType } from '~/kirby/helpers/color-type';
```
jakobe: Don't use `~` imports as it breaks symlinking (there's as reason why the line above uses relative path...)

PR #278, opened 2019-05-16
Enhancement/shadow option for card

szutrifork:


src/app/showcase/card-showcase/card-showcase.component.html
```diff
@@ -2,5 +2,7 @@
   <kirby-card-example></kirby-card-example>
   <kirby-code-viewer [html]="exampleHtml"></kirby-code-viewer>
   <h4>Properties:</h4>
-  <kirby-showcase-properties [properties]="properties"></kirby-showcase-properties> 
+  <kirby-showcase-properties [properties]="properties"></kirby-showcase-properties>
+  <br>
+  <p>Shadow on kirby-card will be automatically removed if there is a (select)="onCardSelected()"</p>
```
jakobe: Nope, should be the opposite :)
If you can click/tap a card it should be elevated (= shadow)

src/kirby/components/card/card.component.html
```diff
@@ -1,4 +1,4 @@
-<div [ngClass]="colortype">
+<div [ngClass]="colortype" (click)="select.emit()">
```
jakobe: We don't do it this way, make a function for that:
```suggestion
<div [ngClass]="colortype" (click)="onCardSelect()">
```

src/kirby/components/card/card.component.tns.html
```diff
@@ -1,4 +1,4 @@
-<FlexboxLayout (loaded)="onViewLoaded($event)" class="background" [ngClass]="cardSizeClass">
+<FlexboxLayout (loaded)="onViewLoaded($event)" class="background" [ngClass]="cardSizeClass" (tap)="select.emit()">
```
jakobe: We don't do it this way, make a function for that:
```suggestion
<FlexboxLayout (loaded)="onViewLoaded($event)" class="background" [ngClass]="cardSizeClass" (tap)="onCardSelect()">
```

src/kirby/components/card/card.component.html
jakobe: I think it should be a @HostListener instead, so we bind the click event to the host-tag (this wrapper div should be removed)

src/kirby/components/card/card.component.tns.ts
```diff
@@ -25,18 +25,18 @@ export class CardComponent extends ContentView implements OnInit {
...
   cardSizeClass = '';
-
+  applyShadow: boolean = true;
   constructor(private zone: NgZone) {
     super();
   }
 
-  ngOnInit() {}
+  ngOnInit() {
+    this.applyShadow = this.select.observers.length === 0;
```
jakobe: What about `onChanges`?

jakobe:
Won't implement, gets resolved by #424

PR #285, opened 2019-05-20
Fix/absolute imports

jakobe:
This change prevents us from importing from `~/kirby` in Kirby source code
It also ensures classes are exported through `index.ts` by importing from `@kirbydesign/designsystem` in the cookbook itself.


jakobe:
> Looking good to me, did we ever manage to create an issue of adding a custom TSLint rule, to prevent us from using absolute imports (inside the components)?

This PR actually solves that exact issue 👍

PR #286, opened 2019-05-20
Changed kirby-icon to use the new size directive.

jonasjensen-brolstark:
I have refactored `kirby-icon` to use the new size directive and aligned the sizes with those in Zeplin.

To support the size directive, i have had to make the `icon.component.tns.scss` global. Aka. rename it to `icon.component.tns.global.scss` and import it in `_native-common.scss`.

I have also added the css class `kirby-icon` on native, so i can style it globally.


src/kirby/components/icon/icon.component.scss
```diff
@@ -1,5 +1,30 @@
...
+
 :host {
-	display: flex;
-    align-items: center;
-    justify-content: center;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  
+  ion-icon {
```
jakobe: No need to nest this selector, as it's already scoped to the host element...

src/kirby/components/icon/icon.component.scss
```diff
@@ -1,5 +1,30 @@
...
+    font-size: size('m');
+  }
+
+  &.sm {
```
jakobe: Consider:
```suggestion
  :host(.sm) {
```

src/kirby/components/icon/icon.component.scss
```diff
@@ -1,5 +1,30 @@
...
+    ion-icon {
```
jakobe: Consider flattening:
```suggestion
    :host(.sm) ion-icon {
```

src/kirby/components/icon/icon.component.tns.global.scss
```diff
@@ -1,15 +1,26 @@
 @import '../../scss/native-imports';
  
-Label {
-    font-family: "Ionicons", "ionicons";
+.sm {
```
jakobe: Ensure you only target the icon when moving styles to global:
```suggestion
kirby-icon.sm {
```

src/kirby/components/icon/icon.component.tns.global.scss
```diff
@@ -1,15 +1,26 @@
...
+    font-size: size('m');
+  }
+}
 
-    &.small {
-        font-size: 16px;
-    }
+.md {
+  > .kirby-icon {
+    font-size: size('l');
```
jakobe: So size `md` doesn't map to font-size `l`??

src/kirby/components/icon/icon.component.tns.global.scss
jonasjensen-brolstark: As i wrote in the [description](https://github.com/kirbydesign/designsystem/pull/286#issue-280359718) of this PR, i can not make it work, if i use the component name as selector on native. 

That is why i added the `kirby-icon` css class to the `Label` on native.

PR #289, opened 2019-05-21
:bug: Add null check for grouped list

jakobe:


src/kirby/components/list/list.component.ts
```diff
@@ -94,9 +94,12 @@ export class ListComponent implements OnInit, OnChanges {
 
   ngOnChanges(): void {
     this.isSectionsEnabled = !!this.getSectionName;
-    if (this.isSectionsEnabled) {
+    if (this.isSectionsEnabled && this.items) {
```
askarby: We should really consider adding this scenario to a test-case, so that we don't make any regressions on this issue.

PR #296, opened 2019-05-21
updated modal size to the correct one

mmmtrifork:


unnamjolner:
Looks good! I have talked to Jonathan from UX about sizing. This size is fine for now. In the future they might want a couple of different sizes (or the options to set the size).

PR #301, opened 2019-05-22
Possibility to suppress styling on section headers and rows

dalsgaard:


src/kirby/scss/base/_typography.scss
```diff
@@ -31,6 +31,7 @@ h3,
   font-weight: font-weight("bold");
 }
 
+p,
```
jakobe: Remove:
```suggestion
```

PR #415, opened 2019-07-05
Enhancement/bump ionic version to 4.6.0

szutrifork:


src/kirby/package.json
```diff
@@ -24,7 +24,7 @@
   "homepage": "https://github.com/kirbydesign/designsystem#readme",
   "types": "index.d.ts",
   "dependencies": {
-    "@ionic/angular": "^4.1.1",
+    "@ionic/angular": "^4.6.0",
```
jonasjensen-brolstark: Should this not be a specific version too?

hafstad: Should be fixed now :)

PR #422, opened 2019-07-09
nativescript-ui-listview version bump

jonasjensen-brolstark:
Version bump to get fix for this [issue](https://github.com/NativeScript/nativescript-ui-feedback/issues/1145) into kirby.

Tested in the cookbook for both Android and iOS


hafstad:
@jonasjensen-brolstark Remember to change the version in src/kirby/package.json as well

PR #428, opened 2019-07-12
Enhancement/button colors gh 390

### LINKED ISSUE #390: [Enhancement] Button colors (light/white)
Right now its pretty complicated how to choose the right color because it can be everything in the long list of theme color. 

We want it to be 2 different choices:

Light & White 

See also:
https://zpl.io/VDKnZAl

jakobe:
Closes #390


src/kirby/helpers/color-helper.ts
```diff
@@ -21,6 +21,10 @@ export class ColorHelper {
     return ColorHelper.getColor(name + '-contrast');
   }
 
+  public static getColorBrightness(name: string) {
```
hafstad:
```suggestion
  static getColorBrightness(name: string) {
```

src/kirby/helpers/color-helper.ts
```diff
@@ -21,6 +21,10 @@ export class ColorHelper {
...
+    return ColorHelper.getColor(name + '-color-brightness');
+  }
+
   public static getThemeColorRgbString(name: string) {
```
hafstad:
```suggestion
   static getThemeColorRgbString(name: string) {
```

src/kirby/helpers/color-helper.ts
jakobe: But then I have to do it on all other methods as well, so I think that would be a candidate for another PR :)

PR #452, opened 2019-07-23
feat(list, web, native): select list row

### LINKED ISSUE #375: [Enhancement] Add option to enable selected state on list items
It should be possible to enable selected state on the Kirby list (e.g. a checkmark to the left of the list item).
The feature should be opt-in (default false).

https://app.zeplin.io/project/5c99d69a79d02b35ea3b8f41/screen/5d022940c297435df258d95c

lydemann:
Closes #375 

Mark text in list row as bold when the list row is selected by adding selected class and apply css.

Web:
image

Native:
image


src/app/examples/nativescript-only/list-section-example/list-section-example.component.tns.scss
```diff
@@ -6,6 +6,10 @@
   margin-bottom: 10;
 }
 
+.hide-check {
+  opacity: 0;
```
lydemann: Visibility didn't work here...

src/kirby/components/list/list-item/list-item.interface.ts
```diff
@@ -0,0 +1,8 @@
+export interface ListItem {
```
jonasjensen-brolstark: The `kirby-list` should not have any knowledge of the format of the items handed to it.
So i think this is not a route we should take, as it's opens a can of worms we don't want to work with 😄 

We hand back the `item` to the user of the list in the templates. So altering the `item` is a bad idea, as we risk to overwrite data.

So we will have to rethink how a item is marked as selected in the code.

lydemann: I added this, so I had a typesafe way of ensuring that the item had an id. This is used for checking if an item is selected. Without an interface, and an id, how would you check if an item is selected?

lydemann:
> See my comment. i would like to have a discussion. Så pinging @hafstad and @jakobe
> 
> Why is it, that kirby should have the responsibility of altering the styling of items in the list?
> Should that not be the responsibility of the user? The list already have a callback when an item is selected. As i see it, this is already possible to do, with the current implementation of the `kirby-list`.
> 
> I personally find it weird when framesworks does magic behind the scene for me 😄

Actually, I agree. I think the easiest and best developer experience for the client and the Kirby team would be to make the list flexible enough to handle all the use cases instead of building in a lot of "magic" to "silverplate" some use cases for you. I can see how this might end up in client overrides when another font-weight is wanted etc. But, this makes for one big push back on this feature in general. :D
The only thing, I think might be nice here, is to add the selected-item class to the row, so it is easier for the client to select it with a style rule.

hafstad:
@lydemann @jonasjensen-brolstark Normally I would totally agree with you, and this is a very interesting discussion. However what we're building is not a framework like bootstrap or material it's a design system. Therefore we're having sensible default according to the UX team. So we're loosing flexibility some places but instead we have consistency across.

This feature is thought as an opt-in because we don't wan't every single team to implement their own way doing this because this leads to inconsistency and more work for the feature teams.  

We should ofc. not apply this for every list which has the select event attach as some of them are not selectable lists but lists which navigates away to another page. What we could do was maybe having a mode="select" or type="select" which has this sensible defaults defined by the UX team.

lydemann:
@hafstad @jonasjensen-brolstark 
Now I have updated the PR with the input markSelectedRow to determine when the selected row text should turn bold.

src/kirby/components/list/list-cell-line/list-cell-line.component.tns.html
```diff
@@ -1 +1 @@
-<Label [text]="text" [ngClass]="{primary: primary}" *ngIf="text"></Label>
+<Label [text]="text" [ngClass]="{primary: primary, selected: selected}" *ngIf="text"></Label>
```
jonasjensen-brolstark: Be aware of potential performance issue with `[ngClass]`. I suggest to use the `[class.primary]`notation instead.

src/kirby/components/list/list.component.html
```diff
@@ -17,6 +17,7 @@
         <ion-item *ngFor="let item of section.items; let idx = index;trackBy: trackByFn"
                   [class.has-divider]="showDivider"
                   [class.selectable]="isSelectable"
+                  [class.selected]="markSelectedRow && selectedItem && item.id === selectedItem.id"
```
jonasjensen-brolstark: You can not take it for granted that the items handed to the kirby-list has an `id` property. So you shouldn't make use of any properties on the items.

You will have to find an other way to identify the selected item.

src/kirby/components/list/list.component.html
```diff
@@ -33,6 +34,7 @@
     <ion-item *ngFor="let item of items;trackBy: trackByFn"
               [class.has-divider]="showDivider"
               [class.selectable]="isSelectable"
+              [class.selected]="markSelectedRow && selectedItem && item.id === selectedItem.id"
```
jonasjensen-brolstark: Same as other comment regarding the use of the `id` property.

src/kirby/components/list/list.component.spec.ts
```diff
@@ -128,14 +128,16 @@ describe('ListComponent', () => {
...
       spyOn(component.itemSelect, 'emit');
-      const itemToBeSelected = { value: 'this is a dummy item' };
+      const itemToBeSelected = { value: 'this is a dummy item', selected: false };
+      component.items = [itemToBeSelected];
 
       component.onItemSelect(itemToBeSelected);
 
       expect(component.itemSelect.emit).toHaveBeenCalledTimes(1);
       expect(component.itemSelect.emit).toHaveBeenCalledWith(itemToBeSelected);
+      expect(itemToBeSelected.selected).toBe(true);
```
jonasjensen-brolstark: Same as other comment regarding the use of properties on the item.

src/kirby/components/list/list.component.tns.html
```diff
@@ -12,13 +12,14 @@
           [class.divider]="showDivider"
           [class.rounded]="shape === 'rounded'"
           [class.first]="isFirstItem(item, index)"
+          [class.selected]="markSelectedRow && !!selectedItem && item.id === selectedItem.id"
```
jonasjensen-brolstark: Same as other comment regarding the use of the `id` property.

src/kirby/components/list/list.component.ts
```diff
@@ -130,7 +142,12 @@ export class ListComponent implements OnChanges {
   }
 
   onItemSelect(args: any) {
-    this.itemSelect.emit(this.listHelper.getSelectedItem(this.items, args));
+    this.selectedItem = this.listHelper.getSelectedItem(this.items, args);
+    this.items.forEach((item) => {
```
jonasjensen-brolstark: I looks like you don't use the `selected` property anywhere. So you can revert this change.

src/kirby/components/list/list-item/list-item.interface.ts
lydemann: As talked about I will:
1. Add _id property with GUID on item input
2. Remove_id on data output
3. Delete ListItem interface

src/kirby/components/list/list.component.html
lydemann: Fixed this by adding dynamic _id guid

src/kirby/components/list/list.component.spec.ts
lydemann: Removed it

src/kirby/components/list/list.component.html
lydemann: Using dynamic internal _id guid

PR #454, opened 2019-07-24
feat(card, web, native): card padding input

### LINKED ISSUE #353: [Enhancement] (Option for enable) default padding in cards
I'm always frustrated when I use a `kirby-card`. For every card we use I need to set the same padding left and right (and top and/or bottom if no header and/or footer is used).

A default padding in the card, with an option to disable it. Almost everytime someone would need a card they want a default padding.

Maybe this could be implemented by creating a `kirby-card-content` component which has the padding. This way you could omit using it if you don't want the default padding.

lydemann:
Closes #353

Added input hasPadding to card, which ads default size padding to card.

Web:
image

Native:
image


src/kirby/components/card/card.component.tns.html
```diff
@@ -1,5 +1,5 @@
 <FlexboxLayout (loaded)="onViewLoaded($event)" class="background" [ngClass]="cardSizeClass" [kirbyElevation]="elevation">
-    <FlexboxLayout class="container" [ngClass]="themeColor">
+    <FlexboxLayout class="container {{themeColor}}" [ngClass]="{'padding': hasPadding}">
```
jonasjensen-brolstark: We have found that the use of `[ngClass]="{...}"` binding can have performance impacts on native. So i suggest that you use the `[class.padding]` notation instead.

src/app/examples/card-example/card-example.component.html
```diff
@@ -6,5 +6,16 @@
     Lorem, ipsum dolor sit amet consectetur adipisicing elit. Mollitia facere molestias recusandae necessitatibus ab veniam repellendus doloremque culpa quam libero, est quo accusamus cumque, in quia itaque cupiditate ratione repellat!
   </div>
 
+  <kirby-card-footer></kirby-card-footer>
+</kirby-card>
+
+<kirby-card [hasPadding]="true">
```
jonasjensen-brolstark: Just for the sake of the discussion, should this not be the other way around? That we default to having padding, but the user should actively choose not to have it?

Or is this done to keep backwards compatibility?

Paging @hafstad

src/kirby/components/card/card.component.tns.html
lydemann: Ok, will change that all places.

lydemann:
> Looks fine 😄
> 
> If you feel up to it, it would be okay with me, if you would 'fix' the `themeColor` class styling in native.
> 
> You could make the native card `scss` file global, add a host class `kirby-card` to the component. That way you should be able to style the `.container` class with like:
> 
> ```css-scss
> .kirby-card {
>   @each $color-name, $color-value in $main-colors {
>     &.#{$color-name} {
>       .container {
>         background-color: #{get-color($color-name)};
>         color: #{get-color($color-name+'-contrast')};
>       }
>     }
>   }
> }
> ```

Would it make sense to make more general global css classes for setting the background color, such as .bg-primary that can then easily be used everywhere?

src/app/examples/card-example/card-example.component.html
lydemann: I agree :) Will reverse it.

src/app/examples/card-example/card-example.component.html
```diff
@@ -6,5 +6,16 @@
     Lorem, ipsum dolor sit amet consectetur adipisicing elit. Mollitia facere molestias recusandae necessitatibus ab veniam repellendus doloremque culpa quam libero, est quo accusamus cumque, in quia itaque cupiditate ratione repellat!
   </div>
 
+  <kirby-card-footer></kirby-card-footer>
+</kirby-card>
+
+<kirby-card>
+  <kirby-card-header title="Padding" subtitle="Card with padding"></kirby-card-header>
```
hafstad: I would like the example to be reverted as well, so we don't have a case of "padding" which is default but a case special case showing you can remove the padding

lydemann: Ok, good point.

src/kirby/components/card/card.component.scss
```diff
@@ -22,6 +22,10 @@ $min-card-width: 304px;
   min-width: $min-card-width;
   overflow: hidden;
 
+  &.padding {
```
hafstad: Right now the whole card gets default padding, which is also visible on the card-example badges. I think we maybe should consider creating a card-content component and apply the right default padding to the card-title, card-content and card-footer. In this way we will have a more flexible solution.

hafstad: I've just talked with UX and this is how they want it for now. We will therefore keep the padding on the whole card.

PR #455, opened 2019-07-24
style(list): styles for colored list items + make header and no-more-items transparent

### LINKED ISSUE #227: [Enhancement] List - Item background color

I would like to be able to determine the background color of single items in the list.

image

Right now, both the `KirbyListCell` and `KirbyListItem` elements are wrapped in an element that contains the background. So i am not able as user of the Kirby list to controll the background color of my items.

##### Option 1

One way of doing this would be to have a callback to return the wanted color for a single item:

```typescript
export class ListComponent implements OnInit {
...
@Input() getItemColor?: (item: any) => 'primary' | 'secondary' | 'tertiary' | 'warning' | 'success' | 'danger';
...
}
```

But I think this option is only a quick fix and will start to clutter up the lists interface.

##### Option 2

An other way is to use the already existing `KirbyListItem` construct and then do like on `KirbyIcon` and `KirbyBadge`:

```html
<kirby-list-item 
    *kirbyListItem="let item" 
    [colortype]="item.isInTheDangeZone ? 'success' : 'danger'" >
...
</kirby-list-item>
```

The same could be done for `KirbyListCell``

```html
<kirby-list-cell 
    *kirbyListCcell="let item" 
    [colortype]="item.isInTheDangeZone ? 'success' : 'danger'" >
...
</kirby-list-cell>
```

### LINKED ISSUE #403: [BUG] white background color on kirby list header and no-more-items in web
image

The are set a background color on the .list-header class and it is needed to be transparent
image

The same with no-more-items
image

image

lydemann:
Closes #227 and closes #403

- #227 Directive bgColorType for setting bg-*color-type* classes on list item
- #403 Making list header and no-more-items transparent

UPDATE: Transclude the whole row to make color fill out the whole row

Web:
list-color-row-rewrite-web

Native:
list-color-row-rewrite-native


src/kirby/components/list/directives/bg-color-type.directive.ts
```diff
@@ -0,0 +1,54 @@
+import { Directive, HostBinding, Input } from '@angular/core';
```
jonasjensen-brolstark: Is there a reason for creating this directive and not use using the `themeColor` directive? It looks like a 100% duplicate of it.

Are we expecting to both have `themeColor` and `bgColorType` on some elements?

If we have to have two directives, lets have a common base class as we don't want to maintain to list of our colors.

lydemann: The difference is that themeColor is styling both the background and the text. We might be able to just use one and apply different styling, as they are only setting a css class.

lydemann:
> I think you want to close #227 and not #277 😄
> 
> But we have a scenario that is not mentioned in the #227 issue. What if i have a kirby-list like:
> 
> ```
> <kirby-list ... >
>   <my-list-item-component *kirbyListItem="let item" [item]="item"></my-list-item-component>
> </kirby-list>
> ```
> 
> How is it possible for me to set the background color of a single item?
> Right now all items are wrapped in a container with class `row`:
> 
> ```
> <StackLayout [kirbyElevation]="2" [iosMasksToBounds]="true">
>   <StackLayout class="row"
>     [class.divider]="showDivider"
>     [class.rounded]="shape === 'rounded'"
>     [class.first]="isFirstItem(item, index)"
>     [class.last]="isLastItem(item, index)">
>     ...
>   </StackLayout>
> </StackLayout>
> ```
> 
> And this `row` class has padding, so i can not just add a background color to `my-list-item-component`.
> 
> I think you have the same issue. The background color of the `kirby-list-item` doesn't go all the way to the edge.
> 
> So you will have to set the background color on the container of the item. And if we don't want the kirby list to know of the data of incoming items, it properly have to be a callback for each single item to get its background color.
> 
> **Edit**: The same issue is on web, where we just have an `ion-item` instead of a class `row`.

We could get the bg-color directive as contentChild and apply the class to the row? Or make it possible to disable the row padding and set margin on the my list item instead? Maybe it would actually be the most flexible solution if the whole row was transcluded using ng-content and then as default the different classes was applied in the same way, so the client still can control everything?

lydemann:
@jonasjensen-brolstark I made the color fill out the whole row by moving the "row wrapper with padding (ion-item and row in native)" into the transcluded items. This also improved the design by separating the rows from the list container better.

src/kirby/components/list/directives/bg-color-type.directive.ts
lydemann: I ended up using just themeColor directive

src/kirby/directives/theme-color/theme-color.directive.ts
```diff
@@ -5,9 +5,7 @@ import { ColorHelper } from '../../helpers/color-helper';
 
 @Directive({
   // tslint:disable-next-line:directive-selector
-  selector: `kirby-card[themeColor],
-             kirby-badge[themeColor],
-             kirby-icon[themeColor]`,
+  selector: `[themeColor]`,
```
jonasjensen-brolstark: We did this for a reason (you couldn't know 😄 )

We did this because we didn't wanted the user of kirby to use `themeColor` on components that don't implement it yet and create a bug report on it. This way around the user would already on develop time see that a component don't support `themeColor`.

So let revert it. I know it is a hassle to maintain.

lydemann: Ok I see. But how about making the color style global, so we make it work for every scenario? @hafstad

jonasjensen-brolstark: > Ok I see. But how about making the color style global, so we make it work for every scenario? @hafstad

I think we should keep that talk in an issue for it self. But for sure, lets have the talk.

lydemann: Let's do. I have fixed this now and will resolve this then.

lydemann:
> container for infinite-scroll still looks weird.
> image
> 
> I also talked with @bommariusser, we don't wonna have all the colors implemented at this moment.
> We only want white and grey for now. Also as :hover the colors tint should be used and on-press the colors shade should be used as background-color.

Thanks for the feedback. This is now fixed.

jonasjensen-brolstark:
@lydemann I am closing this PR for now, as there are some changes to this feature we need to discuss. Lets do this face to face.

PR #478, opened 2019-08-01
Feature/changelog

hafstad:
Add automated generated changelog of kirby.
Releases are stored into firebase. 
The changelog fetches all tags from the kirby repo -> find commits related to the tags -> find pull-requests associated with the commit -> put it inside of firebase.

We will therefore be able to label our pull-requests to make breaking changes visible. 

The page will automatically fetch versions which is not added into firebase, currently at the client-side. However all versions so far is added, so it should not be a problem.

However in future it would be nice to create a cloud function which is triggered by Github webhooks.  

image


src/app/changelog/changelog.service.ts
```diff
@@ -0,0 +1,141 @@
...
+      .map((e, i, final) => final.indexOf(e) === i && i)
+
+      // eliminate the dead keys & store unique objects
+      .filter((e) => arr[e])
+      .map((e) => arr[e]);
+
+    return unique;
+  }
+
+  private versionsToUpdate(fromVersion?, toVersion?): Observable<any> {
```
mmmtrifork: Consider specifying types explicitly
```suggestion
  private versionsToUpdate(fromVersion?: string, toVersion?: string): Observable<any> {
```

src/app/shared/github/github.service.ts
```diff
@@ -0,0 +1,79 @@
...
+        } else {
+          return EMPTY;
+        }
+      }),
+      toArray(),
+      map(() => tags)
+    );
+  }
+
+  compareCommits(previousVersion, nextVersion) {
```
mmmtrifork: Consider adding types explicitly
```suggestion
  compareCommits(previousVersion: string, nextVersion: string) {
```

mmmtrifork: Consider adding types explicitly
```suggestion
  compareCommits(previousVersion: string, nextVersion: string) {
```

src/app/shared/github/github.service.ts
```diff
@@ -0,0 +1,79 @@
...
+  constructor(private http: HttpClient) {}
+
+  // TODO: GET ALL TAGS
+  getTags(page = 1): Observable<any> {
+    if (this.useMocks) {
+      return of(tags);
+    }
+    const options = {
+      headers: new HttpHeaders({
+        Authorization: 'token ' + environment.oauth.githubToken1 + environment.oauth.githubToken2,
```
mmmtrifork: Consider getting the HttpHeaders from a private function (using the same code at 3 places in this file)

src/app/shared/github/github.service.ts
```diff
@@ -0,0 +1,79 @@
+import { Injectable } from '@angular/core';
+import { HttpClient, HttpHeaders } from '@angular/common/http';
+import { EMPTY, Observable, of } from 'rxjs';
+import { delay, expand, map, toArray } from 'rxjs/operators';
```
mmmtrifork: Remove unused import statements
```suggestion
import { expand, map, toArray } from 'rxjs/operators';
```

src/app/shared/github/github.service.ts
```diff
@@ -0,0 +1,79 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class GithubService {
+  useMocks: boolean = false;
+  constructor(private http: HttpClient) {}
+
+  // TODO: GET ALL TAGS
+  getTags(page = 1): Observable<any> {
+    if (this.useMocks) {
```
mmmtrifork: It feels somewhat weird to have code that handles mocks through production code. Consider using an interface for GithubService, with multiple implementations - 1 for production (this file) and 1 which will deal only with mock data. (It could be just me not seeing the purpose of it)

src/app/changelog/changelog.component.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  ngOnDestroy(): void {
+    this.changelogSubscription.unsubscribe();
+  }
+
+  onRefreshVersion(version) {
+    this.changelogService.updateChangelog(version.name, version.name);
+  }
+
+  trackByFn(index, item) {
```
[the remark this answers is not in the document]
hafstad: I'm not sure I like this change as this makes it unclear what's actually passed to the method. My IDE tells me the variable is unused, so I don't think it's a good practice changing it to something unknown even though it's not used. If you wanna use it later on it's easy to understand what's passed to the method without knowing about the functionality.

src/app/changelog/changelog.component.ts
```diff
@@ -0,0 +1,79 @@
...
+    return item.name;
+  }
+
+  lightOrDark(color) {
```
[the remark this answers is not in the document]
hafstad: Agreed

src/app/shared/github/github.service.ts
hafstad: This is for debugging, as you can quickly call the github api way to much with live-reload. You could properly make a prettier solution for this, but will leave as is for now,

src/app/changelog/changelog.service.ts
```diff
@@ -0,0 +1,141 @@
...
+            };
+          }),
+        };
+      })
+    );
+  }
+
+  private getUnique(arr, comp) {
+    const unique = arr
+      .map((e) => e[comp])
```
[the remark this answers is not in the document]
hafstad: I will leave as is for now, not necessary to use more time on this as performance isn't really a problem here.

Issue #385, opened 2019-06-28
[BUG] Action sheet does not show on IOS


laljyskebank:
**Describe the bug**
When opening an ActionSheet on top of a modal, the ActionSheet is not shown on IOS.

**To Reproduce**
1. Go to 'Faste betalinger'
2. Click on 'FAB'
3. Choose 'Opret betalingsservice aftale'
4. Click on 'Betales fra'
5. Choose 'Egen konto'
6. The ActionSheet is not shown

**Kirby version**
 - [e.g. 0.0.179]

**Expected behavior**
When a user select an account with 2 owners, the ActionSheet schould be shown

**Screenshots**
![image](https://user-images.githubusercontent.com/35138247/60334737-73e75880-999c-11e9-9560-b6f2478af14c.png)

**Desktop (please complete the following information):**
 - OS: []
 - Browser []
 - Version []

**Native (please complete the following information):**
 - Device: [iPhone6]
 - OS: [iOS12.2]

**Additional context**

MonsterGameDev:
As soon as i get out of this sprint

MonsterGameDev:
Arh... go to sleep again wont you....

PR #482, opened 2019-08-04
Fix/multimodal from same view

MonsterGameDev:
Edited Example Text in kirby cookbook about modals. Issue:   [BUG] Action sheet does not show on IOS #385 . reported by lal.
Per Høyer


hafstad:
This pull-request includes a lot of changes which should not be done. I will close this one a create a new one with the intended changes.

PR #522, opened 2019-08-15
Enhancement/Invoke function on view loaded abstraction

mmmtrifork:
Added this abstraction to avoid repeating view.on(loaded) and view.off(loaded) and to prevent memory leaks


mmmtrifork:
Hi @jonasjensen-brolstark 
I like your idea, but unfortunately we only have a web testing set-up and `ViewHelper` is a tns-only utility class.

PR #525, opened 2019-08-16
Enhancement/button refactoring

### LINKED ISSUE #341: [Enhancement] Button sizes
Revisit button sizes and add/update them to the following: 
- [x] Width: Defaults to contract around text + LR padding; But size lg has min-width=220
- [x] Sizes: sm: 32; md: 40 (default size); lg: 48; If button is floating - size is 64

mmmtrifork:
fixes #341

This PR changes buttons completely, as we no longer use the kirby-button directive, instead we use a kirby-button component.


mmmtrifork:
Closing as a better solution was found - using horizontalAlignment = 'center' to collapse the button width

PR #530, opened 2019-08-19
fix(list): fix padding in example

### LINKED ISSUE #524: [BUG] Selectable List Items (native)
I have observed the following on native:
- There is whitespace to the left of list items for both iOS and Android, which should not be there (see screenshot below):
Screenshot 2019-08-15 at 12 48 36
As shown in Zepplin, there should be no whitespace there: 
https://app.zeplin.io/project/5c99d69a79d02b35ea3b8f41/screen/5d022940c297435df258d95c

- There is a weird scrolling behaviour that appears on list item select (Android only). This is not expected to appear.
 
Could be related to #452 , #375

lydemann:
Closes #524
Fix padding in example.
Fixed bug with IOS not repainting after selected item

Now it looks like this:
IOS:
Screenshot 2019-08-27 at 11 58 01

Android:
image


mmmtrifork:
Hm, as far as I understood the design, there should be no padding at all - the icon should be hidden  until we select an item - after that, the icon appears and the respective padding for it is added. Maybe I misunderstood it, can you consult with the UXers and see what they meant ?

lydemann:
> Hm, as far as I understood the design, there should be no padding at all - the icon should be hidden until we select an item - after that, the icon appears and the respective padding for it is added. Maybe I misunderstood it, can you consult with the UXers and see what they meant ?

I think the same, just a funny name of the issue :) I have aligned it with how it looks in the Zeplin mockup. In the end, I guess that is how we want it to look :)

src/kirby/components/list/list.component.ts
```diff
@@ -7,7 +7,10 @@ import {
   OnChanges,
   Output,
   TemplateRef,
+  ViewChild,
 } from '@angular/core';
+import { RadListViewComponent } from 'nativescript-ui-listview/angular/listview-directives';
+import { isIOS } from 'tns-core-modules/platform';
```
jonasjensen-brolstark: Try to run a build for web. And you will see it fails 😄 

You can not import any from {N} specific modules in a shared .ts file. You will have to create a `.tns` version of the component to do this. Or a service with both a `.ts` and a `.tns.ts` version.

src/kirby/components/list/list.component.ts
```diff
@@ -144,6 +147,10 @@ export class ListComponent implements OnChanges {
   onItemSelect(args: any) {
     this.selectedItem = this.listHelper.getSelectedItem(this.items, args);
     this.itemSelect.emit(this.selectedItem);
+
+    if (isIOS) {
+      this.listViewRef.listView.refresh(); // IOS needs this to repaint layout after the selected item has changed
```
jonasjensen-brolstark: You can move this call to `refresh()` into the `listHelper.getSelectedItem` function (or create an other function in it).

On native the `args` argument is of type `ListViewEventData` and `args.object` is of type `RadListView`.

In the helpers `.tns` version you will be able to write

```typescript
  getSelectedItem(items: any[], args: ListViewEventData) {
    let selectedItem: any;
    if (args.index >= 0) {
      selectedItem = items[args.index];
    }
    if (args.object.ios) {
      args.object.refresh();
    }
    return selectedItem;
  }
```

I have not tested this code 😄

src/app/examples/nativescript-only/list-section-example/list-section-example.component.tns.scss
```diff
@@ -1,17 +1,23 @@
...
-  opacity: 0;
+.list-item {
+  padding: 16;
 }
 
+.check {
+  width: 24;
+  margin-right: 8;
+}
+
```
jonasjensen-brolstark: 🔫  pew pew.. whitespace!

src/kirby/components/list/list.component.ts
lydemann: That didn't work because it needs to update after the emit :)

src/kirby/components/list/list.component.ts
lydemann: Yes, will use the platform service instead :)

src/app/examples/nativescript-only/list-section-example/list-section-example.component.tns.scss
lydemann: Done :)

lydemann:
> Hm, as far as I understood the design, there should be no padding at all - the icon should be hidden until we select an item - after that, the icon appears and the respective padding for it is added. Maybe I misunderstood it, can you consult with the UXers and see what they meant ?

Yes, that is how it is implemented now. :)

jonasjensen-brolstark:
I can't see how this is ready for a PR. I dosn't even build. 

Please make sure it does this before opening a PR. Closing for now. I need coffee.

PR #561, opened 2019-08-27
fix(action-sheet): animations

### LINKED ISSUE #556: [BUG] (Native) Backdrop should fade-in and not come from the bottom
**Describe the bug**
Right now the backdrop is 100% opacity and scrolls in from the bottom, which looks really weird, especially compared to the web implementation of the same component. 

**To Reproduce**
1. Open an action-sheet or modal
2. See backdrop behavior especially on iOS

**Kirby version**
 - v0.0.265

**Expected behavior**
We want the backdrop to fade-in like the web-implementation of the action-sheet and modal.

**Screenshots**
Skærmbillede 2019-08-27 kl  09 30 32

**Native (please complete the following information):**

**Additional context**

mmmtrifork:
fixes #556

### DIFF
--- a/src/kirby/components/modal/action-sheet/action-sheet.component.tns.html
+++ b/src/kirby/components/modal/action-sheet/action-sheet.component.tns.html
@@ -1,5 +1,5 @@
-<ModalStack class="action-sheet-wrapperz" dismissEnabled="true" (showingModally)="onShowingActionSheet($event)">
-    <GridLayout rows="*, auto, auto">
+<ModalStack dismissEnabled="true" (showingModally)="onShowingActionSheet($event)">
+    <GridLayout id="action-sheet" rows="*, auto, auto">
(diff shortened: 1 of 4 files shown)


mmmtrifork:
> Looks fines besides some small things:
> 
> When running on android i get the following waring in the log. Can you look into that?
> 
> ```shell
> showModal(view: ViewBase, context: any, closeCallback: Function, fullscreen?: boolean, animated?: boolean, stretched?: boolean) is deprecated. 
> Use showModal(view: ViewBase, modalOptions: ShowModalOptions) instead.
> ```
> 
> I would also like some simple unit tests of the `NativeScriptAnimationHelper` class.

I know about the warning, but unfortunately it seems to be something propagated from the plugin that we're using - nativescript-windowed-modal. I have tried getting rid of it before, but no luck yet. 

I will look into testing the animation helper :)

PR #565, opened 2019-08-28
fix(card): removed min height

### LINKED ISSUE #564: [BUG] Remove Card - Min Height (Web)
**Describe the bug**
On native we have removed the Card min height because it caused some issues, but we forgot to do that on Web.

This makes small cards look (<170dp) look weird.

**To Reproduce**
1. Go to Card on Cookbook

**Screenshots**
Skærmbillede 2019-08-28 kl  09 03 51

mmmtrifork:
fixes #564


bommariusser:
@hafstad its not implemented yet or?

PR #575, opened 2019-09-02
Moved arrow-back and updated readme and webpack

AlbertTuft:


webpack.config.js
```diff
@@ -289,6 +289,7 @@ module.exports = (env) => {
         [
           { from: { glob: 'fonts/**' } },
           { from: { glob: 'kirby/icons/fonts/**' }, to: 'fonts', flatten: true },
+          { from: { glob: 'kirby/icons/img/**' }, to: 'assets/img', flatten: true },
```
jonasjensen-brolstark: You properly not want to have `flatten` set to `true`. Read what is does [here](https://github.com/webpack-contrib/copy-webpack-plugin#flatten).

This part of the docs worries me ` If files have the same name, the result is non-deterministic.`. It could create some weird situations, if someone would create a subfolder in the `img` folder.

PR #582, opened 2019-09-03
feat(list): set list component colors

lydemann:
Set list item colors using a list color call back and a list color directive.
Only supports white and light according to UX specs.

Web:
image

image


src/app/examples/list/list-example.component.html
```diff
@@ -57,6 +57,45 @@ <h2>List with selectable items:</h2>
...
+ <hr />
+
+ <h2>List with sections and colored items:</h2>
+  <kirby-list
+    [items]="items"
+    (itemSelect)="onItemSelect($event)"
+    [getSectionName]="getSectionName"
+    [itemColorCb]="itemColorCb"
+    [showDivider]="true">
+  <!-- kirby-list-header is not supported in conjunction with *kirbyListSectionHeader due to a bug in RadListView on iOS -->
```
jonasjensen-brolstark: This comment is no longer relevant. Please delete it. And if you copied from somewhere, please also delete it there.

src/kirby/components/list/list-item-color.directive.ts
```diff
@@ -0,0 +1,22 @@
+import { Directive, Input, Renderer2, ElementRef, OnChanges, SimpleChanges } from '@angular/core';
+
+@Directive({
+  selector: '[kirbyListItemColor]',
+})
+export class ListItemColorDirective implements OnChanges {
+  @Input()
+  colorCb: (item: any) => string;
```
jonasjensen-brolstark: Lets talk about naming 😄 

We have no where in codebase used this naming scheme. I think most (if not all) places we name like `getSectionName`. I don't have any strong feeling on the name it self. But i feel like we should align the naming.

This goes for all the places you use `colorCb`.

src/kirby/components/list/list-item/list-item.component.html
```diff
@@ -1,6 +1,7 @@
-    <div class="title-container">
-        <div class="title">{{title}}</div>
-        <div class="subTitle">{{subTitle}}</div>
-    </div>
-    <div class="details-container">{{detail}}</div>
-    <div class="amount-container">{{amount}}</div>
+<div class="title-container">
```
jonasjensen-brolstark: Good to see som boy scout work done 👍

src/kirby/components/list/list.component.html
```diff
@@ -31,13 +34,16 @@
   </ng-container>
 
   <ng-template #normalList>
-    <ion-item *ngFor="let item of items;trackBy: trackByFn"
+    <ion-item *ngFor="let item of items; let idx = index; trackBy: trackByFn"
```
jonasjensen-brolstark: Is `idx` used anywhere? if not then remove it.

src/kirby/components/list/list.component.scss
```diff
@@ -36,8 +36,20 @@ $shadow: get-elevation(2);
       margin-bottom: $outline-width;
       background-color: $background-color-hover;
     }
-  }
 
+    $list-colors: ('white', 'light');
```
jonasjensen-brolstark: Is there an other way to get these two colors then this? I can see an issue if somewhere down the road the names `white` or `light` changes.

src/kirby/components/modal/action-sheet/action-sheet.component.spec.ts
```diff
@@ -17,6 +17,7 @@ import { InfiniteScrollDirective } from '../../list/directives/infinite-scroll.d
 import { CardHeaderComponent } from '../../card/card-header/card-header.component';
 import { ListFlexItemComponent } from '../../list/list-flex-item/list-flex-item.component';
 import { ButtonComponent } from '../../button/button.component';
+import { ThemeColorDirective } from '@kirbydesign/designsystem/directives/theme-color/theme-color.directive';
```
jonasjensen-brolstark: Why is this import needed?

And why are there changes in this file?

src/kirby/directives/theme-color/theme-color.directive.ts
```diff
@@ -8,7 +8,8 @@ import { ColorHelper } from '../../helpers/color-helper';
   selector: `kirby-card[themeColor],
              kirby-badge[themeColor],
              kirby-icon[themeColor],
-             kirby-toolbar[themeColor]`,
+             kirby-toolbar[themeColor],
+             ion-item[themeColor]`,
```
jonasjensen-brolstark: Is this a leftover?

I don't see you use the `themeColor` directive on an `ion-item`?

src/kirby/components/list/list-item-color.directive.ts
```diff
@@ -0,0 +1,22 @@
+import { Directive, Input, Renderer2, ElementRef, OnChanges, SimpleChanges } from '@angular/core';
```
jonasjensen-brolstark: There is a `directive` subfolder, in the list component folder. Please move it to there.

src/kirby/components/list/list.component.scss
```diff
@@ -36,8 +36,20 @@ $shadow: get-elevation(2);
...
+    @each $color-name, $color-value in $list-colors {
+       &.#{$color-name} {
+               background-color: #{get-color($color-name)};
```
jakobe: Don't we also need to add the text color?
`color:  #{get-color($color-name+'-contrast')};`

jonasjensen-brolstark:
There is something wrong with hover. In the example only the first item on the list have hover background. No other of the items have. Please look into that.

And on last major issue. The unit tests doesn't run. Please fix that.

src/kirby/components/list/list-item-color.directive.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  constructor(private renderer: Renderer2, private elementRef: ElementRef) {}
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (!this.colorCb || !this.item) {
+      return;
+    }
+
+    const color = this.colorCb(this.item);
+    this.renderer.addClass(this.elementRef.nativeElement, color);
```
jakobe: If the `color` is used as a class the name is confusing. Should be `themeColor` or something like that...

src/kirby/components/list/list-item/list-item.component.html
jakobe: True - on the other hand, this file has nothing to do with issue/PR so the change pollutes the version control history...

src/app/examples/list/base-list.component.ts
```diff
@@ -6,90 +6,103 @@ export class BaseListComponent {
...
       detail: -135,
+      color: 'white',
     },
     {
       id: 10,
       title: 'Rudolph Technologies, Inc.',
       subTitle: '1799 pcs',
       amount: '18.442 DKK',
       detail: 86,
+      color: 'kirby-color-brightness-white',
```
jakobe: ??

src/app/examples/list/base-list.component.ts
```diff
@@ -6,90 +6,103 @@ export class BaseListComponent {
...
     },
     {
       id: 11,
       title: 'Atlas Financial Holdings, Inc.',
       subTitle: '129 pcs',
       amount: '75.842 DKK',
       detail: -43,
+      color: 'bg-kirby-color-brightness-ligh',
```
jakobe: ??

src/app/examples/list/base-list.component.ts
```diff
@@ -6,90 +6,103 @@ export class BaseListComponent {
...
     },
     {
       id: 12,
       title: 'Five Star Senior Living Inc.',
       subTitle: '36 pcs',
       amount: '99.125 DKK',
       detail: -144,
+      color: 'kirby-color-brightness-dark',
```
jakobe: ??

src/kirby/components/list/list-item-color.directive.ts
```diff
@@ -0,0 +1,22 @@
+import { Directive, Input, Renderer2, ElementRef, OnChanges, SimpleChanges } from '@angular/core';
+
+@Directive({
+  selector: '[kirbyListItemColor]',
+})
+export class ListItemColorDirective implements OnChanges {
```
jakobe: Could this be merged with the existing themeColor directive? E.g. `color: (item: any) => string | string;`

src/kirby/components/list/list-item-color.directive.ts
jakobe: Consider aliasing this with the directive name, so it can be applied like:
`kirbyListItemColor="myColorCallback"`

src/kirby/components/list/list-item-color.directive.ts
jakobe: Why not just use `@HostBinding` instead of Renderer2?

src/app/examples/list/list-example.component.ts
```diff
@@ -9,6 +9,8 @@ import { BaseListComponent } from './base-list.component';
   styleUrls: ['./list-example.component.scss'],
 })
 export class ListExampleComponent extends BaseListComponent {
+  itemColorCb = (item) => item.color;
```
jakobe: Consider naming....

src/kirby/components/list/list-item/list-item.component.html
lydemann: I think it is fair to do a little boy scout work/refactoring without it is polluting the commit too much. There is a balance.

src/kirby/components/list/list.component.scss
lydemann: Well, I could get it directly from the $system-colors map, which will add some type safety to this, but might make up for some weird code in a scss file...

src/kirby/components/modal/action-sheet/action-sheet.component.spec.ts
lydemann: Needed to not break the integration test here

src/kirby/directives/theme-color/theme-color.directive.ts
lydemann: It was, thanks!

src/kirby/components/list/list-item-color.directive.ts
lydemann: Added it to directives folder

src/kirby/components/list/list.component.scss
lydemann: Yes (y)

src/app/examples/list/base-list.component.ts
lydemann: Just to show this is not being rendered yet

src/app/examples/list/base-list.component.ts
lydemann: Removed these

src/kirby/components/list/list-item-color.directive.ts
lydemann: No I don't think so. They do different things. This calls the getItemColor cb and sets class and the other just maps a color input to a class to append to the host element.

src/kirby/components/list/list-item-color.directive.ts
lydemann: Done.

src/kirby/components/list/list-item-color.directive.ts
lydemann: Renderer2 is less code and is more convenient when we want to set an unknown class than hostBinding.

src/app/examples/list/list-example.component.ts
lydemann: Done

src/app/examples/list/base-list.component.ts
lydemann: Removed

src/app/examples/list/base-list.component.ts
lydemann: Removed

src/kirby/components/list/list-item-color.directive.ts
lydemann: Renamed

src/kirby/components/list/list-item-color.directive.ts
jonasjensen-brolstark: The plus side by merging it into the `ThemeColor` directive would be, that all handling of colors would be in one place. And the cool thing with the `ThemeColor` directive is, that the component has to handle what colors it's supports in its own styling.

So in the case of the list, it would have in its `scss` file:

```scss
:host(.white) {
...
}

:host(.light) {
...
}
```

And as the `item` also is user defined, i see no problem in merging this into the `ThemeColor` directive. So please reconsider it 😄

src/kirby/components/modal/action-sheet/action-sheet.component.spec.ts
```diff
@@ -55,6 +56,7 @@ describe('ActionSheetComponent', () => {
         SpinnerComponent,
         InfiniteScrollDirective,
         ButtonComponent,
+        ListItemColorDirective,
```
jonasjensen-brolstark: Could you please use `MockDirective` from `ng-mocks`.
```suggestion
        MockDirective(ListItemColorDirective),
```

Then we also do real unit testing 🚀

src/kirby/components/list/list-item-color.directive.ts
lydemann: But the color cb is only used in the list and the themeColor directive is used in multiple different places. Don't we pollute the themeColor directive by adding list specific logic to it? I know the cb is user defined, but still, it is created for the list.

src/kirby/components/modal/action-sheet/action-sheet.component.spec.ts
lydemann: But do we really want that here? All the other dependencies are included without mock. Maybe it also makes sense to keep this as much of an integration test as possible

src/kirby/components/list/directives/list-item-color.directive.ts
```diff
@@ -0,0 +1,22 @@
+import { Directive, Input, Renderer2, ElementRef, OnChanges, SimpleChanges } from '@angular/core';
+
+@Directive({
+  selector: '[kirbyListItemColor]',
+})
+export class ListItemColorDirective implements OnChanges {
+  @Input()
+  kirbyListItemColor: (item: any) => string;
```
jonasjensen-brolstark: Set `themeColor` type

PR #584, opened 2019-09-03
Enhancement/swipe actions on kirby list

### LINKED ISSUE #357: [Enhancement] Add swipe actions to Kirby list (mobile web)
- It should be possible to define actions on a list item that is revealed when swiping the item. Both left and right swipe/placement should be supported.
- It should be possible to enable a "long swipe" (swiping the item over a certain threshold). Both left and right long swipe should be supported.
- If there is more than 1 action it's always the 1st action (the action closest to the side of the list) that gets triggered on long swipe (e.g. "mark" in the example screenshot below).
- It should be possible to set the background color of the revealed actions (from the Kirby progress colors: success, warning, danger).
- It should be possible to set the text and (optional) icon for each action.
- It should be possible to "switch" between different actions depending on the state of the data for each item (e.g. "Archive" vs. "Remove archive")

See:
Use case: https://app.zeplin.io/project/5c99d69a79d02b35ea3b8f41/screen/5d0770d6ea9d1b5d986e60b1
Web: https://ionicframework.com/docs/api/item-sliding

jonasjensen-brolstark:
Closes #357 
It is now possible to define swipe actions on the `kirby-list`.

A swipe action model is defined like this:

```typescript
export interface ListSwipeAction {
  position: 'left' | 'right';
  title: ((item: any) => string) | string;
  iconName?: ((item: any) => string) | string;
  themeColor?: ((item: any) => ThemeColor) | ThemeColor;
  onSelected: (item: any) => void;
  isDisabled?: ((item: any) => boolean) | boolean;
}
```

By having have the properties as union types of a callback and a simple type. This should enable us to support most of the cases described in issue #357.

There is maybe some optimization in the way swipe actions are rendered right now. But lets push that ahead of us.
 
image


src/kirby/components/list/list.component.html
```diff
@@ -14,38 +16,42 @@
...
-                  [class.last-in-section]="idx === section.items.length - 1"
-                  (click)="onItemSelect(item)"
-                  tabindex="0"
-                  [kirbyListItemColor]="getItemColor"
-                  [item]="item"
-                  keyHandler>
-          <ng-container *ngTemplateOutlet="listItemTemplate; context: { $implicit: item }"></ng-container>
-          <ng-container *ngTemplateOutlet="listFlexItemTemplate; context: { $implicit: item }"></ng-container>
-        </ion-item>
+        <ion-item-sliding *ngFor="let item of section.items; let idx = index;trackBy: trackByFn" [disabled]="isSwipingDisabled">
```
jakobe: Should we consider performance regarding always wrapping list items in an item-sliding wrapper?
Maybe moving the content of this tag to a template and conditionally wrapping in `ion-item-sliding`:
(We could do this later as part of a general performance overhaul on the list...)
Ohhh - I see we are already duplicating the `ion-item` 😱  I think NOW would be a good time to DRY that up....

src/kirby/components/list/list.component.html
```diff
@@ -14,38 +16,42 @@
...
+          <ng-container *ngTemplateOutlet="swipeActions; context: { $implicit: item }"></ng-container>
+          <ion-item [class.has-divider]="showDivider"
+                    [class.selectable]="isSelectable"
+                    [class.selected]="markSelectedRow && selectedItem && item._id === selectedItem._id"
+                    [attr.role]="isSelectable ? 'button' : null"
+                    [class.last-in-section]="idx === section.items.length - 1"
```
jakobe: Not related to this issue/PR 😄  But can be deleted:
```suggestion
```

src/kirby/components/list/list.component.ts
```diff
@@ -119,22 +139,6 @@ export class ListComponent implements OnChanges {
...
-    }
-    const order = this.orderMap.get(item);
-    if (!order) {
-      console.warn('Order of list item within section not found!');
-      return defaultOrder;
-    }
-    return order;
-  }
-
   isFirstItem(item: any, index: number) {
```
jakobe: Not related to this issue/PR 😄  But `isFirstItem` + `isLastItem` + `getItemOrder` + `orderMap` can be deleted:
```suggestion
```

src/kirby/components/list/list.component.ts
```diff
@@ -172,4 +220,30 @@ export class ListComponent implements OnChanges {
...
+    const order = this.orderMap.get(item);
+    if (!order) {
+      console.warn('Order of list item within section not found!');
+      return defaultOrder;
+    }
+    return order;
+  }
+
+  private initialzeSwipeActions(): void {
+    const large = style.global['$breakpoints'].value['large'].value;
```
jakobe: Consider leaving a `TODO` comment here, I believe we can find a better way to get the breakpoint...

src/kirby/components/list/list.component.html
```diff
@@ -14,38 +16,19 @@
...
-                  [class.last-in-section]="idx === section.items.length - 1"
-                  (click)="onItemSelect(item)"
-                  tabindex="0"
-                  [kirbyListItemColor]="getItemColor"
-                  [item]="item"
-                  keyHandler>
-          <ng-container *ngTemplateOutlet="listItemTemplate; context: { $implicit: item }"></ng-container>
-          <ng-container *ngTemplateOutlet="listFlexItemTemplate; context: { $implicit: item }"></ng-container>
-        </ion-item>
+        <ion-item-sliding *ngFor="let item of section.items; let idx = index;trackBy: trackByFn" [disabled]="isSwipingDisabled">
```
jakobe: I think now that we've moved the item html into it's own template we should do the same regarding `<ion-item-sliding>` wrapper / plain items:
```suggestion
          <ng-container *ngTemplateOutlet="hasSwipeActions ? itemSlidingTemplate : itemTemplate; context: { $implicit: item }"></ng-container>
```

src/kirby/components/list/list-swipe-action.ts
```diff
@@ -0,0 +1,10 @@
+type ActionType = 'success' | 'warning' | 'danger';
```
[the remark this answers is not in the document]
jonasjensen-brolstark: Do i? It is a type i only use in the `ListSwipeAction` interface.

When creating an object with that interface (like in the examples)  Typescripts type inference when only let the property `type` have the string values  success, warning or danger.

So short answer no 😄

src/kirby/components/list/list.component.html
```diff
@@ -56,4 +33,71 @@
...
+    <ng-container *ngTemplateOutlet="listItemTemplate; context: { $implicit: item }"></ng-container>
+    <ng-container *ngTemplateOutlet="listFlexItemTemplate; context: { $implicit: item }"></ng-container>
+  </ion-item>
+</ng-template>
+
+
+<!-- Swipe Action templates -->
+
+<ng-template #slidingItems let-items>
+  <ion-item-sliding *ngFor="let item of items; let idx = index;trackBy: trackByFn">
```
jakobe:
```suggestion
  <ion-item-sliding *ngFor="let item of items ;trackBy: trackByFn">
```

src/kirby/components/list/list.component.html
```diff
@@ -56,4 +33,71 @@
 <p *ngIf="!isLoadOnDemandEnabled && noMoreItemsText" class="no-more-items">{{ noMoreItemsText }}</p>
 <div *ngIf="isLoading" class="loading">
   <kirby-spinner></kirby-spinner>
-</div>
\ No newline at end of file
+</div>
+
+<!-- Item template -->
+<ng-template #listItems let-items>
+  <ng-container *ngFor="let item of items;trackBy: trackByFn">
```
jakobe:
```suggestion
  <ng-container *ngFor="let item of items; trackBy: trackByFn">
```

PR #606, opened 2019-09-16
optional ViewContainerRef and RxJS-operators modal support added

askarby:
:children_crossing: Since ViewContainerRef was not needed anymore, passing it to the ModalController is now optional (will be removed in later version of Kirby). Made operators (to be used with RxJs) to use modals, alerts and action sheets in a reactive manor.

### DIFF
--- a/src/app/examples/alert-example/alert-example.component.ts
+++ b/src/app/examples/alert-example/alert-example.component.ts
@@ -2,25 +2,41 @@ import { Component, ViewContainerRef } from '@angular/core';
 
 import { ModalController } from '@kirbydesign/designsystem/modal';
 import { AlertConfig } from '@kirbydesign/designsystem/modal';
+import { of } from 'rxjs';
+import { delay, map, tap } from 'rxjs/operators';
 
 @Component({
   selector: 'kirby-alert-example',
   templateUrl: './alert-example.component.html',
 })
 export class AlertExampleComponent {
+  private config: AlertConfig = {
+    title: 'Your alert',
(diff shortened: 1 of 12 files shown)


src/app/examples/alert-example/alert-example.component.ts
```diff
@@ -2,25 +2,41 @@ import { Component, ViewContainerRef } from '@angular/core';
 
 import { ModalController } from '@kirbydesign/designsystem/modal';
 import { AlertConfig } from '@kirbydesign/designsystem/modal';
+import { of } from 'rxjs';
```
jonasjensen-brolstark: The linter is complaining about these imports. They need to be at top as they are 3. party.

askarby: Well, that was embarrassing... fixed!

hafstad:
Also please label your pull-req according to which component you're changing, and mark it with "breaking change" if thats going to be the case. In that way our changelog will affect that when it's merged :)
https://cookbook.kirby.design/home/changelog

src/app/examples/alert-example/alert-example.component.ts
askarby: In regards to the "2 second delay", I wanted to show the "cheapest possible solution" to having a RxJs Observable do something.... granted.... I could simply make it wait 10ms... that way the impact wouldn't be annoying.

askarby:
Closing this in favor of: https://github.com/kirbydesign/designsystem/pull/610

PR #611, opened 2019-09-16
Fix/list styling

### LINKED ISSUE #507: [BUG] List with header and sections - rounded corners (web)
**Describe the bug**
If you have a list with list-header and sections the current styles wont make the first list item with rounded corners.

see screenshot

**Expected behavior**
The first item should have rounded corners even though it has a list header. Current styles of the list on web should just be modified. Right now it's actually excluded with :not(.has-header) which is wrong

**Screenshots**
image

**Additional context**
This only affects web and is caused by the ionic change #102 where the styles and implementation of the list was changed

jakobe:
Fixes #507

### DIFF
--- a/src/kirby/components/list/list.component.scss
+++ b/src/kirby/components/list/list.component.scss
@@ -4,60 +4,84 @@ $divider-color: get-color('background-color');
 $background-color-hover: get-color('background-color');
 $shadow: get-elevation(2);
 
-:host {
-  ion-list {
-    box-shadow: $shadow;
-    background: transparent;
-    contain: inherit;
-    padding: 0;
+ion-list {
+  box-shadow: $shadow;
+  background: transparent;
+  contain: inherit;
+  padding: 0;
+}
+
+ion-item {
+  background-color: get-color('white');
+  &:not(:last-of-type).has-divider {
+    border-bottom: 1px solid $divider-color;
   }
 
-  ion-item {
-    background-color: get-color('white');
-    &:not(:last-of-type).has-divider {
-      border-bottom: 1px solid $divider-color;
-    }
+  &.selected ::ng-deep * {
+    font-weight: font-weight('bold');
+  }
+  &:focus {
+    $outline-width: 5px;
+    outline-color: var(--highlight-color-focused);
+    outline-style: auto;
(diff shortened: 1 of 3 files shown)


src/kirby/components/list/list.component.scss
```diff
@@ -3,61 +3,90 @@
...
 
-  ion-item {
-    background-color: get-color('white');
-    &:not(:last-of-type).has-divider {
-      border-bottom: 1px solid $divider-color;
-    }
+  &.selected ::ng-deep * {
+    font-weight: font-weight('bold');
+  }
+  &:focus {
```
jonasjensen-brolstark: There is something wrong with the focus outline:

image

And is it on on purpose outline is always shown and not just when the item is selectable?

src/kirby/components/list/list.component.scss
```diff
@@ -3,61 +3,90 @@
...
-    &.selectable {
-      cursor: pointer;
-      &:hover {
-        background-color: $background-color-hover;
-      }
-    }
+  $list-colors: ('light'); // add supported list item theme colors here
+  @each $color-name, $color-value in $list-colors {
+     &.#{$color-name} {
+             background-color: #{get-color($color-name)};
```
jonasjensen-brolstark: There is some indentation here that is off.

src/kirby/components/list/list.component.scss
```diff
@@ -3,61 +3,90 @@
...
+}
+
+.footer {
+  background-color: get-color('white');
+  border-top: 1px solid $divider-color;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  width: 100%;
+
```
jonasjensen-brolstark: 🔫 pew pew, whitespace. (🔫   ⬅️  my whitespace gun)

src/kirby/components/list/list.component.scss
jakobe: Don't think this is related to the PR...

src/kirby/components/list/list.component.scss
jakobe: Not related to this PR, but fixed :)

Issue #504, opened 2019-08-08
[ENHANCEMENT] No inherent padding on kirby-list-item


PetarRing:
**Is your enhancement request related to a problem? Please describe.**
Since padding is inherently sat on kirby-list-item, it makes templates which would span the entire list width an impossibility.

**Describe the solution you'd like**
Removing the padding on kirby-list-item so the ability to create templates for the kirby-list-item which would span the entire width becomes available. The images below show, what is intended and what is reality.

**Describe alternatives you've considered**
The ability to disable padding on the list via input.

**Additional context**
![image](https://user-images.githubusercontent.com/8911645/62697884-0e937800-b9dc-11e9-9844-0e44654f6d66.png)
![image](https://user-images.githubusercontent.com/8911645/62697912-1ce19400-b9dc-11e9-8569-7b0df1fb6ac6.png)

darpankumar:
I am also facing some issue with list padding attached the screen shot for the same.

jb-roa:
Fixed by #612

PR #612, opened 2019-09-17
Enhancement/no shape lists

jb-roa:


src/app/examples/list/no-shape/list-no-shape-example.component.scss
```diff
@@ -0,0 +1,34 @@
+@import "../../../../kirby/scss/utils";
+
+kirby-card {
+  margin: 5px;
```
jakobe: Consider extracting shadow units (offset-x | offset-y | blur-radius | spread-radius) from `src\kirby\scss\base\_functions.scss` into separate scss variables and using them instead.

src/app/examples/list/no-shape/list-no-shape-example.component.scss
```diff
@@ -0,0 +1,34 @@
...
+  width: 330px;
```
jakobe: Consider calculating from `$breakpoints` map in `src\kirby\scss\base\_variables.scss`

src/app/examples/list/no-shape/list-no-shape-example.component.html
```diff
@@ -0,0 +1,26 @@
+<h2>List with no background shape:</h2>
+<h3>Removes background, border, shadow and padding from the list and the list item containers to provide flexibility
+</h3>
+<kirby-list [items]="items" shape="none" hasItemSpacing="true">
+
+  <kirby-card *kirbyListItem="let item">
+    <kirby-card-header *ngIf="item.amount < 0" title="Account is overdraft"></kirby-card-header>
+    <kirby-list-flex-item hasPadding="true">
```
jakobe: `hasPadding` property doesn't exist on `kirby-list-flex-item` - move up to `kirby-card` element

src/app/examples/list/no-shape/list-no-shape-example.component.scss
```diff
@@ -0,0 +1,34 @@
...
+
+  @include media(">=large") {
+      width: 500px;
+  }
+
+  kirby-card-header {
+      padding: size('xxxs');
```
jonasjensen-brolstark: Nitpicking: You have 4 spaces per tap here but 2 spaces other places in the file.
And all the other code uses 2 spaces. 😃

src/kirby/components/list/list.component.ts
```diff
@@ -24,7 +24,9 @@ import { GroupByPipe } from './pipes/group-by.pipe';
 import { ListSwipeAction } from './list-swipe-action';
 import { ThemeColor } from '@kirbydesign/designsystem/helpers/theme-color.type';
 
-export type ListShape = 'square' | 'rounded';
+export type ListShape = 'square' | 'rounded' | 'none';
+
+export type ItemSpacing = 'true' | 'false';
```
jonasjensen-brolstark: What is the purpose of making `ItemSpacing` a string based type over a boolean?

jb-roa: Well.. 
1: hasItemSpacing="false" is resulting in a string value, i.e.: 'false' is truthy.
And 2: We hope to extend this later, with values such as "compact" or "spacey"

src/app/examples/list/no-shape/list-no-shape-example.component.scss
jb-roa: Doh.. my VS Code stopped auto formatting for some reason. Will fix :-)

src/app/examples/list/no-shape/list-no-shape-example.component.scss
```diff
@@ -0,0 +1,34 @@
+@import "../../../../kirby/scss/utils";
+
+kirby-card {
+  margin: map-get($shadow-offset-y, 2);
```
jb-roa: @jakobe I'm using the variables as requested now. Luckily, the shadow-offset can be used here, but in reality the margin should be based on a calculation I guess, depending on what side the margin is on.
However I would like this not to be in scope of this PR, as I don't see it as a trivial thing to implement.

src/kirby/scss/base/_variables.scss
```diff
@@ -63,3 +63,29 @@ $breakpoints: (
   xlarge: 60em, 
   xxlarge: 80em 
 ) !default;
+
+/* Shadow properties
```
jb-roa: @jakobe Please le me know if this is what you were looking for.

src/app/examples/list/no-shape/list-no-shape-example.component.scss
jb-roa: Fixed manually. My Prettier is still broken for some reason.

src/app/examples/list/no-shape/list-no-shape-example.component.html
jb-roa: Removed

src/app/examples/list/no-shape/list-no-shape-example.component.scss
jb-roa: Now using map-get($breakpoints, ...) instead. Is this what you meant?

src/app/examples/list/no-shape/list-no-shape-example.component.scss
jb-roa: Extracted shadow variables. However calculating the margin based on them is actually not that trivial after all as I see it. Something like:
top = (spread - v_shadow + 0.5*blur)
right = (spread + h_shadow + 0.5*blur)
bottom = (spread + v_shadow + 0.5*blur)
left = (spread - h_shadow + 0.5*blur)

I would prefer it not to be in scope of this PR.

src/kirby/components/list/list.component.ts
jonasjensen-brolstark: 1. I actually didn't know. Well today i learned.
2. Fine.

jakobe: We don't do this in any other components, so I'm reverting to boolean and closing this conversation

jb-roa:
Fixes #504

PR #628, opened 2019-09-24
Feature/form field

### LINKED ISSUE #181: [COMPONENT] Input Field
Inputs provides a way for users to enter data into your solution.

input , textarea etc.

Zeplin: https://zpl.io/agkRvkD

Bitmap

**Additional context**
Example: 
https://ionicframework.com/docs/api/input

### Kick Off:
- [x] Accept Feature incl. name of component (DSL)
- [ ] Update Kirby Component Status and publish to [Kirby Cookbook](https://cookbook.kirby.design)
### Code:
- [x] Create Feature Branch
- [x] Implement Skeleton Component:
    - [x] Web implementation
    - [ ] {N} implementation
- [x] [Implement Examples and Showcase](https://cookbook.kirby.design/home/showcase/button)
- [ ] Add link to showcase on Kirby Component Status (`component-status-items.ts => code.cookbookUrl`)
- [ ] Push commits to GitHub
### Review:
- [ ] Open a pull request (remember to add `closes #issueno` to the description) and ask @kirbydesign/kirby-guild for review
- Merge to master and delete branch

jonasjensen-brolstark:
Closes #181 

Looking at the PR my self, i can maybe see that `input` and `textarea` should be separate components, with their own example pages.

I have made so that the `FormFieldMessageComponent` can not be used other places, then inside a `FormFieldComponent` as i don't see use for it anywhere else. Should we do the same for the `input` and `textarea` components? 

There is a quirk i simply could not find a solution for. `FormFieldMessageComponent` has a `position` input property. I use this in a select for the `ng-conent` tags in `FormFieldComponent`. This does that the `position` property can not be dynamically bound and has to be static. If anyone has a good solution for this, i am open for suggestions.

image


src/kirby/components/form-field/_form-field-inputs.shared.scss
```diff
@@ -0,0 +1,55 @@
...
+  border-width: 1px;
+  box-shadow: get-elevation('2');
+  box-sizing: border-box;
+  color: get-color('black');
+  display: block;
+  font-size: font-size('n');
+  line-height: line-height('n');
+  outline: none;
+  padding: size('s');
+  width: 100%;
```
jb-trm: Suggestion: 
Add resize: none
Gets rid of that ugly resize arrow - I have no idea why it's not an opt in.

src/kirby/components/form-field/form-field.component.scss
```diff
@@ -0,0 +1,39 @@
...
+  }
+}
+
+label {
+  display: block;
+
+  span {
+    font-size: font-size('s');
+    font-weight: font-weight('light');
+    margin-bottom: size('xxxs');
```
jonasjensen-brolstark: @jakobe A floating element can not have margin. So either we have to make this span `display:block` or move the margin to the input elements.

I vote for making the span `display: block`. But i can not remember if we talked about the reason of the span not being `display: block` yestoday?

lydemann: If you need it to be `display: block`, it probably should be a `div` instead as this might cause confusion if you change the basic behavior of `span`.
Btw, how is this a floating element?

src/kirby/components/form-field/input/input.component.ts
```diff
@@ -0,0 +1,32 @@
...
+    search: 'search',
+  };
+  @Input() set type(value: string) {
+    this._inputmode = InputComponent.typeToInputmodeMap[value];
+  }
+  @HostBinding('attr.inputmode')
+  _inputmode: string;
+
+  @HostBinding('class.state-success')
+  _isSuccess: boolean;
```
lydemann: Do we use _ prefix for public fields?

src/kirby/components/form-field/input/input.component.ts
```diff
@@ -0,0 +1,32 @@
...
+export class InputComponent {
+  private static typeToInputmodeMap = {
+    number: 'decimal',
+    search: 'search',
+  };
+  @Input() set type(value: string) {
+    this._inputmode = InputComponent.typeToInputmodeMap[value];
+  }
+  @HostBinding('attr.inputmode')
+  _inputmode: string;
```
lydemann: Do we use _ prefix for public fields?

src/kirby/components/form-field/textarea/textarea.component.html
```diff
@@ -0,0 +1,2 @@
+<ng-content *ngIf="!value"></ng-content>
```
lydemann: Could reference the `templateRef` of the `contentChild` instead, as this will not synchronize the life cycle hooks between this component and the provided textarea component. With ng-content, the host has no control over the transcluded content (eg, `onDestroy` hook will not trigger on transcluded component when the host gets destroyed - maybe not what we want here.

For more info in this, check this article by the author of Spectator and Akita Netanel Netanel Basal [here, read from "Supporting Lazy Content"](https://netbasal.com/creating-behavioral-components-in-angular-8db6080f0cc4)

src/kirby/components/form-field/textarea/textarea.component.ts
```diff
@@ -0,0 +1,23 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  // tslint:disable-next-line:component-selector
+  selector: 'textarea[kirby-textarea]',
+  styleUrls: ['./textarea.component.scss'],
+  templateUrl: './textarea.component.html',
+})
+export class TextareaComponent {
+  @Input() value: string;
+  @HostBinding('class.state-success')
+  _isSuccess: boolean;
```
lydemann: Again, do we use _ prefix for public props?

src/kirby/components/form-field/textarea/textarea.component.ts
```diff
@@ -0,0 +1,23 @@
...
+  @HostBinding('class.state-warning')
+  _isWarning: boolean;
+  @HostBinding('class.state-danger')
+  _isDanger: boolean;
+  @Input() set state(state: 'success' | 'warning' | 'danger') {
```
lydemann: I see the `'success' | 'warning' | 'danger` is duplicated in two places. Could we extract this as a shared type to reference in both places?

src/kirby/components/form-field/form-field.component.scss
jonasjensen-brolstark: I meant inline element. Sorry my mistake 😄

src/kirby/components/form-field/form-field.component.html
```diff
@@ -0,0 +1,21 @@
...
+<ng-container *ngIf="!label">
+  <ng-container *ngTemplateOutlet="slottedInput"></ng-container>
+</ng-container>
+<div *ngIf="messages && messages.length > 0" class="texts">  
+  <div class="right">
+    <ng-content select="kirby-form-field-message[position=right]"></ng-content>
+  </div>
+  <div class="left">
+    <ng-content select="kirby-form-field-message[position=left]"></ng-content>
+    <ng-content select="kirby-form-field-message"></ng-content>
```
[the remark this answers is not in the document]
jonasjensen-brolstark: The last one i a catch-all `ng-content`. If the user dos not define a `position` it will end up i the last `ng-content`.

src/kirby/components/form-field/input/input.component.ts
jonasjensen-brolstark: I think it is because, it is an internally used public field.

src/kirby/components/form-field/_form-field-inputs.shared.scss
```diff
@@ -0,0 +1,40 @@
...
+  &:disabled {
+    background-color: get-color('light-tint');
+  }
+
+  &::placeholder {
+    color: get-color('semi-dark');
+  }
+
+  // for the browsers that has partial support, see https://caniuse.com/#feat=css-placeholder
+  &::-webkit-input-placeholder,
```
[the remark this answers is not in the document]
jonasjensen-brolstark: As of right now we have not configured autoprefix. Maybe we should look in to this. But in an other PR.

src/app/examples/form-field-example/form-field-example.component.scss
```diff
@@ -0,0 +1,15 @@
+@import "../../../kirby/scss/utils";
```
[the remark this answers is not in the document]
jonasjensen-brolstark: Correct me if i am wrong here. 
But is the problem not, that our cookbook app is mixed in with our "libary" code for kirby. 
So if we enable non relative paths in the cookbook, we risk that none relative paths slips out in the kirby code. 

And that would break it for the users of kirby?

src/app/examples/form-field-example/examples/input/state-danger.ts
```diff
@@ -0,0 +1,14 @@
+import { Component } from '@angular/core';
+
+let config = {
```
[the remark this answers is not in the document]
jonasjensen-brolstark: Agree.

src/app/examples/examples.common.ts
```diff
@@ -78,6 +79,7 @@ export const COMPONENT_DECLARATIONS: any[] = [
   EmptyStateExampleComponent,
   ToolbarExampleComponent,
   FabSheetExampleComponent,
+  FormFieldExampleComponent,
```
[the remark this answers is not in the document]
jonasjensen-brolstark: Yes we need to rework the cookbook.

src/kirby/components/form-field/input-counter/input-counter.component.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { InputComponent } from '../input/input.component';
+
+@Component({
+  selector: 'kirby-input-counter',
+  templateUrl: './input-counter.component.html',
+  styleUrls: ['./input-counter.component.scss'],
+})
+export class InputCounterComponent implements OnInit, OnDestroy {
+  @Input() listenTo: InputComponent;
```
jonasjensen-brolstark: We do not want the counter on textareas ?

src/kirby/components/form-field/input/input.component.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  @HostBinding('attr.maxlength')
+  @Input()
+  maxlength: number;
+
+  @HostBinding('attr.inputmode')
+  private _inputmode: string;
+
+  @HostListener('keyup', ['$event.target.value'])
+  private _onKeyUp(value: string) {
```
jonasjensen-brolstark: Is there a reason you not just using the `oninput` event? see [here](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/oninput).

It also is triggered by cut/paste.

PR #630, opened 2019-09-24
Fix/gh 618 button with icon and text

### LINKED ISSUE #618: [Enhancement] Button with icon and text renders as icon only button
**Describe the bug**
A button that has both icons and text inside itself is rendered as a icon only button, which is round. Right now we say which button it is, like "icon only" and if we put text inside it, it gets bugged.

**To Reproduce**
1. Go to "Get Started" page 
2. See button "Become customer" 

**Expected behavior**
The button should "taste" if it has some text inside itself / or "taste" which attributes are inside the button. Instead of saying which button it is, eg. "icon only".

**Screenshots**
Skærmbillede 2019-09-23 kl  10 08 37

jakobe:
Closes #618

### DIFF
--- a/src/app/examples/button-example/button-example.component.html
+++ b/src/app/examples/button-example/button-example.component.html
@@ -76,15 +76,58 @@
   <label>Disabled</label>
 </div>
 
+<h3>Custom Icon</h3>
+<div class="button-example-container">
+  <button kirby-button>
+    <kirby-icon customName="football"></kirby-icon>
+  </button>
+  <label>Attention Level 1<br/><em>(Default)</em></label>
+</div>
+<div class="button-example-container">
+  <button kirby-button attentionLevel="2">
+    <kirby-icon customName="football"></kirby-icon>
+  </button>
+  <label>Attention Level 2</label>
+</div>
+<div class="button-example-container">
+  <button kirby-button attentionLevel="3">
+    <kirby-icon customName="football"></kirby-icon>
+  </button>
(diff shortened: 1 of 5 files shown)


src/app/examples/button-example/button-example.component.html
```diff
@@ -76,15 +76,58 @@ <h3>Medium (default)</h3>
...
+  Icon Left
 </button>
-<button kirby-button text="Icon Right" iconPlacement="right">
+<button kirby-button>
+  Icon Right
   <kirby-icon name="arrow-down"></kirby-icon>
 </button>
-<button kirby-button text="Custom Icon">
+<button kirby-button>
+  <label>Label Left</label>
```
askarby: Holy crap, it's never dawned on me that labels can be used in buttons, I've always seen them as something you use for labeling `input`-, `select`- and `textarea`-elements.

I do however like it:
giphub

jb-trm: I don't think it is w3c valid markup - if that is thing in anymore :D And it seems we need to add a cursor behaviour in this case otherwise the cursor will not highlight the text as clickable inside the button.

jakobe: > I don't think it is w3c valid markup - if that is thing in anymore :D And it seems we need to add a cursor behaviour in this case otherwise the cursor will not highlight the text as clickable inside the button.

Good point, I'll remove it, we should promote weird/non standard practices in the cookbook

PR #643, opened 2019-09-30
Fix/icon button margin padding

jakobe:


src/app/examples/button-example/button-example.component.ts
```diff
@@ -21,8 +20,24 @@ export class ButtonExampleComponent {
...
+    {
+      text: 'Large',
+      value: 'lg',
+    },
+  ];
 
-  onSegmentSelect(segment) {
-    this.activeTab = segment.id;
+  onButtonSizeChange(value) {
+    console.log(value);
```
jonasjensen-brolstark: Maybe remove this? 🤷‍♂

jakobe:
```suggestion
```

Issue #633, opened 2019-09-26
[Enhancement] Finishing touch on List UI


bommariusser:
**Is your enhancement request related to a problem? Please describe.**
Our lists are functionality-wise near complete but they need a last finishing touch to be complete, so thats why should spend some time on finalizing the ui.

Things that we should look at are:
- Font-sizing & weight; They are mostly light in weight now and should be Regular and Bold dependant on situation
- Paddings between the different "components" inside each item like avatar, badge etc.
- Paddings between different components lying under each other like titles and items

See the difference here between Test and Wireframe:
![Skærmbillede 2019-09-26 kl  10 18 15](https://user-images.githubusercontent.com/49033002/65671131-f8eb1800-e046-11e9-9be5-aae630e1cd77.png)

SanderJespersen:
Just to be 100% sure. Does this issue handle that the padding on all 4 sides is the same size within a section ("card") of the list?

bommariusser:
pretty sure @SanderJespersen - the task is about bumping the list to v1.0 - wireframe = dev

but please add a screenshot :)

jonasjensen-brolstark:
Fixed in #676

PR #676, opened 2019-10-11
Fix/padding-to-follow-card-design

jonasjensen-brolstark:
I have tried to align the styling of the list with the card design. 
That means that i have set the top/bottom padding of the first/last `ion-item` in the list.

I have also aligned the padding of the section header to follow the design. 

image


jonasjensen-brolstark:
Closes #633

PR #697, opened 2019-10-21
fix(page): wrap title and toolbar actions inside div

hafstad:
Wrapping title and toolbar actions because Ionic otherwise will conflict with our own transitions.


jakobe:
Fixed by #699

PR #709, opened 2019-10-23
feat(page): add life-cycle hooks

hafstad:
add enter and leave events for page

### DIFF
--- a/src/app/showcase/page-showcase/page-showcase.component.html
+++ b/src/app/showcase/page-showcase/page-showcase.component.html
@@ -15,6 +15,9 @@
         <li>Spacing of content</li>
     </ul>
 
+    <h2>Life-cycle hooks</h2>
+    <p>Be aware that using page with kirby-router-outlet, causes the views are being stacked instead of replaced. This means that you can't trust onInit and onDestroy. Instead you can trust the page events (enter) or (leave) instead.</p>
+
     <div class="page-example">
         <div>
             <h2>Simple Page</h2>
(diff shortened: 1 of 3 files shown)


src/app/showcase/page-showcase/page-showcase.component.html
```diff
@@ -15,6 +15,9 @@
         <li>Spacing of content</li>
     </ul>
 
+    <h2>Life-cycle hooks</h2>
+    <p>Be aware that using page with kirby-router-outlet, causes the views are being stacked instead of replaced. This means that you can't trust onInit and onDestroy. Instead you can trust the page events (enter) or (leave) instead.</p>
```
jakobe:
```suggestion
    <p>Be aware that using page with <code>kirby-router-outlet</code>, causes the views to be stacked instead of replaced. This means that you can't trust <code>onInit</code> and <code>onDestroy</code>. Instead you can trust the <code>kirby-page</code> events <code>(enter)</code> and <code>(leave)</code> instead.</p>
```

src/app/showcase/page-showcase/page-showcase.component.ts
```diff
@@ -46,6 +46,18 @@ export class PageShowcaseComponent {
       inputValues: ['string'],
     },
   ];
+  events: ShowcaseProperty[] = [
+    {
+      name: 'enter',
+      description: 'Emitted when the page is entered',
```
jakobe:
```suggestion
      description: 'Emitted when the page has animated into view',
```

src/kirby/components/page/page.component.ts
```diff
@@ -163,6 +168,7 @@ export class PageComponent implements OnInit, OnDestroy, AfterContentInit, After
 
   ngOnInit(): void {
     this.url = this.router.url;
+    this.enter.emit();
```
jakobe: This doesn't make it fire twice?

hafstad: No because, the ngOnInit only run once, and after that the emit in router events subscriptions take over.

PR #710, opened 2019-10-23
bug: List header creates extra space 

SanderJespersen:
Fixing an issue where the `ion-list-header` created extra space in certain cases:

NB: the green colour is only present as a visual aid and is not actually found in the code:
Here is a picture of the situation when the posting list has upcoming transactions in a card:
image

Another where we have expanded the list and removed the card. Here the button we use to collapse can be seen:
image

The final picture shows and "invisible header" coloured green which is what creates the extra space. In this case we have no upcoming transactions and the card should not be shown but the header is somehow still there:
image

### DIFF
--- a/src/kirby/components/list/list.component.html
+++ b/src/kirby/components/list/list.component.html
@@ -5,7 +5,7 @@
           [class.has-header]="listHeaderTemplate"
           [class.has-footer]="listFooterTemplate"
           (window:resize)="onResize()">
-  <ion-list-header *ngIf="listHeaderTemplate">
+  <ion-list-header *ngIf="listHeaderTemplate" class="list-header">
     <ng-container *ngTemplateOutlet="listHeaderTemplate"></ng-container>
   </ion-list-header>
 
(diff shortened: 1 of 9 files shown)


src/kirby/components/list/list.component.html
```diff
@@ -5,7 +5,7 @@
           [class.has-header]="listHeaderTemplate"
           [class.has-footer]="listFooterTemplate"
           (window:resize)="onResize()">
-  <ion-list-header *ngIf="listHeaderTemplate">
+  <ion-list-header *ngIf="listHeaderTemplate" class="list-header">
```
jakobe: Just use `ion-list-header` as selector:
```suggestion
  <ion-list-header *ngIf="listHeaderTemplate">
```

src/kirby/components/list/list.component.scss
```diff
@@ -30,6 +30,10 @@ ion-list, .list-items {
   }
 }
 
+.list-header {
```
jakobe: Just use `ion-list-header` as selector:
```suggestion
ion-list-header {
```

src/kirby/components/list/list.component.html
SanderJespersen: Done.

PR #721, opened 2019-10-25
:bug: :lipstick: Fix FAB sheet close icon

jakobe:


src/kirby/components/fab-sheet/fab-sheet.component.ts
```diff
@@ -20,6 +31,40 @@ export class FabSheetComponent implements AfterContentInit {
...
+    if (!fabButtonElement || !kirbyCloseIcon || retryCount >= maxRetryCount) {
+      return;
+    }
+    if (fabButtonElement.shadowRoot && fabButtonElement.shadowRoot.innerHTML) {
+      const closeIcon = fabButtonElement.shadowRoot.querySelector('.close-icon ion-icon');
+      if (closeIcon) {
+        const closeIconSvgLoaded = closeIcon.shadowRoot.querySelector('.icon-inner svg');
+        const ionCloseIcon = (closeIcon as unknown) as IonIcon;
+        if (ionCloseIcon && closeIconSvgLoaded) {
+          console.log(
```
hafstad: Either this should be considered removed, or only be logged out during !prod

jakobe: Good point, fixed 👍

PR #725, opened 2019-10-28
Fix/page component ng lifecycle hooks

jonasjensen-brolstark:
I have moved all the initialization of elements from  `ngAfterContentInit` to `ngAfterContentChecked` so changes are added after each change detection cycle.

We saw an issue with the old way, if our page title or action had an `ngIf` on it. So i have added `ngIf` to the advanced page example, so there is a visual way to test this.

As a bonus i fixed the page showcase page, as it was f'ed up and missing an example..

### DIFF
--- a/src/app/examples/page-example/advanced/page-advanced-example.component.ts
+++ b/src/app/examples/page-example/advanced/page-advanced-example.component.ts
@@ -61,7 +61,8 @@ const config = {
 export class PageAdvancedExampleComponent extends BasePageExampleComponent {
   static readonly template = config.template
     .replace(' defaultBackHref="/"', '')
-    .replace(' [innerHTML]="content">', '>...');
+    .replace(' [innerHTML]="content">', '>...')
+    .replace(' *ngIf="true"', '');
 
   items: ActionSheetItem[] = [
     { id: '1', text: 'Option 1' },
--- a/src/app/examples/page-example/advanced/page-advanced-example.component.ts
+++ b/src/app/examples/page-example/advanced/page-advanced-example.component.ts
@@ -27,14 +27,14 @@ const config = {
   
   <!-- Fixed Page Actions -->
   <kirby-page-actions *kirbyPageActions="{fixed: true}">
-    <button kirby-button (click)="onMoreSelect()">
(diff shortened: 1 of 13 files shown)


src/app/examples/page-example/advanced/page-advanced-example.component.ts
```diff
@@ -27,14 +27,14 @@ const config = {
   
   <!-- Fixed Page Actions -->
   <kirby-page-actions *kirbyPageActions="{fixed: true}">
-    <button kirby-button (click)="onMoreSelect()">
+    <button *ngIf="true" kirby-button (click)="onMoreSelect()">
```
jakobe: Please search and replace this from the template example:
```
  static readonly template = config.template
    .replace(' defaultBackHref="/"', '')
    .replace(' [innerHTML]="content">', '>...')
    .replace(' *ngIf="true"', '');
```

src/kirby/components/page/page.component.ts
```diff
@@ -15,6 +14,7 @@ import {
   AfterContentInit,
```
jakobe: :fire: Remove unused import
```suggestion
```

src/kirby/components/page/page.component.ts
```diff
@@ -192,13 +192,10 @@ export class PageComponent implements OnInit, OnDestroy, AfterContentInit, After
     });
   }
 
-  ngAfterContentInit(): void {
+  ngAfterContentChecked(): void {
```
jakobe: Does this have a performance impact? ⚡️

src/kirby/components/page/page.component.ts
jonasjensen-brolstark: It is not unused. Is still used on `PageActionsComponent`.

src/kirby/components/page/page.component.ts
jonasjensen-brolstark: Not as i see it. The most expensive thing we do, is creating and attaching the `IntersectionObserver`. But this is safeguarded so it only happens once. 
The rest is simple DOM traversal.

There is of course a danger, that future changes will introduce performance issues here.

Issue #729, opened 2019-10-28
[BUG] List with divider has no divider on small screens


bommariusser:
**Describe the bug**
A list with dividers has no dividers on small screens, but it should always have a divider.

![Skærmbillede 2019-10-28 kl  13 32 53](https://user-images.githubusercontent.com/49033002/67678801-cc008c80-f987-11e9-996e-2e18930218c7.png)

**To Reproduce**

1. Go to  cookbook

jonasjensen-brolstark:
Closed by #737

PR #737, opened 2019-10-29
:bug: Apply has-divider class to list instead of list items

peterlindholm:
Fix for #729  

The bug is caused because ion-item is not the only child type (ion-sliding-type).

Moved the .has-divider css class from the list items to the list itself and then styled both ion-item and ion-sliding-item (except last child).


src/kirby/components/list/list.component.scss
```diff
@@ -5,6 +5,13 @@ $background-color-hover: get-color('background-color');
 $shadow: get-elevation(2);
 $item-height: size('xxxl');
 
+ion-list.has-divider {
```
jonasjensen-brolstark: Following my personal opinion, i would like this rule to be moved into the `ion-list` rule on line 34.

By doing this we run the risk having split styling rules all over the `scss` file, making it hard to read.

peterlindholm: Fully agree - thanks.

src/kirby/components/list/list.component.scss
```diff
@@ -5,6 +5,13 @@ $background-color-hover: get-color('background-color');
...
+  &:not(:last-of-type) ion-item-sliding,
```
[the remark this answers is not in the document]
peterlindholm: Indeed.

PR #750, opened 2019-11-05
KirbyTestingModule doesn't provide mock of Calendar

askarby:
:bug: CalendarComponent wasn't provided as a mock in testing module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "designsystem",
-  "version": "0.8.30",
+  "version": "0.8.31",
   "scripts": {
     "ng": "ng",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "designsystem",
-  "version": "0.8.30",
+  "version": "0.8.31",
```
hafstad: Is this a mistake? Kirby should bump it's version by the build pipeline, so unless there's a special reason this should be reverted.

askarby: Yes, it's a mistake... I forgot that it automatically performed "the bump"... reverting my change!

PR #764, opened 2019-11-11
Divider color on background

MortenGregersen:
When the divider was added to a modal or something else with `background-color` as background color, it would be invisible (also have `background-color` as color).

Now it will have `background-color` as color when in Card, unless the card has either `white` or `light` theme color. If not in card it will have `medium` as color.

### DIFF
--- a/src/kirby/components/divider/divider.component.scss
+++ b/src/kirby/components/divider/divider.component.scss
@@ -1,10 +1,18 @@
 @import '../../scss/utils';
 
+:host-context(kirby-card) {
+  --kirby-divider-color: #{get-color('background-color')};
+}
+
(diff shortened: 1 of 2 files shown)


src/kirby/components/divider/divider.component.scss
```diff
@@ -1,10 +1,18 @@
 @import '../../scss/utils';
 
+:host-context(kirby-card) {
```
jakobe: Consider merging the 2 selectors:
```suggestion
:host-context(kirby-card),
:host-context(.kirby-color-brightness-white) {
  --kirby-divider-color: #{get-color('background-color')};
}

```

MortenGregersen: Done

PR #781, opened 2019-11-20
Fix/centered title & Truncate longer page-titles

### LINKED ISSUE #739: [BUG] Centered Page-titles have same amount of margin/padding (left&right)
**Describe the bug**
Right now there is a 32dp left & 16dp right, which makes a centered title not centered.

See also: 
Skærmbillede 2019-10-29 kl  19 59 27

**To Reproduce**

1. Go to https://cookbook.kirby.design/home/showcase/page
2. View the centered title

### LINKED ISSUE #780: [Enhancement] Truncate longer page-titles in toolbar on scroll down
Right now we truncate the page-title without a "..." in a not so favorable fashion:
Skærmbillede 2019-11-20 kl  12 45 10

We would like to truncate the page-title with a "..." in the end like here: https://zpl.io/2E1OKBx

hafstad:
closes #739 closes #780


src/kirby/components/page/page.component.scss
```diff
@@ -22,6 +22,13 @@ ion-toolbar {
   ion-buttons[slot='secondary'] {
     order: 6;
   }
+
+  * {
```
jakobe: Consider using `@include slotted('*') {...}` as I guess `*` won't work otherwise? E.g. for customTitleTemplate...

hafstad: You're absolutely right about this won't truncate custom titles, however doing @include slotted('*') doesn't ensure that either as the custom title can be anything for instance our example with flex-inline which makes this kind of styling having no effect at all. So the question is if we wanna have any styling like this at our custom titles, or the user of a custom title should take care of these things themself. Otherwise we can ofc. have both as sort of sensible defaults, what do you think?

jakobe: But by using `*` it will be tagged with the host class, whereas `@include slotted('*')` will render as `> *` meaning a direct descendant of the host? Right?

PR #819, opened 2019-12-09
Add possibility to opt out of margin on divider

### LINKED ISSUE #805: [Enhancement] Configurable margin on kirby-divider
Separating two (or more) `kirby-item`s using a `kirby-divider` adds excessive margin to the bottom of the `kirby-item` before it, and excessive margin to the top of the `kirby-item` after it.

Either completely remove `margin-top` and `margin-bottom` from `kirby-divider` or create a option that can disable the margin.

Using `::ng-deep` but then I felt guilty and decided to file an enhancement request instead.

**Additional context**
It looks like the browser's default styling or `hr`-elements is "bleeding through.

Example from DRB-app:
image

Dev-tools, indicating user agent stylesheet "bleeding through":
image

MortenGregersen:
Closes #805

### DIFF
--- a/src/kirby/components/divider/divider.component.scss
+++ b/src/kirby/components/divider/divider.component.scss
@@ -12,4 +12,11 @@
 hr {
   border: 0;
   border-top: 1px solid var(--kirby-divider-color, get-color('medium'));
+
+  &.no-margin {
+    margin-top: 0;
+    margin-block-start: 0;
+    margin-bottom: 0;
+    margin-block-end: 0;
+  }
 }
--- a/src/kirby/components/divider/divider.component.scss
+++ b/src/kirby/components/divider/divider.component.scss
@@ -10,13 +10,13 @@
 }
 
 hr {
+  margin-top: size('xxs');
+  margin-bottom: size('xxs');
   border: 0;
   border-top: 1px solid var(--kirby-divider-color, get-color('medium'));
 
   &.no-margin {
     margin-top: 0;
-    margin-block-start: 0;
     margin-bottom: 0;
-    margin-block-end: 0;
   }
 }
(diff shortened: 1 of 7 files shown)


src/kirby/components/divider/divider.component.scss
```diff
@@ -12,4 +12,11 @@
 hr {
   border: 0;
   border-top: 1px solid var(--kirby-divider-color, get-color('medium'));
+
+  &.no-margin {
+    margin-top: 0;
```
askarby: I'm pretty sure that `margin-top` and `margin-block-start` is "the same" (that same goes for `margin-bottom` and `margin-block-end`?

Furthermore, we should have an explicit size variable, setting the `margin-top|bottom` when the margin is not disabled. As it is now, the user agent stylesheet simply bleeds through, and will result in a different margin size between browsers. Consider using the kirby `size`-function.

src/kirby/components/divider/divider.component.ts
```diff
@@ -1,8 +1,11 @@
...
 
 @Component({
   selector: 'kirby-divider',
   templateUrl: './divider.component.html',
   styleUrls: ['./divider.component.scss'],
 })
-export class DividerComponent {}
+export class DividerComponent {
+  @Input()
+  hasMargin: boolean;
```
askarby: Shouldn't this be initialized to `true`, to be backwards compatible with existing usage of the component?

MortenGregersen: As of @bommariusser's comment (https://github.com/kirbydesign/designsystem/issues/805#issuecomment-561622108), this should be `false` initially

MortenGregersen: I have added a "BREAKING CHANGE" label

src/kirby/components/divider/divider.component.scss
MortenGregersen: Fixed 👍

PR #820, opened 2019-12-09
feat: tabs scroll to top

### LINKED ISSUE #682: [Enhancement] Scroll up on pressing a selected tab
We want a native behavior, that refreshes the site / scrolls up to the top / goes to the very first site of the menu-point, when the user taps an active menu.

Eg. when a user is in a posting-detail (4 level in accounts) and presses the account tab again, they get to the very first site of the account (account overview)

lydemann:
Closes #682 
Scroll to the top of page when clicking the selected tab (with a 500 ms animation time)


src/kirby/components/page/page.component.html
```diff
@@ -32,7 +32,7 @@
   </ion-toolbar>
 </ion-header>
 
-<ion-content scrollEvents="false" forceOverscroll="false">
+<ion-content scrollEvents="false" #content forceOverscroll="false">
```
jakobe: Consider just selecting on Type as there is only 1 `ion-content`

src/kirby/components/page/page.component.ts
```diff
@@ -109,6 +110,7 @@ export class PageComponent implements OnInit, OnDestroy, AfterViewInit {
   @Output() enter = new EventEmitter<void>();
   @Output() leave = new EventEmitter<void>();
 
+  @ViewChild('content', { static: true }) private content: IonContent;
```
jakobe: Consider just selecting on Type as there is only 1 `ion-content`
```suggestion
  @ViewChild(IonContent, { static: true }) private content: IonContent;
```

src/kirby/components/tabs/tab-button/tab-button.events.ts
```diff
@@ -0,0 +1 @@
+export const tabClicked = 'tabClicked';
```
jakobe: Consider using present tense as this is consistent with built in event naming
```suggestion
export const tabClick = 'tabClick';
```

jakobe: Should Kirby events be prefixed with `kirby`to avoid clashes? 🤔

src/kirby/components/tabs/tab-button/tab-button.component.ts
```diff
@@ -28,8 +29,13 @@ export class TabButtonComponent implements OnInit {
     this.removeWrapper();
   }
 
-  onClick(event: Event) {
+  onClick(event: Event, clickedSelected: boolean) {
```
jakobe: Consider better naming;
```suggestion
  onClick(event: Event, isSelected: boolean) {
```

src/kirby/components/tabs/tab-button/tab-button.component.ts
```diff
@@ -28,8 +29,13 @@ export class TabButtonComponent implements OnInit {
...
     this.click.emit(event);
+
+    if (clickedSelected) {
+      const clickEvent = new CustomEvent(tabClicked);
```
jakobe: Seems a bit odd, that this event "tabClicked" is only dispatched if the tab was already selected??
Consider always dispatching event and adding a click-count that the event handler can take into account...

src/kirby/components/page/page.component.ts
```diff
@@ -175,13 +177,18 @@ export class PageComponent implements OnInit, OnDestroy, AfterViewInit {
         this.onEnter();
       }
     });
+
+    addEventListener(tabClicked, () => {
```
jakobe: What does this method refer to? If it's window.addEventListener it should be explicit...

src/kirby/components/page/page.component.ts
```diff
@@ -175,13 +177,18 @@ export class PageComponent implements OnInit, OnDestroy, AfterViewInit {
...
+      this.content.scrollToTop(KirbyAnimation.Duration.LONG);
+    });
   }
 
   ngOnDestroy(): void {
     if (this.routerEventsSubscription) {
       this.routerEventsSubscription.unsubscribe();
     }
     this.pageTitleIntersectionObserverRef.disconnect();
+    removeEventListener(tabClicked, null);
```
jakobe: What does this method refer to? If it's window.removeEventListener it should be explicit...

src/kirby/components/tabs/tab-button/tab-button.events.ts
lydemann: Good idea

src/kirby/components/tabs/tab-button/tab-button.component.ts
lydemann: Maybe rename event to `kirbyClickSelectedTab`?

src/kirby/components/page/page.component.ts
lydemann: Yes, and maybe use window abstraction

src/kirby/components/tabs/tab-button/tab-button.component.ts
lydemann: I think it is fair here to only dispatch if selected is clicked or it will add more complexity to the listener. Click count will not be enough as we might click different tabs. We would then need to parse the selectedTab id and then add a click count on that. I think just adding the event we actually need here will be an adequate solution. No big design upfront. :)

lydemann: As talked about I renamed the event with a more meaningful name and keep dispatching only if selected tab is clicked.

src/kirby/components/tabs/tab-button/tab-button.events.ts
```diff
@@ -0,0 +1 @@
+export const tabClicked = 'kirbySelectedTabClick';
```
jakobe: Consider naming...
```suggestion
export const selectedTabClickEvent = 'kirbySelectedTabClick';
```

PR #821, opened 2019-12-10
fix(page): make sure page actions and title can come in async with *ngIf

### LINKED ISSUE #813: [BUG] Kirby page actions in *ngIf's are not rendered
**Describe the bug**
If a `*kirbyPageActions` is wrapped in a container with a `*ngIf` it is no longer rendered. This worked perfectly well in version `0.8.49` but not in `0.8.50`.

Version `0.8.49`:

image

Version `0.8.50`:

image

**Kirby version**

- 0.8.50

- All plaforms

hafstad:
- fixes page will update if actions or title appears later because of *ngIf
- add back scroll-events as it's used by our lists

closes #813


src/kirby/components/page/page.component.ts
```diff
@@ -84,13 +82,8 @@ export class PageContentComponent {}
     <ng-content select="button[kirby-button]"></ng-content>
   `,
 })
-export class PageActionsComponent implements AfterContentInit {
+export class PageActionsComponent {
   @ContentChildren(ButtonComponent) buttons: QueryList<ButtonComponent>;
```
jakobe: Consider removing unused property
```suggestion
```

PR #827, opened 2019-12-11
Line break and white space support in empty state

### LINKED ISSUE #765: [Enhancement] Support linebreak in text
In some usecases we have more then one text that we want to put into our alerts. Fx 2 texts coming from different backends. Here we want to be able to line break / seperate those text, because they are supposed to be seperated from each other. 

Like this: https://zpl.io/bl6pyB0 or https://zpl.io/bP18ExA or https://zpl.io/29zNJWA

Maybe the alert can translate a /n to a linebreak so that we are able to do something like that?

SilverGull:
Closes #765 
Support for line breaks and spaces in messages in empty state and alerts

### DIFF
--- a/src/app/showcase/alert-showcase/alert-showcase.component.ts
+++ b/src/app/showcase/alert-showcase/alert-showcase.component.ts
@@ -18,7 +18,8 @@ export class AlertShowcaseComponent {
     },
     {
       name: 'message',
-      description: '(Optional) The message shown under the title (or icon if specified)',
+      description:
+        "(Optional) The message shown under the title (or icon if specified). Use '\\n' for newline",
       inputValues: ['string'],
     },
     {
--- a/src/app/showcase/alert-showcase/alert-showcase.component.ts
+++ b/src/app/showcase/alert-showcase/alert-showcase.component.ts
@@ -19,7 +19,7 @@ export class AlertShowcaseComponent {
     {
       name: 'message',
       description:
-        "(Optional) The message shown under the title (or icon if specified). Use '\\n' for newline",
+        "(Optional) The message shown under the title (or icon if specified). Use '\\n' for newline.",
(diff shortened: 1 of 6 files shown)


src/app/showcase/alert-showcase/alert-showcase.component.ts
```diff
@@ -18,7 +18,8 @@ export class AlertShowcaseComponent {
     },
     {
       name: 'message',
-      description: '(Optional) The message shown under the title (or icon if specified)',
+      description:
+        "(Optional) The message shown under the title (or icon if specified). Use '\\n' for newline",
```
jakobe: Isn't it only 1 backslash?
```suggestion
        "(Optional) The message shown under the title (or icon if specified). Use '\n' for newline",
```

src/app/showcase/empty-state-showcase/empty-state-showcase.component.ts
```diff
@@ -27,7 +27,7 @@ export class EmptyStateShowcaseComponent {
     },
     {
       name: 'subtitle',
-      description: 'The text beneath the title.',
+      description: "The text beneath the title. Use '\\n' for newline",
```
jakobe: Isn't it only 1 backslash?
```suggestion
      description: "The text beneath the title. Use '\n' for newline",
```

src/app/showcase/alert-showcase/alert-showcase.component.ts
SilverGull: the first one escapes the second one

src/app/showcase/empty-state-showcase/empty-state-showcase.component.ts
SilverGull: the first one escapes the second one :-)

PR #832, opened 2019-12-15
feat(modal): scroll to top and bottom

### LINKED ISSUE #807: [Enhancement] Ability to "scroll to bottom"
When the contents of a `kirby-modal` progressively grows, it may exceed the viewport height of the modal (which is intended behavior). The downside is that there is no (programmatic) way to scrolling to the content at the bottom. 

A handle on the `ModalController` to "scroll to bottom" would allow us to reveal the added content.

We've attempted to select "through the shadow DOM", and manually set the `scrollTop`-property... but to no avail.

**Additional context**
Manuallly setting `scrollTop` has no effect!

image

lydemann:
Closes #807 
Adding scrolling support to `ModalController` with `scrollToTop` and `scrollToBottom` methods that will scroll on the last instantiated modal's `IonContent`.
modal scroll

### DIFF
--- a/src/app/showcase/modal-showcase/modal-showcase.component.html
+++ b/src/app/showcase/modal-showcase/modal-showcase.component.html
@@ -108,15 +108,6 @@ myCallback(dataReturnedByModal: any) {{ '{' }}
     this.props = componentProps;
 {{ '}' }}</kirby-code-viewer
 
-
-
-
-
-
-
-
-
-
   ><br />
   <p>
     In order to hide a modal from within your embedded component (and destroy its component), you
--- a/src/app/showcase/modal-showcase/modal-showcase.component.html
+++ b/src/app/showcase/modal-showcase/modal-showcase.component.html
@@ -67,15 +67,18 @@ myCallback(dataReturnedByModal: any) {{ '{' }}
     methods <code>scrollToTop</code> and <code>scrollToBottom</code>. They accept
     <code>KirbyAnimation.Duration</code> as optional param. If not provided the scrolling will
     happen instantaneously.
-    <br />
-    <br />
+  </p>
+  <br />
 
+  <p>
     <code>scrollToTop</code> example (with long scroll animation):
     <kirby-code-viewer language="ts">
       this.modalController.scrollToTop(KirbyAnimation.Duration.LONG);
     </kirby-code-viewer>
-    <br />
+  </p>
+  <br />
 
+  <p>
     <code>scrollToBottom</code> example:
     <kirby-code-viewer language="ts">
       this.modalController.scrollToBottom();
@@ -106,7 +109,6 @@ myCallback(dataReturnedByModal: any) {{ '{' }}
 ) {{ '{' }}
     this.props = componentProps;
 {{ '}' }}</kirby-code-viewer
-
 
   ><br />
   <p>
--- a/src/app/showcase/modal-showcase/modal-showcase.component.html
+++ b/src/app/showcase/modal-showcase/modal-showcase.component.html
@@ -59,6 +59,28 @@ myCallback(dataReturnedByModal: any) {{ '{' }}
     ...
 {{ '}' }}</kirby-code-viewer>
   </p>
+  <h3>
+    Scrolling
+  </h3>
+  <p>
+    The <code>ModalController</code> support scrolling to the top and bottom of the modal using the
+    methods <code>scrollToTop</code> and <code>scrollToBottom</code>. They accept
+    <code>KirbyAnimation.Duration</code> as optional param. If not provided the scrolling will
+    happen instantaneously.
+    <br />
+    <br />
+
+    <code>scrollToTop</code> example (with long scroll animation):
+    <kirby-code-viewer language="ts">
+      this.modalController.scrollToTop(KirbyAnimation.Duration.LONG);
+    </kirby-code-viewer>
+    <br />
+
+    <code>scrollToBottom</code> example:
+    <kirby-code-viewer language="ts">
+      this.modalController.scrollToBottom();
+    </kirby-code-viewer>
+  </p>
   <br />
   <h3>Inside the embedded component:</h3>
   <p>To create a embedded component, ensure:</p>
@@ -84,6 +106,8 @@ myCallback(dataReturnedByModal: any) {{ '{' }}
 ) {{ '{' }}
     this.props = componentProps;
 {{ '}' }}</kirby-code-viewer
+
+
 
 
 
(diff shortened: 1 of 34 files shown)


src/kirby/components/modal/modal-wrapper/modal-wrapper.component.ts
```diff
@@ -26,6 +28,17 @@ export class ModalWrapperComponent {
       providers: [{ provide: COMPONENT_PROPS, useValue: this.config.componentProps }],
       parent: injector,
     });
+
+    modalController.scrollToTop = this.scrollToTop.bind(this);
```
jakobe: Doesn't this prevent garbage collection?

src/kirby/components/modal/services/modal.controller.ts
```diff
@@ -7,11 +7,18 @@ import { ActionSheetHelper } from './action-sheet.helper';
 import { ModalConfig } from '../modal-wrapper/config/modal-config';
 import { ActionSheetConfig } from '../action-sheet/config/action-sheet-config';
 import { AlertConfig } from '../alert/config/alert-config';
+import { KirbyAnimation } from '@kirbydesign/designsystem/animation/kirby-animation';
+
+export const modalScrollPlaceholderFunction = () => {
```
jakobe: Consider naming, e.g.
```suggestion
export const modalScrollNoop = () => {
```
or
```suggestion
export const noWindowsToScroll = () => {
```

src/kirby/components/modal/services/modal.controller.spec.ts
```diff
@@ -37,5 +37,21 @@ describe('modalController', () => {
         modalController.hideAll;
       }).not.toThrow();
     });
+
+    describe('scrollToTop', () => {
+      it('should throw an error when scrolling to top, when no modals have been opened', () => {
+        expect(() => {
+          modalController.hideTopmost();
```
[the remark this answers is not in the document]
lydemann: Yes, wups

src/kirby/components/modal/modal-wrapper/modal-wrapper.component.ts
lydemann: No idea, how would you test that?
We should probably update the `modalController.scrollToTop` when closing modals to avoid leaks as well as support scrolling still working after closing nested modal (we need to make the `scrollToTop` scroll on the right modal).

lydemann: This should be fixed now

src/kirby/components/modal/services/modal.controller.ts
```diff
@@ -7,11 +7,20 @@ import { ActionSheetHelper } from './action-sheet.helper';
 import { ModalConfig } from '../modal-wrapper/config/modal-config';
 import { ActionSheetConfig } from '../action-sheet/config/action-sheet-config';
 import { AlertConfig } from '../alert/config/alert-config';
+import { KirbyAnimation } from '@kirbydesign/designsystem/animation/kirby-animation';
+
+export const modalScrollNoop = () => {
```
askarby: Consider naming this differently... since, this is not a no-op implementation of a callback (it **will** throw an error). Maybe `defaultScrollCallback` is more appropriate?

src/kirby/components/modal/services/modal.controller.ts
```diff
@@ -7,11 +7,20 @@ import { ActionSheetHelper } from './action-sheet.helper';
...
+  throw new Error('No modal windows are currently registered');
+};
+
+export type modal = { close: (data?: any) => {}; config?: ModalConfig };
```
askarby: Wouldn't it be more appropriate to use an interface for this?

Regardless of using `type` or `interface`, they are (by convention) always written with an uppercase first letter.

src/kirby/components/modal/services/modal.controller.ts
lydemann: I agree

src/kirby/components/modal/services/modal.controller.ts
lydemann: Yes, let's do that

src/kirby/components/modal/modal-wrapper/modal-wrapper.component.spec.ts
```diff
@@ -132,4 +139,48 @@ describe('ModalWrapperComponent', () => {
...
+    it('should scroll to bottom with no scroll animation duration', () => {
+      const ionContentElm = fixture.debugElement.query(By.css('ion-content'));
+      const ionContent: IonContent = ionContentElm.componentInstance;
+      spyOn(ionContent, 'scrollToBottom');
+
+      component.scrollToBottom();
+
+      expect(ionContent.scrollToBottom).toHaveBeenCalledWith(0);
+    });
+    it('should scroll to bottom with provided scroll animation duration', () => {
```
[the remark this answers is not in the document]
lydemann: Let's do it

src/app/showcase/modal-showcase/modal-showcase.component.html
```diff
@@ -59,6 +59,28 @@ <h3>Inside the parent (caller) component:</h3>
     ...
 {{ '}' }}</kirby-code-viewer>
   </p>
+  <h3>
+    Scrolling
+  </h3>
+  <p>
```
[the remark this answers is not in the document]
lydemann: I agree, it is more correct HTML

src/app/showcase/modal-showcase/modal-showcase.component.html
```diff
@@ -59,6 +59,31 @@ <h3>Inside the parent (caller) component:</h3>
...
+    The <code>ModalController</code> support scrolling to the top and bottom of the modal using the
+    methods <code>scrollToTop</code> and <code>scrollToBottom</code>. They accept
+    <code>KirbyAnimation.Duration</code> as optional param. If not provided the scrolling will
+    happen instantaneously.
+  </p>
+  <br />
+
+  <p>
+    <code>scrollToTop</code> example (with long scroll animation):
+    <kirby-code-viewer language="ts">
```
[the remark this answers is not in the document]
lydemann: This needed  :)

src/app/showcase/modal-showcase/modal-showcase.component.html
```diff
@@ -110,6 +108,15 @@ <h3>Inside the embedded component:</h3>
     this.props = componentProps;
 {{ '}' }}</kirby-code-viewer
 
+
```
jakobe: Remove added whitespace

PR #859, opened 2020-01-13
fix(input): fixed default value of counter when used with late binding

laljyskebank:
Set value of counter when not present on initialization


laljyskebank:
Close

PR #864, opened 2020-01-13
863/item-corners-on-swipe-safari

### LINKED ISSUE #863: [BUG] Swipe animation in safari/iOS
**Describe the bug**
When swiping an item on a list it does not respect the bounds of it parents element that can have rounded corners. 

**To Reproduce**

1. Go to 'list-swipe' example page in safari
2. Swipe the top most item to the left or right
3. Click on one of the actions
4. See the when animating back in to position, the item is not respecting the bounds of it parents element

**Expected behavior**
The item should respect the bounds of the parent element.

**Screenshots**
image

- OS: MacOS, iOS
- Browser Safari
- Version All

jonasjensen-brolstark:
Closes #863 

The bug is a known issue on safari when combining use of `overflow`, `border-radius` and `transform`.

Fixed by setting the `backface-visibility` and `transform` on the parent element with the `border-radius`.

### Safari Before
image

### Safari After
image

### DIFF
--- a/src/kirby/components/list/list.component.scss
+++ b/src/kirby/components/list/list.component.scss
@@ -79,6 +79,14 @@ ion-item {
 }
 
 ion-item-sliding {
+  //  https://stackoverflow.com/a/16681137
+  backface-visibility: hidden;
+  -webkit-backface-visibility: hidden;
+  -moz-backface-visibility: hidden;
+  transform: translate3d(0, 0, 0);
+  -webkit-transform: translate3d(0, 0, 0);
+  -moz-transform: translate3d(0, 0, 0);
+
   $list-colors: ('light'); // add supported list item theme colors here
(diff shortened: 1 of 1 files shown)


src/kirby/components/list/list.component.scss
```diff
@@ -79,6 +79,14 @@ ion-item {
 }
 
 ion-item-sliding {
+  // backface-visibility and transform, to fix clipping issue on iOS see https://stackoverflow.com/a/16681137 and https://github.com/kirbydesign/designsystem/issues/863
+  backface-visibility: hidden;
+  -webkit-backface-visibility: hidden;
+  -moz-backface-visibility: hidden;
```
hafstad: Are these prefixes necessary? We should have auto-prefixer handling this

PR #883, opened 2020-01-23
Fix/input samsung device

laljyskebank:
**Describe the bug**
On android devices, maxlength on input field is not working. It is possible to write more characters than specified.

**To Reproduce**

1. Go to 'Faste betalinger'
2. Click on 'FAB'
3. Click on 'Tilmeld Betalingsservice'
4. Click on 'Beskrivelse'
5. Enter more than 32 characters

Screen Shot 2020-01-23 at 11 52 42

**Kirby version**
- v0.0.83

**Expected behavior**
Restrict characters to maxlength


jonasjensen-brolstark:
Question: Will this change the behavior if i uses a `formControl` on the `input` element and has a `maxlenght` validation rule on it?

If it changes the behavior, then maybe kirby should not implement this but be implemented in the using app.

hafstad:
> Question: Will this change the behavior if i use a `formControl` on the `input` element and have a `maxlenght` validation rule on it?
> 
> If it changes the behavior, then maybe kirby should not implement this but be implemented in the using app.
> 
> The reason for me asking is, that we on purpose let kirby be semi unaware of the input. We only listen for changes so the counter works. Validation and other rule enforcement was intentionally left for the implementing app.

I totally understand that argument, and therefore I won't take any decision before @jakobe is back. However as I understand it, UX have decided **always** to cut off characters exceeding the max-length so I think it would make sense to place this inside of Kirby.

PR #988, opened 2020-04-15
Ui/modal

### LINKED ISSUE #970: [Enhancement] Modal transitions
This issue is about refactoring our modals with the new ionic 5 modals to be able to use some of their new modal transitions.

jakobe:
This PR closes #970

## PR Checklist

- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [X] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [X] No


libs/designsystem/src/lib/components/app/app.component.spec.ts
```diff
@@ -1,18 +1,22 @@
...
+import { ModalController } from '../modal';
 
 describe('AppComponent', () => {
   let component: AppComponent;
   let fixture: ComponentFixture<AppComponent>;
 
   beforeEach(async(() => {
+    const modalControllerSpy = jasmine.createSpyObj('ModalController', [
+      'registerPresentingElement',
+    ]);
```
askarby, line 14: I believe we're already creating an entire spy object for the `ModalController`, in the `KirbyTestingModule` - consider using that instread?

libs/designsystem/src/lib/components/modal/services/modal.controller.ts
```diff
@@ -22,6 +22,9 @@ export class ModalController implements IModalController {
   ) {}
 
   public showModal(config: ModalConfig, onCloseModal?: (data?: any) => void): void {
+    if (config.dim != undefined) {
```
askarby, line 25: Consider using `!==` instead of `!=`?

Not sure if it would make sense to do:
```suggestion
    if (!config.dim) {
```
instead, perhaps we want to check for not using the property key `dim` at all here (which would make the current solution perfect).

libs/designsystem/src/lib/scss/_global-styles.scss
```diff
@@ -55,6 +55,20 @@ ion-modal:first-of-type {
...
   }
 
+  &.modal-card {
+    .modal-wrapper {
+      height: 100% !important;
+      border-top-left-radius: 0 !important;
+      border-top-right-radius: 0 !important;
+
+      @include media('>=medium') {
+        height: 647px !important;
```
askarby, line 65: Consider extracting  `647px` to a variable, since it's somewhat a "magic number" (needs context)

libs/designsystem/src/lib/components/modal/services/modal.controller.ts
jakobe: We ARE actually checking for whether the property is provided or not - so the most accurate would be: `config.hasOwnProperty('dim')`

jakobe: Fixed

apps/cookbook/src/app/examples/modal-example/modal-example.component.ts
```diff
@@ -4,12 +4,110 @@ import { ModalConfig, ModalController } from '@kirbydesign/designsystem';
 import { FirstEmbeddedModalExampleComponent } from './first-embedded-modal-example/first-embedded-modal-example.component';
 import { ModalCompactExampleComponent } from './compact-example/modal-compact-example.component';
 
-@Component({
+const config = {
```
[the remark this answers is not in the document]
jakobe: I agree with - most of :) - that, but:
* As you said, this was already an issue with the current html-snippets, so it's not in scope for this PR :dogdedthatbullet:
* Some of the code snippes are pseudo-code, e.g.:
```
onClose() {
  ...
}
so we can't have that as real, compilable code anyhoo...
```

Issue #916, opened 2020-02-10
[Enhancement] Make it possible to render calendar uniformly across timezones 


mvindahl:
**Is your enhancement request related to a problem? Please describe.**
The calendar alway obtains today's date from the local timezone. Thus, it may mark one date as the current one if accessed from, say, New Zealand and simultaneously mark another date as the current one if accessed from Hawaii. (This can be emulated by changing the system timezone).

This conflicts with the desired behaviour of our application as we want to present all users with the same view as if they were located in Denmark.

**Describe the solution you'd like**
Where as the behaviour may be fine in some applications, it is not the desired behaviour in our application. We desire a way to override it.

~~I think my preferred solution would be to introduce an optional property to fix the calendar rendering to a specific timezone, using any timezone string accepted by moment-timezone. In our case we would specify it as `[renderingTimezone]="Europe/Copenhagen"`. The calendar would then behave as if the browser were in Copenhagen, marking the current date as whichever date is the current one in .dk.~~
EDIT: On closer consideration, I think this would have been overengineering.

**Describe alternatives you've considered**
Introducing an optional property `todayDate` to explicitly tell the calendar which date it should consider to be today.

**Additional context**
There is another feature request relating to the Calendar and timezones: https://github.com/kirbydesign/designsystem/issues/826.

These two feature requests are orthogonal in the sense that this feature request deals with the internal workings of the calendar where as #826 deals with the date format used by its interface.

mvindahl:
On closer consideration I would vote for the simple `todayDate` property. This would also make it possible to control calendar state when running cypress tests. (I tried cy.clock() but I couldn't make it work).

mvindahl:
I have experimented with introducing `todayDate` as an optional property. It would solve all of our current issues.

I've implemented the change on a fork of the project: https://github.com/mvindahl/designsystem/tree/calendar-add-todaydate-parameter

If the suggestion is approved a pull request would be just a few clicks away.

mvindahl:
The following design has been agreed upon:

- introduce new optional input property `todayDate` to be able to tell the calender which calender date is the current one
- legal values is a Date following the same conventions as properties 'minDate', 'maxDate', etc. That is, it should point to midnight (see also #826).
- when omitted, defaults to constructing a calendar date based upon browser time, for backwards compatibility

@jakobe @Anders-Bloch please voice your objection if this is incorrect

jakobe:
Closed by #994

PR #994, opened 2020-04-21
Feature/calendar today date and timezone

### LINKED ISSUE #826: [Enhancement] Support UTC dates for Calendar
The Calendar component currently expresses dates as Date objects fixed to midnight in the time zone of the browser. It does so internally and it also does when receiving [selectedDate] and emitting (dateChange).

Thus, the actual timestamp of any given calendar date will vary across browsers. It's not a very useful timestamp for communicating calendar dates to a backend.

The standard way for our backend to express calendar dates are timestamps fixed to midnight of *UTC time*. If we naïvely extract timestamps from the Date objects emitted by the Calendar component and pass those to our backend, we will not hit UTC midnight in the general case and we might not even hit the correct calendar date.

It would be terribly convenient if the Calendar were able to (in an optional and non-breaking way) support Date objects expressing UTC midnights.  

Introduce an optional input property [dateFormat] with legal values "local" and "utc". If omitted, it will default to "local" to mimic the current behaviour. If "utc", it will accept and emit Date objects wrapping UTC midnights.

(On a side note, a far more readable and less bug prone date format might have been YYYY-MM-DD strings but oh well).

Out current approach is to immediately normalize any Date emitted by Calendar to make it a UTC midnight Date, and do do the reverse when initially setting its selected date. In that way we stick to UTC in "our" code but let the Calendar have it in its own way.

mvindahl:
This PR closes #826 and #916

## PR Checklist

- [x] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

See #826 and #916.

## What is the new behavior?

Two new input properties have been introduced on the Calendar, both optional, and both designed as to not introduce breaking changes:

- **todayDate** can be used to pass today's date, overriding the default behaviour of inferring today's date from the local time of the browser. This will allow us to render the calendar uniformly across timezones.
- **timezone** ("local" | "UTC") can be used to control which timezone is used for the Date objects used for input and output. Date objects represent points in time (epoch timestamps), and when using it to represent calendar dates, the convention is to make it point to the midnight of a given date. This, in turn, depends upon which timezone is used. For backward compatibility, the default choice will still be to use the local timezone but if UTC is a more convenient choice, there is now a way to override. For input parameters, the calendar will be resilient and will accept Date objects, even if they do not align with the chosen timezone, but it's highly recommended to make them align.

- [ ] Yes
- [x] No


apps/cookbook/src/app/examples/calendar-example/calendar-card-example.component.ts
```diff
@@ -1,48 +1,47 @@
...
-    const date8 = new Date();
-    date8.setDate(today.getDate() + 35);
-    this.disabledDates.push(date1, date2, date3, date4, date5, date6, date7, date8);
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.useTimezoneUTC) {
+      this.updateInputDates();
+
+      if (this.selectedDate) {
+        // realign selectedDate with the timezone that is now used, or the rendered date will
+        // be midleading and confusing
```
jakobe, line 35: Typo:
```suggestion
        // be misleading and confusing
```

libs/designsystem/src/lib/components/calendar/calendar.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="header">
-  <button [disabled]="!canNavigateBack" (click)="changeMonth(-1)">
+  <button class="previous-month-button" [disabled]="!canNavigateBack" (click)="changeMonth(-1)">
```
jakobe, line 2: This css class isn't defined?

libs/designsystem/src/lib/components/calendar/calendar.component.html
```diff
@@ -8,7 +8,7 @@
     ><span class="year">{{ activeYear }}</span>
   </div>
 
-  <button [disabled]="!canNavigateForward" (click)="changeMonth(1)">
+  <button class="next-month-button" [disabled]="!canNavigateForward" (click)="changeMonth(1)">
```
jakobe, line 11: This css class isn't defined?

libs/designsystem/src/lib/components/calendar/calendar.component.spec.ts
```diff
@@ -0,0 +1,237 @@
+import { createHostFactory, SpectatorHost } from '@ngneat/spectator';
+import moment from 'moment';
+import { MockComponent } from 'ng-mocks';
+
+import { CalendarComponent, IconComponent } from '..';
+
+// NOTE: when specifying multiple input properties, set selectedDate
+// as the last one. This makes the component update without the need to
+// explicitly call spectatir.comlponent.ngOnChanges()
```
jakobe, line 9: Typo:
```suggestion
// explicitly call spectator.component.ngOnChanges()
```

libs/designsystem/src/lib/components/calendar/calendar.component.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  it('should initially render the current month if selectedDate is not specified', () => {
+    const currentDayMoment = moment().startOf('day');
+    const currentMonthMoment = moment().startOf('month');
+
+    verifyMonthAndYear(currentMonthMoment.format('MMMM YYYY'));
+    expect(spectator.query('.day.today')).toHaveText(currentDayMoment.format('D'));
+  });
+
+  it('should initially render the month of that date if selectedDate is specified', () => {
```
jakobe, line 35: Typo:
```suggestion
  it('should initially render the month of selectedDate if specified', () => {
```

libs/designsystem/src/lib/components/calendar/calendar.component.html
jakobe: Ok, I see it's used on the unit test...

libs/designsystem/src/lib/components/calendar/calendar.component.html
jakobe: Ok, I see it's used on the unit test...

libs/designsystem/src/lib/components/calendar/calendar.component.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    spectator.setInput('disabledDates', localDates);
+    expect(spectator.component.disabledDates).toEqual(localDates);
+    spectator.setInput('disabledDates', utcDates);
+    expect(spectator.component.disabledDates).toEqual(localDates);
+  });
+
+  // constants and utility functions
+
+  const SEL_NAV_BACK = '.previous-month-button';
```
jakobe, line 204: Consider using button placement as selector instead of unused css class:
```suggestion
  const SEL_NAV_BACK = '.header button:first-of-type';
```

libs/designsystem/src/lib/components/calendar/calendar.component.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+  const SEL_NAV_FORWARD = '.next-month-button';
```
jakobe, line 205: Consider using button placement as selector instead of unused css class:
```suggestion
  const SEL_NAV_FORWARD = '.header button:last-of-type';
```

libs/designsystem/src/lib/components/calendar/calendar.component.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+  function utcMidnightDate(yyyyMMdd) {
+    return moment.utc(yyyyMMdd).toDate();
+  }
+
+  function clickDayOfMonth(dateOneIndexed: number) {
+    spectator.click(spectator.queryAll('.day.current-month')[dateOneIndexed - 1]);
+  }
+
+  function trimmedTexts(selector: string) {
+    return spectator.queryAll(selector).map((_) => _.textContent.trim());
```
jakobe, line 220: Consider using `[HTMLElement.innerText](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText)` instead:
> Basically, innerText is aware of the rendered appearance of text, while textContent is not
```suggestion
    return spectator.queryAll<HTMLElement>(selector).map((_) => _.innerText);
```

libs/designsystem/src/lib/components/calendar/calendar.component.ts
```diff
@@ -57,11 +62,54 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
     this.setActiveMonth(value);
     if (this.hasDateChanged(value, this._selectedDate)) {
       this.onSelectedDateChange(value);
-      this._selectedDate = value;
-      this.dateChange.emit(this._selectedDate);
+      this._selectedDate = this.normalizeDate(value);
+
+      if (!this._selectedDate || this.timezone === 'local') {
```
jakobe, line 67: Not very readable - effectively checking for whether this._selectedDate has a value belongs to the `else` statement.
Consider moving there or only emitting if there is a date at all... (Can it be not selected??)

libs/designsystem/src/lib/components/calendar/calendar.component.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+    spectator.setInput('alwaysEnableToday', true);
+    spectator.setInput('selectedDate', localMidnightDate('1997-08-29'));
+
+    const captured = captureDateChangeEvents();
+
+    clickDayOfMonth(28); // August 28th is was disabled but we override
+    expect(captured.event).toEqual(today);
+  });
+
+  it('should emit dateChange event as UTC midnights when when timezone is set to UTC', () => {
```
jakobe, line 159: Typo:
```suggestion
  it('should emit dateChange event as UTC midnights when timezone is set to UTC', () => {
```

libs/designsystem/src/lib/components/calendar/calendar.component.spec.ts
mvindahl: Yep, that's a better approach. I'll remove the CSS classes as well.

libs/designsystem/src/lib/components/calendar/calendar.component.html
mvindahl: No longer used so removed it

libs/designsystem/src/lib/components/calendar/calendar.component.html
mvindahl: No longer used so removed it

libs/designsystem/src/lib/components/calendar/calendar.component.ts
mvindahl: Moved the logic so that we now only emit events when the selectedDate changed due to user interaction. This also removed the need for the guard.

PR #1003, opened 2020-04-29
Fix/modal

jakobe:
This PR fixes 2 modal bugs that was introduced with #988:
* Respecting iOS safe area in modal flavor (and prevent it in drawer flavor as drawer already has top padding)
* Jagged close animation when keyboard is visible because of height of viewport:
When closing modal (call `.close()`, click close button or swiping header) we'll blur active element to force keyboard to close (if keyboard is visible we add a small delay to allow keyboard to hide before starting close animation)

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [X] No

### DIFF
--- a/libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
+++ b/libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
@@ -86,6 +86,12 @@ export class ModalWrapperComponent implements Modal, AfterViewInit, OnInit, OnDe
     }
   }
 
+  blurActiveElement() {
+    if (document.activeElement instanceof HTMLElement) {
+      document.activeElement.blur();
+    }
+  }
+
   private checkForEmbeddedFooter() {
     const embeddedComponentElement = this.ionContentElement.nativeElement.firstElementChild;
     if (embeddedComponentElement) {
--- a/libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
+++ b/libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
@@ -107,16 +107,16 @@ export class ModalWrapperComponent implements Modal, AfterViewInit, OnInit, OnDe
   }
 
   @HostListener('window:keyboardWillShow')
-  protected onKeyboardWillShow() {
+  _onKeyboardWillShow() {
     this.keyboardVisible = true;
   }
 
   @HostListener('window:keyboardWillHide')
-  protected onKeyboardWillHide() {
+  _onKeyboardWillHide() {
     this.keyboardVisible = false;
   }
 
-  protected blurActiveElement() {
+  blurActiveElement() {
     if (document.activeElement instanceof HTMLElement) {
       document.activeElement.blur();
     }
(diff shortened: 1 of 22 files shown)


libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
```diff
@@ -64,9 +65,28 @@ export class ModalWrapperComponent implements Modal, AfterViewInit, OnInit, OnDe
     this.ionContent.scrollToBottom(scrollDuration || 0);
   }
 
-  close(data?: any) {
+  async close(data?: any): Promise<void> {
```
lydemann, line 68: No tests for this?

libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
```diff
@@ -86,6 +106,22 @@ export class ModalWrapperComponent implements Modal, AfterViewInit, OnInit, OnDe
     }
   }
 
+  @HostListener('window:keyboardWillShow')
```
lydemann, line 109: Tests?

askarby: I believe that the `keyboardWillShow`-event is dispatched by cordova (capacitor), but won't we experience "the same wierdness" when running on a mobile browser?

libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
```diff
@@ -86,6 +106,22 @@ export class ModalWrapperComponent implements Modal, AfterViewInit, OnInit, OnDe
...
+  _onKeyboardWillShow() {
+    this.keyboardVisible = true;
+  }
+
+  @HostListener('window:keyboardWillHide')
```
askarby, line 114: I believe that the `keyboardWillHide`-event is dispatched by cordova (capacitor), but won't we experience "the same wierdness" when running on a mobile browser?

jakobe: @askarby Yep, this is a "temporary" fix for the native app - we are nowhere near supporting PWAs on native devices yet...

libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
lydemann: In the case, that this is also a problem for mobile screen keyboards, it seems that the resize event will be an option:
https://stackoverflow.com/questions/27132796/is-there-any-javascript-event-fired-when-the-on-screen-keyboard-on-mobile-safari

libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
jakobe: Added 👌

libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
jakobe: Added 👌

PR #1049, opened 2020-05-25
Replace `whenHydrated` with `whenReady`

jakobe:
Clean up some test utility methods


libs/designsystem/src/lib/testing/test-helper.ts
```diff
@@ -26,10 +30,15 @@ export class TestHelper {
     });
   }
 
+  /** Checks for the Ionic Web Component being hydrated, ie. the Shadow DOM is ready for query */
```
askarby, line 33: Looks like a copy of the documentation for the `whenHasCssClass`-function, I'm guessing this function doesn't have the same behaviour?

jakobe: Actually it was the other way around ;)

askarby (dismissed):
giphub

lydemann (dismissed):
LGTM

hafstad (approved):

Issue #866, opened 2020-01-14
[Enhancement] 44dp touch area


bommariusser:
**Is your enhancement request related to a problem? Please describe.**
We would like to have a touch area around some of our smaller components, so that's easier to hit them with the finger :) For that a 44x44px area would be optimal!

![Skærmbillede 2020-01-14 kl  08 22 02](https://user-images.githubusercontent.com/49033002/72322582-fc161c80-36a6-11ea-9edf-2b1490882053.png)

Components that needs to be looked on are:
- Buttons
- Checkmarks
- Radio (not ready yet)
- Item (Small should be 44 not 40)
- Others?

EdwinSilvaDK:
Item [Small] already has the size of 44px

RasmusKjeldgaard:
Just checked up on the state of this issue:
- Small size of `Button`, `Chip` in `Segmented Control`, `Dropdown` and `Toggle` all have a touch area that matches their size, which is `32 px`. 
- `Checkbox` and `Radio` has an extra small size, that explicitly limits the height of the touch area to `32px`. 
- `Item` with size small is `44px` and it is explicitly mentioned that extra small size is not supposed to be used for touch. 

So what do you think @bommariusser, do you agree that the focus should be on making the touch area of all the 44px for all the interactive elements that have a 32px size (`Button`, `Chip` in `Segmented Control`, `Dropdown` and `Toggle`)?

bommariusser:
yes @RasmusKjeldgaard i agree :)

RasmusKjeldgaard:
When adding the 44px touch area around a small button (height of 32px), it extends above and below the button and its 4px margin.

This means that the touch areas overlap if you place multiple buttons vertically without any additional spacing, like our attention level showcase in the cookbook:
https://user-images.githubusercontent.com/42470636/115253449-ac859280-a12c-11eb-9579-a97eb76b9e00.mov

This could also happen for other elements, like toggle and chip. 

So I think we should discuss if and how we want to solve this, here are some thoughts on what we could do: 
- Introduce spacing in components where two or more elements could potentially be placed horizontally 
- Accept the functionality as is, as placing elements like this is probably a rare use case anyways
- Remove the touch area functionality again, if it causes too many problems to have the touch area extend over the visual part of the element.

PR #1057, opened 2020-06-02
Make back buttons on Page bigger

MortenGregersen:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Right now the back buttons are 24x32 px, because of the lack of texts in them. This is a bit too small to hit on device. Apple recommends buttons to be at least 44 px on one of the sides, for them to be hittable.

Issue Number: N/A

## What is the new behavior?

With this change, the buttons will be 44x44 px, and thereby easier easier to hit. It is possible to set a `--min-width` on the back button, but it centers the icon (not preferred), so we need to use the `--padding-end`. This creates an empty space that the title can fill out (may be a problem).

- [ ] Yes
- [x] No


bommariusser:
@MortenGregersen this task is pretty much related to ur task or? https://github.com/kirbydesign/designsystem/issues/866

The back button, is that a kirby button? 
Is the back button the same as the buttons on the right side of the toolbar? 

:)

MortenGregersen:
It is related. Didn’t see that issue. 

It is the button on the left side of the page title when the user can navigate back.

bommariusser:
but its also related to the other buttons in the toolbar? :) ill look at it tomorrow. excited much!

MortenGregersen:
After talking with @bommariusser, I have now made the back button 44x44 px.

I haven't touched the other buttons. I think this should be done separately, as the buttons are given by the consuming app.

PR #1265, opened 2021-02-03
Enhancement/font weight black

### LINKED ISSUE #1037: [Enhancement] Add fontweight:"black"
1. We would like our H1 to be weight:black instead of weight:bold. It's default bold now, but it should be black. 

The reason is for that is we want the H1 to stick more out then the rest of our fonts like h2 and h3.

2. Also we want our Display font(s) to be Black as default instead of Light. See also: https://cookbook.kirby.design/home/showcase/fonts

Most of our display fonts are not styled manually and they use "default" styling. In one of our meetings we decided that this default should be Black instead of light: See also

Skærmbillede 2021-01-15 kl  13 34 02

The same goes for our h1 font which is bold by default and should be black.

MadsBuchmann:
This PR closes #1037 

## PR Checklist

- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [X] Other... Please describe: update styles

Issue Number: #1037

## What is the new behavior?
Fonts: h1, display 1, display 2 and display 3 (https://cookbook.kirby.design/home/showcase/fonts) now defaults to font-weight black.

- [ ] Yes
- [ ] No
- [X] in doubt

In the context of this project, is changing the default of a font considered breaking? Projects using the defaults will have their appereance altered. 

## Other information
Does this change have to be tested? If yes: are there any examples in the code where css rules are tested without a component? Been looking around, but can't seem to find anything. Where would be the appropiate place to store these tests?


libs/designsystem/src/lib/scss/base/_variables.scss
```diff
@@ -51,6 +51,8 @@ $font-weight: (
   light: 300,
   normal: 400,
   bold: 700,
+  // black will be treated as a named css color if not quoted
+    'black': 900,
```
MortenGregersen, line 55: Indentation seems to be off here

jakobe, line 55: 👍 for commenting why the quote is needed - consider readability though:
```suggestion
    'black': 900, // black will be treated as a named css color if not quoted
```

jakobe (approved):
LGTM! Please keep the wording from the PR template regarding the issue, otherwise the automation won't work:
`This PR closes #1037`
We should also consider adding tests for the typography - but that might be out of scope for this PR... (Maybe create an issue for that? 🙏 )

libs/designsystem/src/lib/scss/base/_variables.scss
MadsBuchmann: Agree - seems to be a rule related to comments, as it matches with the line-weight map in the same file.

MadsBuchmann: Gotcha! - i will go ahead and create a new issue regarding testing of typography 😀

libs/designsystem/src/lib/scss/base/_variables.scss
```diff
@@ -51,6 +51,7 @@ $font-weight: (
   light: 300,
   normal: 400,
   bold: 700,
+    'black': 900, // black will be treated as a named css color if not quoted
```
jakobe, line 54: Indentation is off...
```suggestion
  'black': 900, // black will be treated as a named css color if not quoted
```

jakobe (changes requested):

jakobe (approved):
Shippit!

MadsBuchmann:
I have created an issue regarding testing of typography: https://github.com/kirbydesign/designsystem/issues/1270

PR #1278, opened 2021-02-07
Add Kirby Bot token for version bump

jakobe:
Nothing to see here...

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -31,6 +31,7 @@ jobs:
         run: |
           git config user.name "Kirby Bot"
           git config user.email "<>"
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -31,6 +31,7 @@ jobs:
         run: |
           git config user.name "Kirby Bot"
           git config user.email "<>"
+          git remote add gh-origin https://kirbycibot:${secrets.VERSION_BUMP_GITHUB_TOKEN}@github.com/kirbydesign/designsystem.git
```
bogedal, line 34: We might need to change the push commands to be git push gh-origin master

bogedal (approved):
LGTM

bogedal (dismissed):

.github/workflows/publish.yml
jakobe: Add via token to checkout action instead

bogedal (approved):

PR #1456, opened 2021-04-08
Fix wrong colour (Skjern Bank brand)

### LINKED ISSUE #1395: [BUG] Wrong text colour (white) in actionsheet items and icon colour (white) in "more" icon.
**Describe the bug**
In "Skjern Bank" version of DRB some Action sheet items have white text on white background. Some icons eg. "more" menu action and "close" action is white on light grey background.

**To Reproduce**

1. Install Skjern Bank DRBMock
2. Go to 'Indbakke'
2. Click on 'Dokumenter'
3. In top right corner you see white icon (more menu)
4. Click on "more" icon
5. Action sheet item has white text on white background.

**Kirby version**
- [e.g. 1.4.21]

**Expected behavior**
Text and icons should be dark grey "#black-base"

**Screenshots**
IMG_2403
IMG_2404

- OS: [ iOS]

**Additional context**
Comps der bruger primary-contrast uden at forholde sig til den reelle bg farve:
image (6)

PI's will start to test the app by end march. Please have this bug fixed by mid april at the latest. 

## Tasks:
- [x] Have the bug report verified by contacting a member of @kirbydesign/kirby-guild
- [x] Move issue from "Ready to do" -> "in progress" on our [project board](https://github.com/kirbydesign/designsystem/projects/1)
- [x] Create Fix Branch
- [ ] Create test reproducing the error
- [ ] Fix
- [ ] If relevant [Update Examples and Showcase](https://cookbook.kirby.design/home/showcase/button)
- [ ] Push commits to GitHub
- [ ] Open a pull request (remember to add `closes #issueno` to the description) and ask @kirbydesign/kirby-guild for review
- [ ] Merge to master and delete branch

jkaltoft:
This PR closes #1395

## PR Checklist

Please check if your PR fulfils the following requirements:

## PR Type

Attention level 3 and 4 buttons have transparent background color (not part of change). They are expected to be displayed on a light background color, so this change makes their (text) color `black` instead of `primary-contrast`. With Skjern Bank brand colors `primary-contrast` is `white`. For all other brands `primary-contrast` is `black`.

- [x] Bugfix
- [ ] Feature
- [ ] Enhancement (to existing content)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #1395

## What is the new behavior?

Text color for attention level 3 and 4 buttons will be `black` regardless of brand colors.

- [ ] Yes
- [x] No

### DIFF
--- a/libs/designsystem/src/lib/components/button/button.component.spec.ts
+++ b/libs/designsystem/src/lib/components/button/button.component.spec.ts
@@ -267,6 +267,24 @@ describe('ButtonComponent', () => {
   });
 
   describe('when configured with attentionlevel 4', () => {
+    const canvasColor = 'background-color';
+
+    beforeAll(() => {
+      window.document.documentElement.style.setProperty(
+        'color',
+        getColor(canvasColor, 'contrast').value
+      );
+      window.document.documentElement.style.setProperty(
+        'background-color',
+        getColor(canvasColor).value
+      );
+    });
+
+    afterAll(() => {
+      window.document.documentElement.style.removeProperty('color');
+      window.document.documentElement.style.removeProperty('background-color');
+    });
+
     beforeEach(() => {
       spectator.component.attentionLevel = '4';
       spectator.detectChanges();
@@ -286,7 +304,7 @@ describe('ButtonComponent', () => {
 
     it('should render with correct color', () => {
       expect(element).toHaveComputedStyle({
-        color: getColor('primary', 'contrast'),
+        color: getColor(canvasColor, 'contrast'),
       });
     });
 
--- a/libs/designsystem/src/lib/components/button/button.component.spec.ts
+++ b/libs/designsystem/src/lib/components/button/button.component.spec.ts
@@ -205,6 +205,24 @@ describe('ButtonComponent', () => {
   });
 
   describe('when configured with attentionlevel 3', () => {
+    const canvasColor = 'background-color';
+
+    beforeAll(() => {
+      window.document.documentElement.style.setProperty(
+        'color',
+        getColor(canvasColor, 'contrast').value
+      );
+      window.document.documentElement.style.setProperty(
+        'background-color',
+        getColor(canvasColor).value
+      );
+    });
+
+    afterAll(() => {
+      window.document.documentElement.style.removeProperty('color');
+      window.document.documentElement.style.removeProperty('background-color');
+    });
+
     beforeEach(() => {
       spectator.component.attentionLevel = '3';
       spectator.detectChanges();
@@ -223,9 +241,8 @@ describe('ButtonComponent', () => {
     });
 
     it('should render with correct color', () => {
-      // TODO: Fix this test - should expect canvas contrast since bg is transparent
       expect(element).toHaveComputedStyle({
-        color: getColor('medium', 'contrast'),
+        color: getColor(canvasColor, 'contrast'),
       });
     });
 
--- a/libs/designsystem/src/lib/components/button/button.component.spec.ts
+++ b/libs/designsystem/src/lib/components/button/button.component.spec.ts
@@ -109,6 +109,7 @@ describe('ButtonComponent', () => {
     });
 
     it('should render with correct color', () => {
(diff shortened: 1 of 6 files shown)


libs/designsystem/src/lib/components/slide-button/slide-button.component.scss
```diff
@@ -31,6 +31,7 @@
     display: flex;
     justify-content: center;
     align-items: center;
+    color: $slide-button-text-color;
```
RasmusKjeldgaard, line 34: Why move this here?

libs/designsystem/src/lib/components/button/button.component.spec.ts
```diff
@@ -110,7 +128,7 @@ describe('ButtonComponent', () => {
 
     it('should render with correct color', () => {
       expect(element).toHaveComputedStyle({
-        color: getColor('white', 'contrast'),
```
RasmusKjeldgaard, line 113: I agree that this assumption was faulty, and with your changes the tests are now actually testing the intended behavior.

libs/designsystem/src/lib/components/button/button.component.spec.ts
```diff
@@ -19,6 +19,24 @@ describe('ButtonComponent', () => {
     declarations: [MockComponent(IconComponent)],
   });
 
+  const canvasColor = 'background-color';
+
+  beforeAll(() => {
```
RasmusKjeldgaard, line 24: Would a comment explaining why this is necessary be appropriate here?

RasmusKjeldgaard (commented):
I think it looks good, nothing much to add except for a couple of questions. 

Your refactoring of the tests make sense in both a light and dark primary color setup, and I have also manually tried both combinations, and all is good.

libs/designsystem/src/lib/components/button/button.component.spec.ts
jkaltoft: The test renders a button only with no context. That means, that the `color` property of the button will inherit its value from the document. The initial value (defined by the spec) for `color` is `#000000` - black, not kirby-black.

When kirby-primary is light e.g., `background-color` then kirby-primary-contrast will be kirby-black. This makes the assertions below (lines ~131 and ~245) fail because kirby-black is not completely black (#000000).

This could be left-overs from an earlier iteration and may be obsolete. I'll look into it 😊

libs/designsystem/src/lib/components/slide-button/slide-button.component.scss
jkaltoft: A rule of thumb is to never declare `background-color` without declaring `color` also (and vice versa) - and to always apply them to the same element/target (preferably within the same ruleset). The `color` property inherits its value when not declared explicitly, so `.slide-button-text` will inherit from here.

RasmusKjeldgaard: Makes perfect sense, thanks :)

libs/designsystem/src/lib/components/button/button.component.spec.ts
jkaltoft: > The test renders a button only with no context. That means, that the `color` property of the button will inherit its value from the document. The initial value for `color` is black (`#000000`), not kirby-black (`#1c1c1c`). (The initial value for color varies from one browser to another and in our case it's black).
> 
> When kirby-primary is light e.g., `background-color` then kirby-primary-contrast will be kirby-black. This makes the assertions below (lines ~131 and ~245) fail because kirby-black is not completely black (`#000000`).
> 
> This could be left-overs from an earlier iteration and may be obsolete. I'll look into it 😊

It is no longer necessary to arrange a canvas/context. Attention levels 3-4 have `background-color: transparent` and assume to always be displayed on a light background. Hence the `color` is always kirby-black regardless of brand colors (see https://github.com/kirbydesign/designsystem/blob/8c62cfef29a76c578cb10b634232d2b2c38f2c78/libs/designsystem/src/lib/components/button/button.component.scss)

I have removed it.

RasmusKjeldgaard (approved):
LGTM!

PR #1463, opened 2021-04-14
Enhancement/alert button notification color

### LINKED ISSUE #1142: [Enhancement] Alerts should be notification color, not primary (No branding)
Right now our alerts have branding inside them and they use a primary color on buttons. This hasn't be a problem in the Jyske Bank context, because they are so alike (branding & success). In another context like sydbank, almbrand and so on this isn't what we expected.

Alerts can by definition be either warning or destructive. The buttons inside them can either be destructive (for destructive actions) or success (for warning alerts that are always built upon at least one success button.

Skærmbillede 2020-09-17 kl  08 37 49

The primary color on our alerts should be replaced with a success color.

This will have an effect like this on a bank like Sydbank:

Skærmbillede 2020-09-17 kl  08 40 44

**Additional context**
We can't do this until we allow buttons to be notification color - RIght now they only can take danger when looking at our notification color.

RasmusKjeldgaard:
This PR closes #1142 

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Enhancement (to existing content)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #1142 

## What is the new behavior?
Inside alerts, the button would use the primary color (as expected) for the `okBtn`, because it has attention level 1. But inside alerts, we want this color to be the success color instead, so it does not dependent on the branding. 

- [ ] Yes
- [x] No


libs/designsystem/src/lib/components/modal/alert/alert.component.spec.ts
```diff
@@ -96,9 +105,46 @@ describe('AlertComponent', () => {
...
+    `);
+    element = spectator.element;
+  });
+
+  it('should create', () => {
+    expect(spectator.component).toBeTruthy();
+  });
+
+  it('should have success colors on button', () => {
+    const okButton = element.getElementsByClassName('ok-btn')[0];
```
jkaltoft, line 143: Suggestion: You could use `querySelector()` or `querySelectorAll()` to get the button element.

```
const okButton = element.querySelector('.ok-btn');
```

or

```
const okButton = element.querySelectorAll('.ok-btn')[0];
```

jkaltoft (commented):
image

I think it's worth rewriting the existing tests to use spectator, preferably in a separate housekeeping issue.

jkaltoft (approved):

PR #1465, opened 2021-04-14
Added minimum touch area to button and chip

### LINKED ISSUE #866: [Enhancement] 44dp touch area

RasmusKjeldgaard:
This PR closes #866 

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Enhancement (to existing content)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #866 

- [ ] Yes
- [x] No


libs/designsystem/src/lib/scss/base/_functions.scss
```diff
@@ -157,3 +157,19 @@
     @content;
   }
 }
+
+// Apply click area of at least 44 x 44 px
+// and ignore rounded corners to ensure it is square
+@mixin increase-target-size() {
```
jkaltoft, line 163: I'm not a big fan of mixins without arguments, among other things because their content gets duplicated each time the mixin is used, but that discussion is probably out of scope for this issue. We should discuss the use of mixins at some point, though 😊

What you *could* consider is add arguments for the target size like this:

```scss
@mixin increase-target-size($width, $height: $width) {
...
}
```

It will require that you call the mixin with at least one argument:

```scss
@include increase-target-size($fat-finger-size);
```

libs/designsystem/src/lib/components/chip/chip.component.scss
```diff
@@ -22,8 +22,10 @@
   margin: size('xxxs');
   cursor: pointer;
   user-select: none;
+  position: relative;
   @include theme-background('white');
   @include selected-and-hover('black');
+  @include increase-target-size();
```
jkaltoft, line 28: Mixin includes with no content block should be declared before local declarations

libs/designsystem/src/lib/components/button/button.component.scss
```diff
@@ -99,6 +99,7 @@ $button-width: (
   user-select: none;
   vertical-align: middle;
   white-space: nowrap;
+  position: relative;
```
jkaltoft, line 102: I think the `position: relative` declaration should be moved to the `increase-target-size()` mixin.

jkaltoft (changes requested):

libs/designsystem/src/lib/scss/base/_functions.scss
```diff
@@ -157,3 +157,19 @@
...
+  &:after {
+    content: '';
+    position: absolute;
```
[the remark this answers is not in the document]
RasmusKjeldgaard: Really good point, I originally wanted it to be clear in the components style file that position was set, but not that I move the mixin to the top, it should be clear enough.

libs/designsystem/src/lib/components/button/button.component.spec.ts
```diff
@@ -6,9 +6,7 @@ import { IconComponent } from '../icon/icon.component';
 
 import { ButtonComponent, ButtonSize } from './button.component';
 
-const getColor = DesignTokenHelper.getColor;
-const size = DesignTokenHelper.size;
-const fontSize = DesignTokenHelper.fontSize;
+const { getColor, size, fontSize, fatFingerSize } = DesignTokenHelper;
```
jkaltoft, line 9: Thank you 😊

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.spec.ts
```diff
@@ -96,6 +99,13 @@ describe('SegmentedControlComponent', () => {
       expect(component.value).toBe(items[2]);
     });
 
+    it('should have touch area with minimum size equal to fat finger size', () => {
+      const element = spectator.element.querySelector('ion-segment-button');
+
+      expect(window.getComputedStyle(element, '::after').minHeight).toEqual(fatFingerSize());
+      expect(window.getComputedStyle(element, '::after').minWidth).toEqual(fatFingerSize());
```
jkaltoft, line 106: This seems to test what the values of the CSS properties are rather than the actual width/height of the pseudo-element. `min-width` and `min-height` won't change depending on the width/height of the content of the element. The computed values of the `width` and `height` properties will.

Consider to test that the width and height of the pseudo-element is greater than or equals `fatFingerSize()` instead:

```
expect(window.getComputedStyle(element, '::after').height).toBeGreaterThanOrEqual(fatFingerSize());
```

Disclaimer: I'm not an expert in writing tests, so you better check that `toBeGreaterThanOrEqual` is a real matcher function 😉

libs/designsystem/src/lib/components/chip/chip.component.spec.ts
```diff
@@ -39,6 +40,13 @@ describe('ChipComponent', () => {
     });
   });
 
+  it('should have touch area with minimum size equal to fat finger size', () => {
+    expect(window.getComputedStyle(spectator.element, '::after').minHeight).toEqual(
+      fatFingerSize()
+    );
+    expect(window.getComputedStyle(spectator.element, '::after').minWidth).toEqual(fatFingerSize());
+  });
```
jkaltoft, line 48: See other comment regarding test (segmented-control)

libs/designsystem/src/lib/components/button/button.component.spec.ts
```diff
@@ -71,6 +70,11 @@ describe('ButtonComponent', () => {
       expect(element).toHaveComputedStyle({ margin: size('xxxs') });
     });
 
+    it('should have touch area with minimum size equal to fat finger size', () => {
+      expect(window.getComputedStyle(element, '::after').minHeight).toEqual(fatFingerSize());
+      expect(window.getComputedStyle(element, '::after').minWidth).toEqual(fatFingerSize());
+    });
```
jkaltoft, line 76: See other comment regarding test (segmented-control)

libs/designsystem/src/lib/components/chip/chip.component.scss
```diff
@@ -11,6 +11,10 @@
 }
 
 :host {
+  @include theme-background('white');
+  @include selected-and-hover('black');
+  @include accessible-target-size();
```
jkaltoft, line 16: Nice boy scouting :)

jkaltoft (changes requested):

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.spec.ts
RasmusKjeldgaard: Fixed in 44d80b33f8d0b350bdf2f3518eb6013be1bd122b

libs/designsystem/src/lib/components/chip/chip.component.spec.ts
RasmusKjeldgaard: Fixed in 44d80b33f8d0b350bdf2f3518eb6013be1bd122b

libs/designsystem/src/lib/components/button/button.component.spec.ts
RasmusKjeldgaard: Fixed in 44d80b33f8d0b350bdf2f3518eb6013be1bd122b

jkaltoft (approved):

jkaltoft (approved):

PR #1756, opened 2021-09-20
Remove dual deploy

bdythe:


RasmusKjeldgaard (approved):
Beautiful, with active deployment hook working and all! Looks good. 

I will make sure to look into the current routing strategy after this is merged.

PR #1838, opened 2021-10-25
Clickable cards should change cursor type

### LINKED ISSUE #1032: [Enhancement] "Clickable" as a modifier/attribute on cards
Right now the cards we have, that are "clickable", don't come out of the box with an `<a>` tag or `<button>` tag - thus they are not accessible, because the user doesnt get any help on which cards can be pressed and which cannot.

We would like to be able to set a card to "clickable/selectable" like our items can be modified. 

Technically the component that changes state should change its fill color by putting a fillcolor on top of the color with a given opacity, fx:

- If a white card changes state it should get a black color (Whites contast color) on top of it so it gets "grey"

**Specs are different from light to dark:**

Hover light component (dark contrast color): 2%
Hover dark component (white contrast color): 8%

Pressed light component (dark contrast color): 6%
Pressed dark component (white contrast color): 24%

This would also imply that the card gets the needed states like press and hover, see also: 

Hover: https://zpl.io/bAxdkEx
Pressed: https://zpl.io/agmDqN1

Skærmbillede 2021-10-26 kl  20 31 40

jeppevaldemar:

This PR closes #1032

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


libs/designsystem/src/lib/components/card/directives/card-is-clickable.directive.ts
```diff
@@ -0,0 +1,16 @@
+import { Directive, OnInit } from '@angular/core';
```
RasmusKjeldgaard, line 1: General comment: I would love if we could make this directive more generally applicable, so it could work for other elements at a later point if we need similar functionality. Therefore, I think we should try to figure out a way to separate this a bit from the card component. I have made some suggestions below in this regard.

libs/designsystem/src/lib/components/card/directives/card-is-clickable.directive.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import { CardComponent } from '../card.component';
+
+@Directive({
+  // tslint:disable-next-line
+  selector: 'kirby-card[click]',
+})
+export class CardIsClickableDirective implements OnInit {
+  constructor(private card: CardComponent) {}
+
```
RasmusKjeldgaard, line 11:
```suggestion
@HostBinding('attr.role') role: string = 'button';
```
Consider adding this host binding on the directive level.

The reason we need role="button" is for this to be more accessible, so screen readers will announce and treat this element as a button. You can read more in that regard [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role).  

Now that we have the `role="button"` attribute, we could remove the class host binding, and use this for our 'hook' for styling instead. See my comment in the card.component.scss file.

libs/designsystem/src/lib/components/card/directives/card-is-clickable.directive.ts
```diff
@@ -0,0 +1,16 @@
...
+  ngOnInit(): void {
+    this.card.cardIsClickable = true;
```
RasmusKjeldgaard, line 13: Consider removing this line based on above details.

libs/designsystem/src/lib/components/card/directives/card-is-clickable.directive.ts
RasmusKjeldgaard: In order to make the improvements described in the above linked MDN docs with regards to the elements acting as buttons, we should make sure that the keyDown event for space and enter will also fire the click action. I believe this could be achieved with something like:
```suggestion
  @HostListener('keydown.space', ['$event'])
  @HostListener('keydown.enter', ['$event'])
  onKeydownHandler(evt: KeyboardEvent) {
    this.clickableElement.nativeElement.click(evt);
  }
```
This would require us to inject something like `private clickableElement: ElementRef` in the constructor.

libs/designsystem/src/lib/components/card/directives/card-is-clickable.directive.ts
RasmusKjeldgaard: For the above to work, the card needs to be focusable by the keyboard. Consider adding another hostbinding for this, like 
`  @HostBinding('attr.tabindex') tabindex: number = 0;`

libs/designsystem/src/lib/components/card/card.component.ts
```diff
@@ -39,6 +39,7 @@ export class CardComponent implements OnInit, OnDestroy {
     this.sizesSortedByBreakpoint = this.sortSizesByBreakpoint(value);
   }
 
+  @HostBinding('class.card-is-clickable') cardIsClickable: boolean = false;
```
RasmusKjeldgaard, line 42: Consider removing this based on my comments to the directive (below)

libs/designsystem/src/lib/components/card/card.component.scss
```diff
@@ -31,6 +31,10 @@
   &.flat {
     box-shadow: none;
   }
+
+  &.card-is-clickable {
+    cursor: pointer;
+  }
```
RasmusKjeldgaard, line 37: This could be replaced with 
```suggestion
  &[role="button"] {
    cursor: pointer;
  }
```
if you agree with my suggestions to the directive (below)

RasmusKjeldgaard: And I just talked with @henrikvoetmand. We would also like an outline-offset to the focus ring of 2px (`outline-offset: 2px;`). Remember to use a size variable here.

libs/designsystem/src/lib/components/card/directives/card-is-clickable.directive.ts
RasmusKjeldgaard: With these changes, a new name should be considered. Maybe 'ElementAsButtonDirective'?

RasmusKjeldgaard (changes requested):
This is some damn good work. I have some suggested changes based on enhanced accessibility and keyboard navigation that I would like to include in this issue, as this is core functionality for buttons.

libs/designsystem/src/lib/components/card/directives/card-is-clickable.directive.ts
```diff
@@ -0,0 +1,16 @@
...
+    this.card.flat = false;
```
RasmusKjeldgaard, line 14: One last thing, to decouple from card, we could make the injected card dependency optional, and make a check here

`if (this.card) {
    this.card.flat = false;
  }`

RasmusKjeldgaard (changes requested):

apps/cookbook/src/app/examples/card/card-clickable-example/card-clickable-example.component.html
```diff
@@ -0,0 +1,14 @@
+<kirby-card (click)="function()">
+  <kirby-card-header title="Clickable"></kirby-card-header>
+  <div class="card-content">
+    <p>
+      Giving a card any click event <code>(click)="function()"</code> will change your cursor to a
+      pointer when hovering the card. There is also added accessibility so that you can focus the
+      card by tabbing and clicking with either space or enter.
```
RasmusKjeldgaard, line 7: I would suggest something like this just to be a bit more specific on the behavior: 
```suggestion
      pointer when hovering the card. To ensure that the card is accessible by keyboard, a tab-index is added so the card can be focused by tabbing. To mirror a buttons behavior, any registered click-event will also be called by pressing enter and space when the card is focused. 
```

RasmusKjeldgaard (approved):
Good job 💪

PR #1918, opened 2021-12-02
Create Item Sliding component 

### LINKED ISSUE #1885: [COMPONENT] Create a Kirby-item-sliding component

An "Item Sliding" component should be created with selector `kirby-item-sliding`. 

It shall allow for the creation of lists [with items that can be swiped](https://cookbook.kirby.design/#/home/showcase/list-swipe) to reveal actions.  

It should be possible to use it like: 
```html
<kirby-list> 
    <kirby-item-sliding [swipeActions]="swipeActions"> 
       <kirby-item> 
      <p>This is where content will go</p> 
   </kirby-item> 
   </kirby-item-sliding> 
   
</kirby-list>
```

Take inspiration from `kirby-list-item` (which has only been used internally) - it might be a solution to modify it such that it wraps kirby-item; it could then be renamed and exported. 
This is part of  
- #1883 
<hr />

## Checklist:

### Refinement

- [ ] ~~Request that the issue is [UX refined](https://github.com/kirbydesign/designsystem/blob/master/.github/CONTRIBUTING.md/#ux-refinement); do not proceed until this is done.~~
- [ ] Request that the issue is [tech refined](https://github.com/kirbydesign/designsystem/blob/master/.github/CONTRIBUTING.md/#tech-refinement); do not proceed until this is done.

### Implementation 

### Review

MadsBuchmann:

This PR closes #1885 

## What is the new behavior?
A new `ItemSlidingComponent` with selector `kirby-item-sliding` has been created. 

A `kirby-item` component can be slotted into it. If any `swipeActions` are given to it, it will be possible to swipe either left or right depending on `side` to reveal the actions below. 

It has the following input properties: 
- swipeActions: the actions that will be revealed when the item is swiped 
- side: defines which side the swipeActions will be displayed in. Was originally defined by 'position' on the first swipeAction object. But as they could only be displayed in one side and it was not possible to split them up, such that one set of actions were shown when swiping left and another when swiping right I decided to create a side input property instead.

For each swipeAction it is possible to set: 
- title: The title it will be displayed with 
- icon: If given it will be displayed with an icon 
- type: Will control the color the action is displayed with. 
- onSelected: the function that will be executed when an action is clicked 
- isDisabled: if true, will cause the action to not be rendered. 

The swipeActions have the following signature: 
```ListSwipeAction {
  position: ListSwipeDirection;
  title: ((item: any) => string) | string;
  icon?: ((item: any) => string) | string;
  type?: ((item: any) => ListSwipeActionType) | ListSwipeActionType;
  onSelected: (item: any) => void;
  isDisabled?: ((item: any) => boolean) | boolean;
}
```

But as we do not have any item included with the structured approach none of the functions are actually called with an item. I should find a way to cause the type to be correct before publishing this. 

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


libs/designsystem/src/lib/components/item-sliding/item-sliding.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  private evaluateSwipeAction(swipeAction: ListSwipeAction): EvaluatedListSwipeAction {
+    /* TODO: is it possible to autogenerate these based on the type? */
+    const keysToExclude: (keyof ListSwipeAction)[] = ['onSelected', 'position'];
+
+    const evaluatedEntries = Object.entries(swipeAction).map(([key, value]) => ({
+      [key]: value instanceof Function && !keysToExclude.includes(key as any) ? value() : value,
+    }));
+
+    return Object.assign({}, ...evaluatedEntries);
+  }
```
RasmusKjeldgaard, line 33: I wonder if we are jumping through too many hoops, and creating a bit much complexity for ourselves here?

The way I see it, this is to keep the external API for `ListSwipeAction` intact, which I really love. But, would an alternative be to simply change the `EvaluatedListSwipeAction` to something like `SwipeAction` and accept that teams would have to modify their code to support this way of using the sliding item?

The reason I got to think about this is the fact that I think the 'ListSwipeAction` is not a great name for something that is actually very item-centric. So maybe we could accept maintaining these two types, one for each of the ways to build out a list. 

I am truly in doubt.

RasmusKjeldgaard: But that would mean that we will have a much harder time using this new sliding item in the old list, right?

RasmusKjeldgaard (commented):
Really, really awesome work.
I think this is such a good job simplifying the API and reducing the components complexity. 
Well done! 

I dont have much for this draft review, expect for a comment regarding the EvaluatedListSwipeAction type.

libs/designsystem/src/lib/components/item-sliding/item-sliding.component.ts
MadsBuchmann: Warning! Stream of consciousness type'a'comment. 

No matter the conclusion, I will definitely rename `EvaluatedListSwipeaction` to `ItemSwipeAction` to make it more focused on item. 

I have until now (perhaps a bit blindly) just implemented the api from the original `ListSwipeAction`. Thinking a bit about it, I think we should remove the ability of passing in functions for the various values as we can no longer pass in an `item` for these functions. Which greatly reduces their usefulness and I can't actually see why you would use them. 

In the end, what the ability to pass functions give you, is to modify behaviour based on item. 

So the following: 

```typescript
swipeActions = {
   ...
   title: (item) => item.id % 2 === 0 ? `even: ${item.title}` : `odd: ${item.title}`
   ...
}; 
``` 

Would be equivalent to doing something like this (assuming swipeActions are ordered and id's are going from 0,1,2...swipeActions.length): 

HTML: 
```html
  <kirby-item-sliding
    *ngFor="let item of items; let idx = index"
    [swipeActions]="getSwipeActions(idx)"
  >
    <kirby-item>{{ item.value }}</kirby-item>
  </kirby-item-sliding>
```

TS: 
```typescript 
class SwipeExample {
    ...
    swipeActions = [{..., title: 'Hello'}, {..., title: 'World'}, {..., title: 'how'}, {..., title: 'ya'}, {..., title: 'doing?'}]

   getSwipeActions(idx: number): ListSwipeAction {
      const { title } = swipeActions[idx]
      return {
      ...
      title: idx % 2 === 0 ? `even: ${title}` : `odd: ${title}`
      ...
   }
}
```

As can be seen it is more work for the consumer. But that is part of the tradeoff when opening up and allowing consumers to construct items declaratively, instead of relying on them supplying item as a structured object that we then handle. 

... Thinking a bit more about it a solution that allows for both approaches, should be to allow for providing data for the callback via an input property, perhaps `contextData`? We would then just pass that data back when calling the functions in the ListSwipeActions. To recreate the above examples you would then: 

Template
```html
  <kirby-item-sliding
    *ngFor="let item of items;"
    [swipeActions]="swipeActions"
    [contextData]="item"
  >
    <kirby-item>{{ item.value }}</kirby-item>
  </kirby-item-sliding>
```

ItemSlidingComponent: 
```Typescript
 @Input() contextData: any; 

 private evaluateSwipeAction(swipeAction: ListSwipeAction): EvaluatedListSwipeAction {
    const keysToExclude: (keyof ListSwipeAction)[] = ['onSelected', 'position'];

     const evaluatedEntries = Object.entries(swipeAction).map(([key, value]) => ({
      // Note: the important change is on the below line: 
      [key]: value instanceof Function && !keysToExclude.includes(key as any) ? value(this.contextData) : value,
    }));

     return Object.assign({}, ...evaluatedEntries);
  }
```

Example Component (again, assuming swipeActions are ordered and id's are going from 0,1,2...swipeActions.length):  
```typescript
swipeActions = {
   ...
   title: (contextData) => contextData.id % 2 === 0 ? `even: ${contextData.title}` : `odd: ${contextData.title}`
   ...
}; 
```

MadsBuchmann: Having slept on this, I do not think we should create a contextData field until it has proven necessary. 

We can always add the option of having contextData - it is harder to remove it.

libs/designsystem/src/lib/components/list/list-item/list-item.component.ts
```diff
@@ -138,7 +141,8 @@ export class ListItemComponent implements OnInit, AfterViewInit {
 
   private initializeSwipeActions(): void {
     if (this.swipeActions && this.swipeActions.length) {
-      this._isSwipingEnabled = this.platform.isTouch();
+      //this._isSwipingEnabled = this.platform.isTouch();
+      this._isSwipingEnabled = true;
```
MadsBuchmann, line 145: Revert - it was for testing purposes.

libs/designsystem/src/lib/components/list/list-item/list-item.component.ts
```diff
@@ -113,6 +115,7 @@ export class ListItemComponent implements OnInit, AfterViewInit {
     return 'end';
   }
 
+  /* TODO: implemented in swipeAction class */
```
MadsBuchmann, line 118: Nopesies. Turned out differently. Delete comment.

libs/designsystem/src/lib/components/list/list-swipe-action.type.ts
```diff
@@ -4,11 +4,23 @@ export type ListSwipeDirection = 'right' | 'left';
...
 export type ListSwipeEnd = 'start' | 'end';
 
-export interface ListSwipeAction {
-  position: ListSwipeDirection;
-  title: ((item: any) => string) | string;
-  icon?: ((item: any) => string) | string;
-  type?: ((item: any) => ListSwipeActionType) | ListSwipeActionType;
+export type ItemSwipeAction = {
+  title: string;
+  // TODO: item is not provided in the function. Remove it.
```
MadsBuchmann, line 9: I should do that.

libs/designsystem/src/lib/components/list/list-swipe-action.type.ts
```diff
@@ -4,11 +4,23 @@ export type ListSwipeDirection = 'right' | 'left';
...
   onSelected: (item: any) => void;
-  isDisabled?: ((item: any) => boolean) | boolean;
-}
+  icon?: string;
+  type?: ListSwipeActionType;
+  isDisabled?: boolean;
+};
```
MadsBuchmann, line 14: Move this type to a `components/item-sliding/item-sliding.types.d.ts` file.

libs/designsystem/src/lib/components/list/list.component.scss
```diff
@@ -128,6 +129,7 @@ ion-item-divider {
 
   .item-content {
     display: inline-grid;
+    /* TODO: This min-width is shared*/
```
MadsBuchmann, line 132: It kinda is. But had to give the other button a 92px min width, to get it to behave visually as this one does. Look into that. Is there a way to share the min-width?

libs/designsystem/src/lib/components/list/list.component.scss
```diff
@@ -113,6 +113,7 @@ ion-item-divider {
   color: #{get-color('black')};
   background-color: transparent;
 
+  /* TODO: This for loop is shared */
```
MadsBuchmann, line 116: It indeed is - find a way to share it.

libs/designsystem/src/lib/helpers/element-has-ancestor.ts
```diff
@@ -0,0 +1,28 @@
+/* 
+  Utility function to check if a given element 
+  has a ancestor with a specific node name.
```
MadsBuchmann, line 3:
```suggestion
  has an ancestor with a specific node name. 
```

libs/designsystem/src/lib/components/list/list-swipe-action.type.ts
```diff
@@ -4,11 +4,23 @@ export type ListSwipeDirection = 'right' | 'left';
...
-  icon?: ((item: any) => string) | string;
-  type?: ((item: any) => ListSwipeActionType) | ListSwipeActionType;
+export type ItemSwipeAction = {
+  title: string;
+  // TODO: item is not provided in the function. Remove it.
   onSelected: (item: any) => void;
-  isDisabled?: ((item: any) => boolean) | boolean;
-}
+  icon?: string;
+  type?: ListSwipeActionType;
```
MadsBuchmann, line 12:
```suggestion
  type?: ItemSwipeActionType;
```

Move ListSwipeActionType to `item-sliding.types.d.ts` and reexport it with a new name from here to not break existing functionality.

MadsBuchmann (commented):
Double check that necessary types are exported.

MadsBuchmann:
Apparently it does not build either. I should fix that 💥

libs/designsystem/src/lib/components/item-sliding/item-sliding.component.scss
```diff
@@ -0,0 +1,12 @@
+@import '~@kirbydesign/core/src/scss/utils';
+
+ion-item-option {
+  min-width: 92px;
```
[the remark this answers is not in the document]
MadsBuchmann: I can't seem to find any.

libs/designsystem/src/lib/components/item-sliding/item-sliding.component.scss
```diff
@@ -0,0 +1,12 @@
...
+
+  @each $color-name, $color-value in $main-colors {
+    &.#{$color-name} {
+      background-color: #{get-color($color-name)};
+      color: #{get-color($color-name + '-contrast')};
+    }
+  }
```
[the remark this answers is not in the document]
MadsBuchmann: Fixed it.

libs/designsystem/src/lib/components/list/list.component.scss
RasmusKjeldgaard: Just unresolving this to make sure you are aware of this todo in the code. Feel free to resolve again!

RasmusKjeldgaard (changes requested):
Good friggin' job! 

Minor suggestions is all I got.

libs/designsystem/src/lib/components/item-sliding/item-sliding.component.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    const sideScenarios: { [key in ItemSlidingSide]: string } = {
+      left: 'start',
+      right: 'end',
+    };
+
+    Object.entries(sideScenarios).forEach(([side, slot]) => {
+      describe(`when 'side' is set to '${side}'`, () => {
+        beforeEach(() => {
+          // TODO: why do i need any here?
```
MadsBuchmann, line 98: Take care of this todo

libs/designsystem/src/lib/components/item-sliding/item-sliding.component.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+          spectator.setInput('side', side as any);
+        });
+
+        it(`should render swipe actions in the correct slot`, () => {
+          //TODO: Is there a better way to check this?
+          expect(ionItemOptionsElement.getAttribute('ng-reflect-side')).toEqual(slot);
```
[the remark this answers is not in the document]
MadsBuchmann: You might be right. I might just go ahead and delete this test instead. Testing if `_side` is correct when `side` is set also seems to be a tad too implementation specific.

libs/designsystem/src/lib/components/item-sliding/item-sliding.component.ts
```diff
@@ -0,0 +1,21 @@
...
+export class ItemSlidingComponent {
+  @Input() swipeActions: ItemSwipeAction[];
+
+  _side: 'start' | 'end' = 'start';
+  @Input() set side(value: ItemSlidingSide) {
+    this._side = value === 'left' ? 'start' : 'end';
+  }
+
+  get _hasSwipeActions(): boolean {
+    return Array.isArray(this.swipeActions) && this.swipeActions.length !== 0;
```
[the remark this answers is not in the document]
MadsBuchmann: You're completely right. It's actually a bit of leftover code from when i was goofing around with it.

libs/designsystem/src/lib/components/list/list-swipe-action.type.ts
```diff
@@ -1,14 +1,23 @@
...
-export type ListSwipeEnd = 'start' | 'end';
-
-export interface ListSwipeAction {
+// Copy each property in ItemSwipeAction but make it possible to provide them via a function.
+type ListSwipeActionBaseTypes = Pick<ItemSwipeAction, 'title' | 'icon' | 'type' | 'isDisabled'>;
+export type ListSwipeAction = {
+  [Property in keyof ListSwipeActionBaseTypes]:
+    | ((item: any) => ListSwipeActionBaseTypes[Property])
+    | ListSwipeActionBaseTypes[Property];
+} & {
```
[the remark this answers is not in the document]
MadsBuchmann: None of the tests for the template driven list failed and the cookbook was able to build. So i took that as a sign that nothing was broken. But it could be a good idea to test it out in the cookbook.

libs/designsystem/src/lib/components/list/list.component.scss
MadsBuchmann: Lol thank you. Totally forgot to remove it 🤦

apps/cookbook/src/app/showcase/item-sliding-showcase/item-sliding-showcase.component.html
```diff
@@ -0,0 +1,63 @@
+<article>
+  <section class="introduction">
```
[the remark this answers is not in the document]
MadsBuchmann: Removed both the class and the scss file

libs/designsystem/src/lib/components/item-sliding/item-sliding.component.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+      it('should not render the disabled swipeAction', () => {
+        expect(ionItemOptionElements.length).toBe(swipeActions.length - 1);
+      });
+    });
+
+    const swipeActionTypesScenarios: { [key in ItemSwipeActionType]: string } = {
+      warning: '#ffca3a',
+      success: '#2cf287',
+      danger: '#ff595e',
+    };
```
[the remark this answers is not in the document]
MadsBuchmann: Done

libs/designsystem/src/lib/components/item-sliding/item-sliding.component.ts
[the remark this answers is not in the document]
MadsBuchmann: Ah remembered why i did it. 

The alternative should be to do something like this: 
```typescript
this.swipeActions?.length !== 0;
```

This is however true in the case that `this.swipeActions` is undefined. As `undefined !== 0` is a true statement. 

But what can be done instead is: 

```typescript
   return this.swipeActions?.length > 0; 
```

as `undefined > 0` is false. 

In my head at least, i think that for future developers that might look at this, it might not be clear why larger than is used instead of not equals. Therefore to express intention Array.isArray() is used. 

Buuuut there's no reason the computer should do extra work while it is completely negligble to express that intention when i could just throw in a comment instead. 

So that is what i will be doing 😄

RasmusKjeldgaard (approved):
Installed it locally in DRB, and spun it up without any issues w.r.t. types, so it sure seems like we are good to go 🚀

PR #1941, opened 2021-12-15
Make all core scss files available from designsystem by using forward rule

### LINKED ISSUE #1939: [Housekeeping] @forward all Sass files instead of copying

Copying Sass files (including their content) from `core` to `designsystem` as we currently do introduces a few problems. First and foremost a bug was introduced in relationsbanken due do this (configuring the same module two times). Fixed by #1936.

**Idea for how to handle scss files**

We could create a copy of each file in core with the same name and file position in the designsystem package. The content of the copy should then be replaced with a `@forward <file name>` rule.

This would mean that the core scss are the source of truth and we do not have to change names two places if we have to rename them.

MadsBuchmann:

This PR closes #1939

## What is the new behavior?

A script that is executed when publishing has been created which forwards all scss files from `libs/core` to the designsystem distribution bundle. 

To test it out execute: `npm run publish`. 

The distribution has a `scss` folder at the following path with files forwarding the corresponding file from core scss: `dist/libs/designsystem/scss`. 

Example`cat dist/libs/designsystem/scss/print/components/_kirby-page.scss`: 
> @forward "~@kirbydesign/core/scss/print/components/kirby-page";

Also the `copyScssFiles` step in `publish.js` has been removed as it is unnecessary with the forwarded files.

- [ ] Yes
- [X] No

## Checklist:

### Reminders

### Review  
- [ ] ~~Request that the changes are [UX reviewed](https://github.com/kirbydesign/designsystem/blob/master/.github/CONTRIBUTING.md/#ux-review) (only necessary if your PR introduces visual changes)~~


copy-scss-files.js
```diff
@@ -0,0 +1,46 @@
...
+
+function createForwardScssFile(sourceFilePath, rootDir, targetFilePath) {
+  const newFileName = targetFilePath + sourceFilePath.split(rootDir).pop();
+
+  const newFileFolder = path.dirname(newFileName);
+  if (!fs.existsSync(newFileFolder)) {
+    fs.mkdirSync(newFileFolder, { recursive: true });
+  }
+
+  const forwardLocation = sourceFilePath.replace('./libs', '@kirbydesign');
```
RasmusKjeldgaard, line 32:
```suggestion
  const forwardLocation = sourceFilePath.replace('./libs', '~@kirbydesign');
```
I believe there should be a `~` here so we target the styles in the node_modules package correctly.

If I add it to the generated files, it works!

jkaltoft (changes requested):
Noice!

Usually when you use [URLs with Sass](https://sass-lang.com/documentation/at-rules/use#finding-the-module) you do not include the extension, e.g., instead of

```
@forward "@kirbydesign/core/scss/_utils.scss";
```

it should output

```
@forward "@kirbydesign/core/scss/utils";
```

Also, when a file is named `index.scss` you can omit it, e.g.

```
@forward "@kirbydesign/core/scss/print/components/_index.scss";
```

should be

```
@forward "@kirbydesign/core/scss/print/components";
```

Is it possible to output proper URLs when `@forward`-ing?

copy-scss-files.js
```diff
@@ -0,0 +1,63 @@
...
+  return ext[ext.length - 1];
+}
+
+function createForwardRule(filePath, packageAlias, libsFolder) {
+  return pipe(
+    (fileName) => (fileName[0] === '_' ? fileName.slice(1) : fileName),
+    (fileName) => (fileName === 'index.scss' ? '' : fileName),
+    (fileName) =>
+      `@forward "~${path.dirname(filePath).replace(libsFolder, packageAlias)}/${fileName}";`
+  )(path.basename(filePath));
```
MadsBuchmann, line 25: remove appended slash if fileName is empty. Remove extension from fileName.

scripts/forward-scss-files.js
```diff
@@ -0,0 +1,64 @@
...
+  const targetFileDirName = path.dirname(targetFilePath);
+  if (!fs.existsSync(targetFileDirName)) {
+    fs.mkdirSync(targetFileDirName, { recursive: true });
+  }
+
+  const forwardRule = createForwardRule(sourceFilePath, packageAlias, sharedRootDir);
+  fs.writeFileSync(targetFilePath, forwardRule);
+}
+
+module.exports.forwardScssFiles = (sourceRootDir, targetRootDir, packageAlias, sharedRootDir) => {
```
jkaltoft, line 59: Same as mentioned before (object literal). Also totally optional.

scripts/publish.js
```diff
@@ -112,31 +118,27 @@ function copyReadme() {
...
-
-function copyScssUtilsFile() {
-  console.log('Copying SCSS utils file...');
-  const onlyScssFiles = (input) => ['', '.scss'].includes(path.extname(input));
-  return fs.copy(`${angularLibDir}/scss/_utils.scss`, `${distTarget}/scss/_utils.scss`, {
-    filter: onlyScssFiles,
+function forwardCoreScssFilesToDesignsystem() {
+  console.log('Forwarding core SCSS files to designsystem ...');
+  return new Promise((resolve) => {
+    forwardScssFiles(`${coreLibDir}/scss`, `${distTarget}/scss`, packageAlias, libsRootDir);
```
jkaltoft, line 124: Optional.

Just a thought: Would it be more "future proof" to

* rename this function to `forwardCoreScssFiles()` and
* pass the destination as argument

or would that be a premature optimization?

jkaltoft (changes requested):

jkaltoft (commented):
It seems like something is a bit buggy.

I did this:

1. Create 2 new files in `libs/core/src/scss`
    - `awesome-stylesheet.scss`
    - `_awesome-partial.scss`
2. `npm run publish`

image

This resulted in the following output in `dist/libs/designsystem/scss/`:

image

jkaltoft (commented):
Ok, this is weird. Just ran `npm run publish` again and now my dist directory looks like this:

image

It may be a local problem for me - maybe some npm caching - but you should probably try and see what happens when you do the same (add new stylesheets and partials).

scripts/forward-scss-files.js
```diff
@@ -0,0 +1,64 @@
...
+function createForwardRule(filePath, packageAlias, sharedRootDir) {
+  return pipe(
+    (fileName) => (fileName[0] === '_' ? fileName.slice(1) : fileName),
+    (fileName) => (fileName === 'index.scss' ? '' : fileName),
+    (fileName) => path.basename(fileName, '.scss'),
+    (fileName) => `~${path.dirname(filePath).replace(sharedRootDir, packageAlias)}/${fileName}`,
+    (url) => (url[url.length - 1] === '/' ? url.slice(0, -1) : url),
+    (url) => `@forward "${url}";`
+  )(path.basename(filePath));
+}
```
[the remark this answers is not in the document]
MadsBuchmann: I will go with the second one 😄

scripts/forward-scss-files.js
```diff
@@ -0,0 +1,64 @@
...
+  return [...filesWithExtInDirectory, ...filesWithExtInSubdirectories.flat()];
+}
+
+function createForwardScssFile(
+  sourceFilePath,
+  sourceRootDir,
+  targetRootDir,
+  packageAlias,
+  sharedRootDir
+) {
```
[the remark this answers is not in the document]
MadsBuchmann: Good idea - i will have it use an object literal.

scripts/publish.js
MadsBuchmann: I'm not particularly a fan of passing in arguments to the function as it isn't done for any of the other functions in the ".then" chain at the bottom `publish..js`. 

How about renaming the step to `forwardCoreScssFiles` but having it take no arguments. Instead we will just have a call to `forwardScssFiles` for each library we want the files forwarded to. 

The `forwardCoreScssFiles` can then look something like: 
```
function forwardCoreScssFilesToDesignsystem() {
  console.log('Forwarding core SCSS files ...');
  const locations = [{from: 'sourcePath', to: 'targetPath'}, {from: 'sourcePath2', to: 'targetPath2'}]
  return new Promise((resolve) => {
     locations.forEach(({sourcePath, targetPath}) => {...}
   }
...
``

Comment (MadsBuchmann) [reply scripts/publish.js:124]: Ended up doing the following: 
```javascript
function forwardCoreScssFiles() {
  console.log('Forwarding core SCSS files...');

  return new Promise((resolve) => {
    pathsToForwardCoreScssFilesTo.forEach((targetPath) => {
      forwardScssFiles(`${coreLibDir}/scss`, targetPath, packageAlias, libsRootDir);
      resolve();
    });
  });
}
```

`pathsToForwardCoreScssFilesTo` can be found at the top of the file - closer to the variables it references.

scripts/forward-scss-files.js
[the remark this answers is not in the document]
MadsBuchmann: I've refactored the function to make it easier readable

jkaltoft (approved):

PR #1991, opened 2022-01-12
[Enhancement] Create hover interaction state utility

### LINKED ISSUE #1935: [Enhancement] Interaction State Utility - Hover

## Describe the enhancement

I would like a styling utility that can be applied to all elements/components that can be hovered. Ideally it should be prepared for handling _all_ interaction states (hover, press, focus), but this issue should just implement hover.

This issue derives from the research made in #1866

Inspired by [Material Design's interaction states](https://m3.material.io/foundations/interaction-states), this utility should create a state layer that can have varying degrees of opacity. The color should be derived from its content or configurable, cf. Material Design's "On colors".

However, the actual implementation should not rely on an empty DOM element if possible. We prefer to implement the state layer as either a simple change of background color or as a pseudo element. More research is needed. 

Experiment with HSLA color, brightness() CSS filter, opacity, pseudo element solution etc.

- [x] Create new PoC
- [x] Should we offer option to configure opacity degree and "on color"?

This issue does not concern elements/components with non-solid backgrounds, e.g. image as background. Those cases should be handled individually when/if we bump into any issues in the future.

We should try to implement the solution using utility mixins as suggested in [comments on the research issue](https://github.com/kirbydesign/designsystem/issues/1866#issuecomment-984482799).

The solution could end up with 2 ways to implement state layers. We may need both implementations:

### Simple background color change

For all elements/components including self-closing/singleton/atomic ones.

- Can _not_ contain other components, e.g. `<input />` (text, checkboxes, radio buttons etc.)
- Implement solution as a calculated background color based on opacity and "on color"
- Utility can only be applied to elements/components with fully opaque backgrounds (solid color)

### Empty DOM element (or pseudo element)

📣 UPDATE: Implementation should be moved to a separate issue

For container elements/components.

- Can contain other components, e.g. `<div>` and `<kirby-card>`
- Implement an actual layer (DOM element) - see [PoC](https://codepen.io/jkaltoft/pen/gOGrEON?editors=0100)
- Utility can be applied to elements/components with _all_ kinds of backgrounds, including images, gradients and completely transparent backgrounds

## Subtasks

Add subtasks as they emerge during research.

Some "spin-off tasks" may be out of scope. Take them into consideration and either solve them, plan them for later or discard them (before closing this issue).

- [x] #1992
- [x] Create `apply-hover()` mixin
    - [x] Verify parameter validity
    - [x] Determine how to handle transparent background
- [x] Utilize `apply-hover()`, e.g., in `<Button>` component
    - Needs to be applied multiple places - every variant, ie., attention levels, danger, disabled etc.
- [ ] Organize/re-structure non-component stylesheets, e.g., using ITCSS or similar methodology
- [x] ~~Create color converter - e.g. as Sass function or mixin - that can at least convert HEX and RGB colors to HSL. Use the Sass color module's functions.~~
    - ~~Alternative A: Add HSL variants of color custom properties~~
    - ~~Alternative B: Add function that can create HSL variants where needed~~
- ~~Refactor CSS custom properties for colors and related Sass variables/functions/mixins~~
- [x] Create design tokens for "loudness"
- [ ] #2006
- [ ] #2008
- [ ] #2014
- [ ] #2015
- [x] #2016

<hr />

## Checklist:

### Refinement

- [x] Request that the issue is [UX refined](https://github.com/kirbydesign/designsystem/blob/master/.github/CONTRIBUTING.md/#ux-refinement); do not proceed until this is done.
- [x] Request that the issue is [tech refined](https://github.com/kirbydesign/designsystem/blob/master/.github/CONTRIBUTING.md/#tech-refinement); do not proceed until this is done.

### Implementation 

- [x] Make sure you have read: "[Before you get coding](https://github.com/kirbydesign/designsystem/blob/master/.github/CONTRIBUTING.md/#before-you-get-coding)".
- [x] Create a branch from the [master branch](https://github.com/kirbydesign/designsystem/tree/master) following our [branch naming convention](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Branch). 
- [x] Publish a WIP implementation to Github as a draft PR and ask for feedback. 
- [x] ~~Make sure you have implemented tests following the guidelines in: "[The good: Test](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Test)".~~
- [x] ~~Update the [cookbook](https://cookbook.kirby.design) with examples and showcases.~~

### Review

jkaltoft:

This PR closes #1935 

## What is the new behavior?

Not exactly new behavior, but hover styles have been modified - in particular how hover is applied to elements/components.

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


libs/core/src/scss/interaction-state/_hover.scss
```diff
@@ -0,0 +1,54 @@
+@use 'sass:color';
+@use 'sass:meta';
+@use '../utils';
+
+// TODO: Accessibility - Loudness must respect user preferences
+// See https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast
+// TODO: Await UX decision on default loudness and update
+$default-loudness: utils.get-loudness('loud');
+
+@mixin apply-hover($background-color, $color, $property, $loudness: null) {
```
MadsBuchmann, line 10: Is it possible to do something equivalent to this: 
```suggestion

@mixin apply-hover($background-color, $color, $property, $loudness: utils.get-loudness('loud')) {
```

or perhaps: 
```suggestion
$default-loudness: utils.get-loudness('loud');

@mixin apply-hover($background-color, $color, $property, $loudness: $default-loudness) {
```

libs/core/src/scss/interaction-state/_hover.scss
```diff
@@ -0,0 +1,54 @@
...
+  // transparent.
+  //
+  // Sidenote: The color transparent is defined as black with 0% opacity, i.e.,
+  // 100% transparent. Often declared as rgba(0, 0, 0, 0%) in browsers.
+  // ---------------------------------------------------------------------------
+  $relative-lightness: $default-loudness !default;
+
+  @if $loudness {
+    $relative-lightness: $loudness;
+  }
```
MadsBuchmann, line 36: If we can eliminate the use of `null` - we can probably let go of this if-clause 😄

libs/core/src/scss/interaction-state/_hover.scss
```diff
@@ -0,0 +1,54 @@
...
+
+  @if color.lightness($background-color) > color.lightness($color) {
+    // Decrease relative lightness
```
MadsBuchmann, line 39:
```suggestion
    // Reverse relative lightness
```

libs/core/src/scss/interaction-state/_hover.scss
```diff
@@ -0,0 +1,54 @@
...
+    $relative-lightness: $relative-lightness * -1;
+  }
+
+  cursor: pointer;
+  transition-property: background-color;
+  transition-duration: utils.get-transition-duration('quick');
+  transition-timing-function: utils.get-transition-easing('static');
+
+  @include utils.hover() {
+    // TODO: Accessibility - check that contrast is (still) sufficiently high
```
MadsBuchmann, line 49: TODO should be handled before we merge 😄

libs/core/src/scss/interaction-state/_hover.scss
MadsBuchmann: If i understand it correctly 🤷

libs/core/src/scss/interaction-state/_index.scss
```diff
@@ -0,0 +1 @@
+@forward 'hover';
```
MadsBuchmann, line 1: We should double check this does not cause ngcc to catch fire - barrel files of various kinds gave us some problems with that mystery bug some time ago \*thousand yard stare\*

libs/designsystem/src/lib/components/button/button.component.scss
```diff
@@ -70,8 +90,26 @@ $button-width: (
...
+  );
+
   &.destructive {
     --kirby-button-color: #{utils.get-color('danger')};
+
+    @include interaction-state.apply-hover(
+      $background-color: transparent,
+      $color: utils.get-color($variant: 'danger', $getValueOnly: true),
+      $property: '--kirby-button-background-color'
+    );
```
MadsBuchmann, line 112: Is it possible to define transparent as a named color variant such that we can avoid having to use `interaction-state.apply-hover` directly?

MadsBuchmann (changes requested):
Looking good! - i especially like the one-liner in `button.component.scss`: 
```scss
@include _button-hover('danger', 'danger-contrast');
```
Pointed out some todos that i think we should determine what to do with before merging. Also i got a couple of questions that i've commented as well. 

image

libs/core/src/scss/base/_functions.scss
```diff
@@ -133,6 +137,7 @@
   color: $text-color;
 }
 
+// TODO: Move hover() mixin to libs/core/src/scss/interaction-state/_hover.scss
```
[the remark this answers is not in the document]
jkaltoft: More a left-on-purpose TODO :)

It's not critical to do it now, but it's not a big thing.

libs/core/src/scss/interaction-state/_hover.scss
```diff
@@ -0,0 +1,54 @@
...
+$default-loudness: utils.get-loudness('loud');
+
+@mixin apply-hover($background-color, $color, $property, $loudness: null) {
+  @if meta.type-of($background-color) != color {
+    @error "$background-color must be a valid CSS color value. See https://developer.mozilla.org/en-US/docs/Web/CSS/color_value";
+  }
+
+  @if meta.type-of($color) != color {
+    @error "$color must be a valid CSS color value. See https://developer.mozilla.org/en-US/docs/Web/CSS/color_value";
+  }
```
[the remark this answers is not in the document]
jkaltoft: Hmmm, I don't see a DRY'er way to do it. It's two different conditions and two different error messages and they are not dependant on each other. I think it would require more control statements to merge them.

```
if A is not a color
   output "Error A"

if B is not a color
   output "Error B"
```

would become something like

```
if either A or B is not a color

    if it is A that is not a color
        output "Error A"

    else 

    if it is B that is not a color
        output "Error B"
```

which doesn't seem DRY 😁

Do you have any suggestions?

libs/core/src/scss/interaction-state/_hover.scss
```diff
@@ -0,0 +1,54 @@
+@use 'sass:color';
+@use 'sass:meta';
+@use '../utils';
+
+// TODO: Accessibility - Loudness must respect user preferences
```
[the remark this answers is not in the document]
jkaltoft: We can't handle it within this issue. But we can create an issue for the Accessibility Epic if it exists.

libs/core/src/scss/interaction-state/_hover.scss
jkaltoft: Hmm. Technically it is "reversed" as in "negate" or "flip the positive/negative sign". But it is supposed to tell the reader that "if background color is lighter than foreground color then REDUCE lightness". Reduce lightness = decrease relative lightness = make it darker.

I don't think "Reverse" is a correct term for this. Do you have other suggestions?

jkaltoft: Perhaps the comment should just be: 

`// Make it darker`

😁

libs/core/src/scss/interaction-state/_hover.scss
```diff
@@ -0,0 +1,54 @@
...
+// See https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast
+// TODO: Await UX decision on default loudness and update
```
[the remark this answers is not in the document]
jkaltoft: https://github.com/kirbydesign/designsystem/pull/1991#discussion_r791830823 👆🏻

libs/core/src/scss/interaction-state/_hover.scss
jkaltoft: Initially `$default-loudness` was a private member `$_default-loudness` but it had to be public so it could be exposed to e.g. `button.component.scss`. However, it's not being used outside of this module at the moment, so maybe it can be changed again.

I'll see what I can come up with.

libs/core/src/scss/interaction-state/_hover.scss
jkaltoft: Initiatially it was implemented without this condition, but I had to change it to accommodate different use cases.

I'll try to see if I can come up with something else, but I'm not sure there's another solution that covers everything.

If you have any other suggestions I'm all ears 👂🏻

libs/designsystem/src/lib/components/button/button.component.scss
jkaltoft: @RasmusKjeldgaard and I already discussed this. The conclusion was, that it wouldn't make sense having a named color variant called "transparent".

Think about it this way: Which values would you expect `transparent-contrast`, `transparent-tint` and `transparent-shade` to have? 🤔😉

RasmusKjeldgaard (commented):
Beautiful work. I've only added some comments to some of the existing discussions.

Do we intend to merge this to release-next as is, and then do the fixes described in https://github.com/kirbydesign/designsystem/issues/1935 before releasing?

libs/core/src/scss/interaction-state/_hover.scss
jkaltoft: As a sidenote, one of the recommended ways to make function/mixin arguments optional is to assign `null` as default value. However, in this case we _do_ have a reasonable default value we can assign instead.

Maybe there is a problem in having two mixins `_button-hover()` and `apply-hover()` that both need to handle:

1. `$loudness` has been assigned a proper value
1. `$loudness` has been assigned to `null`
1. `$loudness` has not been assigned

but I'm not completely sure I have a better solution.

I can move the default assignment to `_button-hover()` to ensure that we are always invoking `apply-hover()` with a proper $loudness argument. But there may be cases where we don't want to do that.

Btw, assigning `null` as argument value in Sass is not the same as leaving the argument out. The latter will make the preprocessor fail with an error.

I'll keep digging.

jkaltoft:
> Do we intend to merge this to release/next as is, and then do the sub-tasks described in #1935 before releasing?

I think so, but we first need to decide on what to release when. If we want to release button attention levels changes independently of the interaction state stuff, we should not merge this PR with `release/next` yet.

So, we're back at having to finish the discussion about release management 🙂

libs/designsystem/src/lib/components/button/button.component.scss
```diff
@@ -70,8 +90,26 @@ $button-width: (
   --kirby-button-color: #{utils.get-color('black')};
   --kirby-button-border-color: #{utils.get-color('medium')};
 
+  // NOTE: This won't work!
+  //
+  // @include _button-hover(transparent, 'primary-contrast');
+  //
+  // Because transparent is not defined as a named color variant.
+  // Use the apply-hover() mixin directly instead.
```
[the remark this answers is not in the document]
jkaltoft: https://github.com/kirbydesign/designsystem/pull/1991#discussion_r792409627

Regarding the loudness scale. I agree that the names should not be hostile 🙂

The different levels _has_ to have _some_ name and value, but it's not important what they are for now. That will be a UX decision. Since they haven't defined a loudness scale yet we can't use actual names and values. On the other hand, there is no need to stall development while we're waiting for their decision.

I'll change the names of the loudness levels now, but it seems a bit redundant. They could be named `marble`, `golf-ball`, `tennis-ball`, `soccer-ball` etc. - it doesn't really matter. They must be renamed anyway later. We don't know how many we need either. Probably 4-6 levels, but that's also a UX decision we're awaiting 🙂

So, to sum it up: We can't not use names and values for the loudness scale, because technically we need the scale. And we can't use the correct names and values, because they haven't been defined yet.

libs/core/src/scss/base/_functions.scss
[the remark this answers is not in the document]
jkaltoft: I've changed my mind about moving this mixin. I'd rather either not move it or wait until we walk through other components that use hover. Changes to Chip styles seem unrelated to this PR.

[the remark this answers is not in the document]
jkaltoft: I will remove the TODO comment instead.

libs/core/src/scss/interaction-state/_hover.scss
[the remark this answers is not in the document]
jkaltoft: Issue created: #2015

libs/core/src/scss/interaction-state/_hover.scss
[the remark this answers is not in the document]
jkaltoft: Issue created: #2016

libs/core/src/scss/interaction-state/_hover.scss
jkaltoft: Issue created: #2014

libs/core/src/scss/interaction-state/_hover.scss
```diff
@@ -0,0 +1,52 @@
...
+
+  $relative-lightness: $loudness !default;
+
+  @if color.lightness($background-color) > color.lightness($color) {
+    // Make it darker
+    $relative-lightness: $relative-lightness * -1;
+  }
+
+  cursor: pointer;
+  transition-property: background-color;
```
MadsBuchmann, line 43:
```suggestion scss
  transition-property: var(#{$property});
```

Hmmmm 🤔 💭

jkaltoft: Thank you. Let's do that 🤘🏻

libs/designsystem/src/lib/components/button/button.component.scss
```diff
@@ -44,11 +45,35 @@ $button-width: (
   }
 }
 
+// Mixin parameters must be a named color variant. CSS color values won't work.
+// See libs/core/src/scss/themes/_colors.scss
```
MadsBuchmann, line 49: This is just me being nitpicky; feel free to leave as is - but consider not referring directly to a file path in comments. 

If we ever change the location this comment will be invalid and we might not remember to update the comment 😄

MadsBuchmann: To throw an additional tool on top of the mountain of tools that constitutes a _modern_ (put into that word what you wish) frontend stack - a tool that makes it possible to reference files & objects would be kinda cool; failing if they do not exist anymore. 

// Mixin parameters must be a named color variant. CSS color values won't work.
// See: libs/core/src/scss/themes/_colors.scss$[file-ref]

MadsBuchmann (approved):
LGTM! Got a single comment but nothing  blocking. 

Gordon Ramsay Brilliant

libs/designsystem/src/lib/components/button/button.component.scss
jkaltoft: We will without a doubt most certainly forget to update the comment if we move the file to a different location 😁.

I'll remove line 49:

```
// See libs/core/src/scss/themes/_colors.scss
```

jkaltoft, line 49:
```suggestion
```

PR #2059, opened 2022-02-21
Refactor interaction state solution for hover

### LINKED ISSUE #2008: Refactor interaction state solution and apply hover
📣 Update: Implement state layer based solution

~~Other than `<button>` which components and elements should utilize the `apply-hover()` mixin?~~

~~A quick search for `:hover` says it's only the [calendar component](https://github.com/kirbydesign/designsystem/blob/45f0fd17d7f996319ab8ff5dff8c6e24f12f10df/libs/designsystem/src/lib/components/calendar/calendar.component.scss) and the [chip component](https://github.com/kirbydesign/designsystem/blob/45f0fd17d7f996319ab8ff5dff8c6e24f12f10df/libs/designsystem/src/lib/components/chip/chip.component.scss).~~

The chip components is (apparently) the only place we use the `theme-background()` mixin. Consider if there's room for improvement, i.e., can it be replaced by using the `apply-hover()` mixin?

~~Consider splitting this issue into at least 3 separate issues: Calender, Chip and Item + Dropdown.~~

### Button 
- [x] Apply state layer based solution to the kirby button directive.

### Calendar

- [x] Apply hover to `.day` elements
  - Must handle `selected` and `today` also
- [x] Check year navigator button (fixed by Dropdown?)
- [x] Check month navigator buttons
- Hover color on all of the above should be the same (consult UX)

### Chip

- [x] Remove use of `theme-background()` 
- [x] Apply hover using the `apply-hover()` mixin

### Item and Dropdown

- [x] Refactor hover on Items in Card in Dropdown
  - The hover effect is currently controlled by `ion-item`
  - Consider if `apply-hover()` can/should be used
  - ⚠️ What are the consequences if we can't use `apply-hover()`?
    - Do we want to use 3 (or more) different solutions for interaction states? 
    - Solution: Use separate mixin for Ionic components (part of current refactorization)
- Clean up old API/interface
  - [x] Remove `--kirby-item-background-hover` state related CSS custom properties
  - [x] Update Cookbook showcase

### Other Components

- [x] Page - Toolbar buttons etc.
- [x] Segmented Control - fix what doesn't get fixed by applying state layer to Chip
- [x] Slide Button
- [x] #2056
- [x] #2057
- [x] Improve suboptimal solutions for `ion-back-button` and `ion-segment-button` (in Page and Segmented Control)
- [x] FAB Button: Ionic variant of solution + transition mixin
- [x] #2016 

### Refactor interaction state solution and apply hover - Part 2

See #2028

- #2058
- #2089
- #2090
- #2091
- #2092

### Misc.

- [x] ♿️ Accessibility: Hide state layer
- [x] 🔲 Borders: Should state layer cover border or not?
  - See e.g. button attention level 3 vs other attention levels
- [x] Verify that there are no other places we should replace `:hover` with `apply-hover()`
  - ion-reorder - maybe later
- [x] 🚶 Walk through all hover cases with UX/UI - adjust on the fly as needed 
- [ ] (Optional) Move the `hover()` mixin to `libs/core/src/scss/interaction-state/_hover.scss`

**Note:** If this issue is a can of worms in terms of rewriting chip scss & theme-background mixin - it should be split into a seperate issue.

jkaltoft:

This PR closes #2008 

## What is the new behavior?

Introduces layer based solution for interaction states in general and for hover state in particular.

- [ ] Yes
- [ ] No

## Checklist:

### Reminders

### Review  


libs/designsystem/src/lib/components/button/button.component.scss
```diff
@@ -155,22 +133,26 @@ $button-width: (
   // Outline is applied on button border instead,
   // to keep the rounded shape:
   outline: none;
-  border: 1px solid var(--kirby-button-border-color, transparent);
+  border-width: var(--border-width, 0px);
```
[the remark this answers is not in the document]
MadsBuchmann: Alternatively replace 0px with 1px and delete line 110.

libs/designsystem/src/lib/components/button/button.component.ts
```diff
@@ -102,7 +102,8 @@ export class ButtonComponent implements AfterContentInit {
     this.hasText = !!this.elementRef.nativeElement.textContent;
     if (this.iconElementRef !== undefined && this.hasText) {
       this._isIconLeft =
-        this.elementRef.nativeElement.firstChild === this.iconElementRef.nativeElement;
+        this.elementRef.nativeElement.querySelector('.content-layer').firstChild ===
+        this.iconElementRef.nativeElement;
```
MadsBuchmann, line 106: Alternatively we could consider using viewChild for this. But i'm actually a bit in doubt whether that would improve anything. It just feels like the _angular way™️_ of doing it - which i think is a poor argument for doing it. 

Now i'm just rambling. 

Feel free to do as you wish, this is just my toughts on this line of code 😛 
```
@ViewChild('.content-layer', {read: ElementRef}) contentLayer: ElementRef<HTMLElement>; 
...
this.contentLayer.nativeElement.firstChild 
```

libs/designsystem/src/lib/components/chip/chip.component.scss
```diff
@@ -28,21 +30,51 @@
...
-    @include utils.theme-background(transparent);
-    @include selected-and-hover('black');
 
     @include utils.media('<=small') {
       --inline-margin: 0;
     }
+
+    min-width: 44px;
+    flex-shrink: 0;
+    justify-content: center;
```
MadsBuchmann, line 78: A bit nitpicky, but consider moving `min-width`, `flex-shrink`, `justify-content`, `--inline-margin` and `--inline-padding` to be before any of mixins and nested selectors in. I think they drown a bit being after all this.

libs/designsystem/src/lib/components/fab-sheet/fab-sheet.component.scss
```diff
@@ -9,6 +10,14 @@ $fab-sheet-margin: utils.size('s');
 // Ionic fab can be customized by using the css properties listed in the CSS Custom Properties section of the below docs page:
 // https://ionicframework.com/docs/api/fab-button
 ion-fab-button {
+  @include ionic.apply-hover;
+
+  // NOTE: Also see ion-back-button in page component and ion-segment-button in segmented control component
+  &::part(native)::after {
+    @include ionic.transition;
+  }
+  // end NOTE
```
MadsBuchmann, line 19: What should i note when looking at ion-back-button in page component and ion-segment-button in segmented control component? Consider elaborating.

libs/designsystem/src/lib/components/page/page.component.scss
```diff
@@ -65,10 +66,27 @@ ion-title {
...
 ion-back-button {
+  @include ionic.apply-hover('s');
+
   --color: #{utils.get-color('black')};
   --icon-font-size: 24px;
+
   height: utils.$fat-finger-size;
   width: utils.$fat-finger-size;
+
+  // NOTE: Also see ion-segment-button in segmented control component
```
MadsBuchmann, line 77: Same point as above comment regarding references to page file and this one.

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
```diff
@@ -64,7 +62,26 @@ ion-segment-button {
...
   --margin-start: 0;
   --margin-top: 0;
+
+  min-height: utils.size('xl');
+  font-weight: utils.font-weight('normal');
+  font-size: utils.font-size('s');
+  text-transform: none;
   margin: 0;
+
+  // NOTE: Also see ion-back-button in page component
```
MadsBuchmann, line 72: Same point as comment regarding references to page and segmented control component.

libs/designsystem/src/lib/components/slide-button/slide-button.component.scss
```diff
@@ -1,24 +1,45 @@
...
+$slide-button-container-bg-color: utils.get-color('primary');
+$slider-thumb-bg-color: utils.get-color('white');
+$slide-button-text-color: utils.get-color('primary-contrast');
+$total-slider-thumb-width: $slide-button-width + $slide-button-spacing * 2;
+$slide-button-container-radius: $slide-button-width * 0.5;
+$slide-button-text-padding: 0 $slide-button-container-radius 0 $total-slider-thumb-width;
 
 @mixin _slider-thumb() {
+  @include interaction-state.transition;
+  // This actually doesn't use the state layer and the custom properties - only the mixin "shell"
```
MadsBuchmann, line 19: This comment is a bit vague and i fear it will lose it's meaning over time. What is meant by "shell"? Why does it only use the mixin shell?

libs/designsystem/src/lib/components/slide-button/slide-button.component.scss
```diff
@@ -1,24 +1,45 @@
...
+$slide-button-container-width: 256px;
+$slide-button-spacing: 2px;
+$slide-button-width: 52px;
+$slide-button-container-height: $slide-button-width + 2 * $slide-button-spacing;
+$slide-button-container-bg-color: utils.get-color('primary');
+$slider-thumb-bg-color: utils.get-color('white');
+$slide-button-text-color: utils.get-color('primary-contrast');
+$total-slider-thumb-width: $slide-button-width + $slide-button-spacing * 2;
+$slide-button-container-radius: $slide-button-width * 0.5;
+$slide-button-text-padding: 0 $slide-button-container-radius 0 $total-slider-thumb-width;
```
MadsBuchmann, line 15: Should we remove the variables and just apply the values directly to the properties (those that are not used multiple places that is)?

MadsBuchmann (changes requested):
Looking good - i got some comments 😄

libs/core/src/scss/base/_variables.scss
```diff
@@ -199,10 +199,10 @@ $loading-overlay-backdrop-opacity: 0.8;
...
-  'moderate': 12%,
-  'loud': 16%,
-  'very-loud': 24%,
-  'extremely-loud': 32%,
+  's': 4%,
+  'm': 8%,
+  'l': 12%,
+  'xl': 16%,
+  'xxl': 24%,
+  'xxxl': 32%,
```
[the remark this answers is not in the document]
jkaltoft: Yes, this is the scale requested by UX

libs/core/src/scss/interaction-state/_layer.scss
```diff
@@ -0,0 +1,81 @@
...
+@use '../utils';
+
+// Put state layer above content layer as default.
+// You can change move the content layer on top by declaring e.g.:
+//
+//    --content-layer-z-index: calc(#{layer.$state-layer-z-index} + 1);
+//
+// in a component that uses interaction state layer.
+$_content-layer-z-index: utils.z('default');
+$state-layer-z-index: $_content-layer-z-index + 1;
```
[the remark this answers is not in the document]
jkaltoft: I don't think I agree entirely, but it's good you point it out. We should definitely have a discussion about it (outside of code review 😁). 

In my opinion, the `$z-layers` map is a double-edged sword: It's a good thing to keep related things together, but depending on the stacking contexts in which these values are used you won't get what you'd expect. An element with z-index of 1 _can_ be on top of another element with a z-index of 100 if it's in a context "above" the second one.

This means that you could have a layer where you use e.g. `avatarBadge` (2) and not be able to make e.g. `loading-overlay` (1001) "win" over it, which would be very confusing. It all depends on the stacking contexts and the relation between the different layers.

In this case, it's not relevant what the actual values are, as long as `$state-layer-z-index` has a higher value than `$_content-layer-z-index`. They are not related or dependant on any of the other values in the `$z-layers` map. This is why they are only declared locally.

libs/core/src/scss/interaction-state/_layer.scss
```diff
@@ -0,0 +1,81 @@
...
+}
+
+@mixin extend-content {
+  .content-layer {
+    @content;
+  }
+}
+
+/// Apply streamlined transition to interaction state changes
+/// @param {*} $property [all]
```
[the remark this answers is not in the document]
jkaltoft: It's [Sassdoc](http://sassdoc.com/). Also see https://sass-lang.com/documentation/syntax/comments#documentation-comments

@RasmusKjeldgaard and I discussed it on Slack, but that was apparently in a DM and not in the team-kirby channel. Sorry about that 🙂

libs/core/src/scss/interaction-state/ionic/_index.scss
```diff
@@ -0,0 +1,2 @@
+@forward '../layer' show transition;
```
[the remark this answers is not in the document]
jkaltoft: The custom "Ionic hover module" only needs the `transition` mixin from `layer` so we hide the rest.

See https://sass-lang.com/documentation/at-rules/forward#controlling-visibility

libs/core/src/scss/interaction-state/_layer.scss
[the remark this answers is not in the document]
jkaltoft: Another thing is, that `$_content-layer-z-index` must be a private member while `$state-layer-z-index` must be public. That's not possible if we use the `$z-layers` map.

Yet another thing 🙂, neither variable should be open for modification. Since `$z-layers` is declared using `!default` I think you'd be able to change the value in the map.

libs/core/src/scss/interaction-state/_layer.scss
```diff
@@ -0,0 +1,81 @@
...
+  position: relative;
+
+  .content-layer {
+    position: relative;
+    z-index: var(--content-layer-z-index, $_content-layer-z-index);
+  }
+
+  .state-layer {
+    position: absolute;
+    inset: calc(var(--border-width, 0px) * -1); // negate border-width on parent
```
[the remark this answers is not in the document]
jkaltoft: I think that the "somewhere" where we note it down should be the codebase itself 🙂

libs/designsystem/src/lib/components/button/button.component.scss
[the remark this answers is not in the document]
jkaltoft: I think this might be due to some experiment not getting cleaned up properly 🙂

`--border-width` is necessary for calculating the `inset` value in the `layer` module (see your comment https://github.com/kirbydesign/designsystem/pull/2059#discussion_r821570110), so we can't delete line 110.

libs/designsystem/src/lib/components/button/button.component.ts
jkaltoft: I'm not very familiar with `@ViewChild`. I also don't really care how it's implemented as long as I'm testing the correct DOM element 😁, so I'm totally open for suggestions.

libs/designsystem/src/lib/components/fab-sheet/fab-sheet.component.scss
jkaltoft: If we could add hyperlinks to code it would be great. The note is just an attempt to refer to the other places we use the same custom implementation of hover.

I hope we can consolidate this somehow when we finish the remaining issues for hover, where we need special solutions for different ways to make it work for Ionic components.

libs/designsystem/src/lib/components/page/page.component.scss
jkaltoft: Same answer as above 🙂

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
jkaltoft: Same answer as above 🙂

libs/designsystem/src/lib/components/slide-button/slide-button.component.scss
jkaltoft: The shell is the part of the mixin that wraps everything else (another mixin). I think it's hard to describe without too many code specific details, but if have any suggestions let me know 🙂

The point of the comment is to make it explicit that it doesn't use the state layer or the custom properties even though you apply the same transition as in other components that _do_ use the state layer.

Another option is to leave out the comment, but then there's a risk that readers won't catch the fact. Maybe that's ok? I don't know 🙂

Let me know if you think it makes more sense to leave the comment out and not be explicit about the details.

libs/designsystem/src/lib/components/slide-button/slide-button.component.scss
jkaltoft: This block was moved out of https://github.com/kirbydesign/designsystem/blob/main/libs%2Fdesignsystem%2Fsrc%2Flib%2Fcomponents%2Fslide-button%2Fslide-button.component.shared.scss because it was only used in one single place. I didn't want to rewrite it in the first place, because it would look like something introduced by this PR.

It's not related to hover and interaction states so I'd prefer not to touch it unless it's very important. We should also have a look at why we're using hardcoded values here and see if there's a better solution - which is very out of scope, I think 😁

libs/core/src/scss/interaction-state/_layer.scss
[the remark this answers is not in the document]
MadsBuchmann: Good points! 

Regarding storing the z-index values in a maps, i've been wondering myself if it makes sense due to stacking contexts. So i agree with what you're saying. 

There might have been a _kings tanks_ (google translates version of "kongstanke") with the z-layers that could be interesting to see if we can track down. Otherwise we just have to form it in our own image 👼 

Feel free to resolve this conversation!

libs/core/src/scss/interaction-state/_layer.scss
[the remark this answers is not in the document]
MadsBuchmann: "Somewhere" has been successfully found! I agree 😄

libs/core/src/scss/interaction-state/ionic/_index.scss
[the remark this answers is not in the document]
MadsBuchmann: Those Sass modules just keeps on giving - i love it!

libs/designsystem/src/lib/components/button/button.component.ts
MadsBuchmann: `@ViewChild` is Angulars mechanism for injecting references to custom components & DOM elements in the hosts template (great 10 minute read about it [here](https://blog.angular-university.io/angular-viewchild/)). 

I think the benefit of utilizing ViewChild is that it's more explicit which elements the controller expects to be present in the template.

libs/designsystem/src/lib/components/chip/chip.component.scss
MadsBuchmann: That comment made my brain grow a bit! Thanks :D

libs/designsystem/src/lib/components/fab-sheet/fab-sheet.component.scss
MadsBuchmann: Aaaah i see! I took a stab at reformulating it :) 
```suggestion
  /* 
  NOTE: this is a custom implementation of the hover interaction state; 
  identical to PageComponent's ion-back-button & SegmentedControlComponent's 
  ion-segment-button. 
  */
  &::part(native)::after {
    @include ionic.transition;
  }
  // end NOTE
```

MadsBuchmann (approved):
LGTM!

PR #2093, opened 2022-03-07
Allow non-enum Input in components

### LINKED ISSUE #1985: [Enhancement] Enable strictTemplate checks

## Describe the enhancement
The `fullTemplateTypeCheck` flag that we are using in our library configuration (tsconfig.lib.json) will be deprecated in Angular 13, and be replaced by the `strictTemplates` flag.

With this check enabled in our current repo, we will see errors where we try to use strings for inputs that are actually typed, e.g. 'sm' for the size parameter on the button, that has the type ButtonSize. 
Same goes for ThemeColor, other sizes like AvatarSize and so on.

I think in many of these cases, it should be possible to use a string as input value, but that is not possible if we or any projects enable the strictTemplates flag, which will probably be the case for almost everyone when upgrading to Angular 13. 

Change the enum to a string union-type or a union of strings and the specific size/themecolor enums that already exist.
Beware of breaking changes in this regard. 

<hr />

## Checklist:

### Refinement

- [ ] Request that the issue is [UX refined](https://github.com/kirbydesign/designsystem/blob/master/.github/CONTRIBUTING.md/#ux-refinement); do not proceed until this is done.
- [ ] Request that the issue is [tech refined](https://github.com/kirbydesign/designsystem/blob/master/.github/CONTRIBUTING.md/#tech-refinement); do not proceed until this is done.

### Implementation 

### Review

MortenGregersen:

This PR closes #1985

## What is the new behavior?

Everywhere I could find a component with an enum-based `Input`, the type declaration is changed so it now also takes one of the cases of the enum as string. One exception is `ButtonSize`, where the name and value of the enum cases aren't the identical. In that case, the enum cases as strings are added to the union directly.

This change should not be a breaking change, as it just adds to the list of types/values that can be used on `Input` properties.

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  

### DIFF
--- a/libs/designsystem/src/lib/components/avatar/avatar.component.ts
+++ b/libs/designsystem/src/lib/components/avatar/avatar.component.ts
@@ -21,7 +21,7 @@ export class AvatarComponent {
   @Input() shadow: boolean;
   @Input() text: string;
   @Input() overlay: boolean;
-  @Input() size: AvatarSize | keyof AvatarSize = AvatarSize.SM;
+  @Input() size: AvatarSize | 'xs' | 'sm' | 'md' | 'lg' = AvatarSize.SM;
   @Input()
   themeColor: NotificationColor | BrandColor | 'medium' | 'white' | 'dark' | 'light' | 'semi-light';
 
--- a/libs/designsystem/src/lib/components/avatar/avatar.component.ts
+++ b/libs/designsystem/src/lib/components/avatar/avatar.component.ts
@@ -21,7 +21,7 @@ export class AvatarComponent {
   @Input() shadow: boolean;
   @Input() text: string;
   @Input() overlay: boolean;
-  @Input() size: AvatarSize = AvatarSize.SM;
+  @Input() size: AvatarSize | keyof AvatarSize = AvatarSize.SM;
   @Input()
   themeColor: NotificationColor | BrandColor | 'medium' | 'white' | 'dark' | 'light' | 'semi-light';
 
--- a/libs/designsystem/src/lib/components/form-field/input/input.component.ts
+++ b/libs/designsystem/src/lib/components/form-field/input/input.component.ts
@@ -38,7 +38,7 @@ export class InputComponent implements OnChanges {
 
   @HostBinding('class')
   @Input()
-  size: InputSize = InputSize.large;
+  size: InputSize | keyof InputSize = InputSize.large;
 
   /**
    * Removes padding, width, rounded borders and drop-shadow when set to `true`.
--- a/libs/designsystem/src/lib/components/segmented-control/segmented-control.component.ts
+++ b/libs/designsystem/src/lib/components/segmented-control/segmented-control.component.ts
@@ -28,7 +28,7 @@ export class SegmentedControlComponent {
     }
   }
 
-  @Input() mode: SegmentedControlMode = SegmentedControlMode.default;
+  @Input() mode: SegmentedControlMode | keyof SegmentedControlMode = SegmentedControlMode.default;
 
   @HostBinding('class')
   get _modeCssClass() {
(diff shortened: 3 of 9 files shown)


libs/designsystem/src/lib/components/avatar/avatar.component.ts
```diff
@@ -21,7 +21,7 @@ export class AvatarComponent {
   @Input() shadow: boolean;
   @Input() text: string;
   @Input() overlay: boolean;
-  @Input() size: AvatarSize = AvatarSize.SM;
+  @Input() size: AvatarSize | keyof AvatarSize = AvatarSize.SM;
```
MadsBuchmann, line 24: Unfortunately `keyof AvatarSize` returns all of the keys for a string object. 

I've created an example here: [link to typescript playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBMAdgVwLZwIIDcCGMdQDKAlgF7BwDeAUHHABqFwC8cA5CAM5sA0tchALIt2nVL36CAIiLaoAJhLoAZAOKyANgHMJAX2rUYATzAUAKsE4wzJiq2x4CJcnAA+cANbAjEAGaZcfCIyYABuIA).

Notice the type of `TestType`
Screenshot 2022-03-08 at 14 52 58

I think we need something like `valueof AvatarSize`. 

This comment goes for all of the places where `keyof` has been used.

MadsBuchmann (changes requested):
Thank you for the PR! 🎉 
Unfortunately I can't approve it yet - see comment.

libs/designsystem/src/lib/components/avatar/avatar.component.ts
MortenGregersen: It is only the case where the case in the enum maps to a string. I already fixed it for `ButtonSize`, but didn't see it for `AvatarSize`. I found 2 others.

libs/designsystem/src/lib/components/form-field/input/input.component.ts
```diff
@@ -38,7 +38,7 @@ export class InputComponent implements OnChanges {
 
   @HostBinding('class')
   @Input()
-  size: InputSize = InputSize.large;
+  size: InputSize | keyof InputSize = InputSize.large;
```
MadsBuchmann, line 41: Resolves to: 
```typescript
(property) InputComponent.size: number | InputSize | "toString" | "charAt" | "charCodeAt" | "concat" | "indexOf" | "lastIndexOf" | "localeCompare" | "match" | "replace" | "search" | "slice" | "split" | ... 29 more ... | "padEnd"
```

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.ts
```diff
@@ -28,7 +28,7 @@ export class SegmentedControlComponent {
     }
   }
 
-  @Input() mode: SegmentedControlMode = SegmentedControlMode.default;
+  @Input() mode: SegmentedControlMode | keyof SegmentedControlMode = SegmentedControlMode.default;
```
MadsBuchmann, line 31: Resolves to: 
```
(property) SegmentedControlComponent.mode: number | "toString" | "charAt" | "charCodeAt" | "concat" | "indexOf" | "lastIndexOf" | "localeCompare" | "match" | "replace" | "search" | "slice" | "split" | "substring" | "toLowerCase" | ... 28 more ... | SegmentedControlMode
```

MadsBuchmann (commented):
I've marked some values which resolves wrongly. 

But in general i do not think creating a union type by explicitly writing out the string values of the enum is a good solution as it introduces duplication. 

I've however discovered that we might be able to use [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to solve this!

I've taken the liberty of using your branch as the base branch for [a new PR](https://github.com/kirbydesign/designsystem/pull/2099/files) resolving your issue using those. 

Unfortunately our version of Prettier does not support template literal types as it is quite old (1.19.1)! I will discuss with the rest of the team tomorrow if we should look into upgrading prettier.

PR #2112, opened 2022-03-17
Interaction state hover for links

### LINKED ISSUE #2104: Interaction State Hover for links
- [x] Use pseudo element for state layer
- [x] Be careful not to break `accessible-target-size` mixin - should probably use `::before` for state layer
- [x] Must play nice with :focus styles
- [x] Easy opt-out: Create module + mixin
- [x] Consider to move global link styles into separate module (similar to `html-lists.scss`)

jkaltoft:

This PR closes #2104 

- [x] Yes
- [ ] No

Since hover effect is being applied with at state layer (a pseudo element) it may interfere with links that are styled to not look like regular text hyperlinks. Host applications need to handle this by actively opting-out of the default link styling. For that purpose an opt-out Sass module is exposed.

## Checklist:

### Reminders

### Review  

### DIFF
--- a/libs/core/src/scss/base/_functions.scss
+++ b/libs/core/src/scss/base/_functions.scss
@@ -113,6 +113,18 @@
   @return $classes;
 }
 
+/// Remove the unit of a length
+/// @param {Number} $number - Number to remove unit from
+/// @return {Number} - Unitless number
+/// Source: https://css-tricks.com/snippets/sass/strip-unit-function/
+@function strip-unit($number) {
+  @if meta.type-of($number) == 'number' and not math.is-unitless($number) {
+    @return $number / ($number * 0 + 1);
+  }
+
+  @return $number;
+}
+
 @mixin slotted($selectors...) {
   /* stylelint-disable-next-line selector-pseudo-element-no-unknown */
   ::ng-deep > {
--- a/libs/core/src/scss/base/_functions.scss
+++ b/libs/core/src/scss/base/_functions.scss
@@ -113,18 +113,6 @@
   @return $classes;
 }
 
-/// Remove the unit of a length
-/// @param {Number} $number - Number to remove unit from
-/// @return {Number} - Unitless number
-/// Source: https://css-tricks.com/snippets/sass/strip-unit-function/
-@function strip-unit($number) {
-  @if meta.type-of($number) == 'number' and not math.is-unitless($number) {
-    @return $number / ($number * 0 + 1);
-  }
-
-  @return $number;
-}
-
 @mixin slotted($selectors...) {
   /* stylelint-disable-next-line selector-pseudo-element-no-unknown */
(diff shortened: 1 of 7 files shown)


RasmusKjeldgaard:
I think we should push back a bit on this styling. It seems _off_ to me, and the usability is not great - it looks a bit like a button when hovering, and the color is too similar to the background.

It does not really follow any 'conventions' I've seen before. I would much prefer a different text color and perhaps adding/removing the underline when hovering, which seems to be a very used pattern.

RasmusKjeldgaard (changes requested):
I think we should push back a bit on this styling. It seems off to me, and the usability is not great - it looks a bit like a button when hovering, and the color is too similar to the background.

It does not really follow any 'conventions' I've seen before. I would much prefer a different text color and perhaps adding/removing the underline when hovering, which seems to be a very used pattern.

jkaltoft:
> I think we should push back a bit on this styling. It seems off to me, and the usability is not great - it looks a bit like a button when hovering, and the color is too similar to the background.
> 
> It does not really follow any 'conventions' I've seen before. I would much prefer a different text color and perhaps adding/removing the underline when hovering, which seems to be a very used pattern.

I agree. Let's discuss this with @henrikvoetmand before proceeding with this PR

libs/core/src/scss/base/_functions.scss
```diff
@@ -112,6 +113,18 @@
...
+/// @return {Number} - Unitless number
+/// Source: https://css-tricks.com/snippets/sass/strip-unit-function/
+@function strip-unit($number) {
+  @if meta.type-of($number) == 'number' and not math.is-unitless($number) {
+    @return $number / ($number * 0 + 1);
+  }
+
+  @return $number;
+}
+
```
RasmusKjeldgaard, line 127: Is this still needed?

RasmusKjeldgaard: Oops sorry, was too quick. Just got a notification, and thought you had requested a review, but it was just changes to this PR. I'll wait until you request a review.

jkaltoft: You may be right. I'll just check if it's still being used.

jkaltoft: Nope. It's not needed after things got cleaned up.

I _do_ think it's a very useful function to have. Should we just leave it? We can create a small issue for adding it instead, if you prefer.

jkaltoft: I've removed it and created [a new issue](https://github.com/kirbydesign/designsystem/issues/2162)

RasmusKjeldgaard (approved):
I really like this solution! 👍 
Code looks good.

jkaltoft:
Cookbook build keeps failing. Attempt to fix by opening a new PR.

PR #2123, opened 2022-03-23
Hover solution for non solid backgrounds in card

### LINKED ISSUE #2006: Create hover solution for non-solid backgrounds in card (pseudo element and opacity)
With interaction state (#1935) we've tried to avoid creating an additional element in the DOM to alter the background on elements when hovering. 

Some places we have multi-colored backgrounds such as a card with an image background - this requires an additional element in the DOM, as we can't just alter the background color as is done in #1935. 

Instead we should put an additional layer on top to alter the background-color; inspired by material design interaction layers. 

A general solution for doing this should be created. 

### Questions: 
- [x] Should #2022 be carried out? If not, just close this issue.    
- [x] How do we toggle whether we use solution from #1935 or the new one where we introduce a new element to color the elements. There might not exist a pretty solution :-) 
   - Should we use an input property to toggle, css custom property?

### Requirements: 
- If it is interactive (i.e. click event is given) & it has an image background image: use new solution with additional element in DOM. 
- If it is interactive & it has a solid background: use solution from #1935 

### Blocked by: 
- [x] #2022

jkaltoft:

This PR closes #2006 

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  

### DIFF
--- a/libs/designsystem/src/lib/components/card/card.component.html
+++ b/libs/designsystem/src/lib/components/card/card.component.html
@@ -1,5 +1,10 @@
-<ng-content select="kirby-card-header"></ng-content>
-<div class="content-wrapper" [class.padding]="hasPadding">
-  <ng-content></ng-content>
-</div>
-<ng-content select="kirby-card-footer"></ng-content>
+<ng-container>
+  <div class="state-layer" aria-hidden="true"></div>
+  <div class="content-layer">
+    <ng-content select="kirby-card-header"></ng-content>
+    <div class="content-wrapper" [class.padding]="hasPadding">
+      <ng-content></ng-content>
+    </div>
+    <ng-content select="kirby-card-footer"></ng-content>
+  </div>
+</ng-container>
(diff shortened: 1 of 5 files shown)


libs/designsystem/src/lib/components/card/card.component.html
```diff
@@ -1,5 +1,10 @@
-<ng-content select="kirby-card-header"></ng-content>
-<div class="content-wrapper" [class.padding]="hasPadding">
-  <ng-content></ng-content>
-</div>
-<ng-content select="kirby-card-footer"></ng-content>
+<ng-container>
```
RasmusKjeldgaard, line 1: Does this `ng-container` have any use when we are not adding a structural directive or similar?

RasmusKjeldgaard, line 1: Does this `ng-container` have any use when we are not adding a structural directive or similar?

RasmusKjeldgaard (changes requested):

RasmusKjeldgaard (approved):

PR #2140, opened 2022-03-28
Update Angular to v12

### LINKED ISSUE #1931: [Housekeeping] Upgrade to Angular 12

We need to upgrade to Angular 12, to support the new Ionic 6 version.
See more here: https://ionicframework.com/docs/next/intro/upgrading-to-ionic-6

No, we should definitely do this! 😄

RasmusKjeldgaard:

This PR closes #1931 

- [x] Yes
- [ ] No

This is a breaking change because we move to Ivy partial compilation, and this means that older projects that still use the Legacy View Engine (which will be removed in Angular 13) might run into trouble. 

I have surveyed around in DRB (including micro frontends) and JB-only projects, and they all use Ivy already. 

## Checklist:

### Reminders

### Review  


libs/designsystem/testing-base/src/lib/components/index.ts
```diff
@@ -0,0 +1,67 @@
+//TODO: This needs to be auto-generated!
```
RasmusKjeldgaard, line 1: Actually, I think this should be a separate issue. 

What we want to do is to expose our mocks via a barrel file like this, because Ivy complains that these components are exposed through our TestingModule, without also being exported via a barrel file. 

We do this already for our KirbyModule, in https://github.com/kirbydesign/designsystem/blob/main/libs/designsystem/src/lib/index.ts (also hand-held), and something similar should happen here, but we should be able to easily automate it with our existing mock generation script.

mictro: Agree!

libs/designsystem/src/lib/index.ts
```diff
@@ -5,9 +5,4 @@ export * from './helpers';
 export * from './scss/scss-helper';
 // export * from './testing';
 
-// Export BadgeComponent here instead of in ./components as we will otherwise
-// have an UMD compilation error when compiling the Kirby bundle in other
-// Angular projects. See https://github.com/kirbydesign/designsystem/issues/1740.
-export { KirbyBadge as BadgeComponent } from './components/web-component-proxies.component';
-
```
MadsBuchmann, line 12: JOY!

MadsBuchmann:
LGTM! 
Tim & Eric LGTM

MadsBuchmann (approved):
LGTM! 
Tim & Eric LGTM

PR #2146, opened 2022-03-29
Upgrade prettier and stylelint

### LINKED ISSUE #2103: [Housekeeping] Upgrade Prettier

I would like to upgrade our version of prettier to the newest version. 

As part of #2099 i had to use [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html). This however gives me a Prettier error when trying to commit due to using the template literal types. The code in #2099 had to be committed and pushed with `--no-verify`. 

There exists [an issue in the Prettier repository](https://github.com/prettier/prettier/issues/11533) which reports the same behavior i'm experiencing. According to that issue, it is resolved by upgrading Prettier :-)

RasmusKjeldgaard:

This PR closes #2103

## What is the new behavior?
Prettier has been updated to the newest version, and `prettier --write` has been run on the entire repository.

Furthermore, stylelint is upgraded to match, and has likewise been run on the entire repository. A new config is added in `.stylelintrc` for this newer version, based on what is recommended in the stylelint docs for scss: https://stylelint.io/user-guide/get-started/#linting-everything-else

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


libs/core/src/scss/_global-styles.scss
```diff
@@ -69,25 +72,28 @@ ion-modal.kirby-overlay {
...
     box-sizing: border-box;
+
     --background: var(--kirby-modal-background, #{utils.get-color('background-color')});
     --border-radius: 0;
     --height: auto;
 
     &.kirby-modal-compact {
       --border-radius: #{utils.$border-radius};
       --max-width: #{utils.$compact-modal-max-width};
+
```
jkaltoft, line 83: Love it! I'm a strong proponent of these line breaks. Just had to mention it 😁

jkaltoft:
More review comments coming up 🙂

libs/core/src/scss/base/_line-clamp.scss
```diff
@@ -36,7 +36,7 @@
 // See https://caniuse.com/mdn-css_properties_-webkit-line-clamp
 @supports (-webkit-line-clamp: 1) {
   .kirby-line-clamp {
-    display: -webkit-box;
+    display: box;
```
jkaltoft, line 39: I don't believe `box` is a valid value for the `display` property.

What's even worse, line-clamp will not work in any browsers without  `display: -webkit-box`.

Do you know what setting/configuration triggers this change? We _do_ have a `.browserslistrc` file. Maybe stylelint/prettier uses it to determine which corrections to apply...?

jkaltoft:
```suggestion
    display: -webkit-box;
```

jkaltoft (changes requested):

libs/core/src/scss/base/_list.scss
```diff
@@ -44,27 +46,27 @@
  */
 
 .rounded {
-  @include rounded();
+  @include rounded;
```
jkaltoft, line 49: Thank you 🙏🏻

libs/core/src/scss/base/_variables.scss
```diff
@@ -71,15 +70,14 @@ $icon-font-sizes: (
 ****************************************************************************/
 $z-layers: (
   default: 1,
-  segmentBadge: 2,
-  avatarBadge: 2,
+  segment-badge: 2,
+  avatar-badge: 2,
```
jkaltoft, line 74: Nice 🙌🏻

jkaltoft: Have usage been updated everywhere?

jkaltoft (changes requested):

apps/cookbook/src/app/examples/modal-example/modal-route-example/modal-route-page1-example.component.ts
```diff
@@ -8,8 +8,8 @@ import { ActivatedRoute } from '@angular/router';
     <h4>The standard Lorem Ipsum passage, used since the 1500s</h4>
     <span
       >QueryParams:
-      <pre>{{ queryParams$ | async | json }}</pre></span
-    >
+      <pre>{{ queryParams$ | async | json }}</pre>
+    </span>
```
jkaltoft, line 12: Wow. That was ugly before 🙂

But line 9-10 still look a bit odd. I would expect the `>` to be closing the `span` on line 9.

libs/designsystem/src/lib/components/dropdown/dropdown.component.ts
```diff
@@ -44,7 +44,8 @@ import { KeyboardHandlerService } from './keyboard-handler.service';
   ],
 })
 export class DropdownComponent
-  implements AfterContentChecked, AfterViewInit, OnDestroy, ControlValueAccessor {
+  implements AfterContentChecked, AfterViewInit, OnDestroy, ControlValueAccessor
+{
```
jkaltoft, line 48: Is this correct?

libs/designsystem/src/lib/components/fab-sheet/fab-sheet.component.ts
```diff
@@ -1,21 +1,21 @@
...
+  ViewChild,
 } from '@angular/core';
-import { DOCUMENT } from '@angular/common';
 import { IonFab, IonFabButton, IonIcon } from '@ionic/angular';
 
-import { ActionSheetComponent } from '../modal/action-sheet/action-sheet.component';
-import { kirbyIconSettings } from '../icon/kirby-icon-settings';
 import { Icon } from '../icon/icon-settings';
+import { kirbyIconSettings } from '../icon/kirby-icon-settings';
+import { ActionSheetComponent } from '../modal/action-sheet/action-sheet.component';
```
jkaltoft, line 18: Great. Do we have any control over how imports are sorted?

libs/designsystem/src/lib/components/fab-sheet/fab-sheet.component.ts
```diff
@@ -62,7 +61,7 @@ export class FabSheetComponent implements AfterContentInit, AfterViewInit {
       const closeIcon = fabButtonElement.shadowRoot.querySelector('.close-icon ion-icon');
       if (closeIcon) {
         const closeIconSvgLoaded = closeIcon.shadowRoot.querySelector('.icon-inner svg');
-        const ionCloseIcon = (closeIcon as unknown) as IonIcon;
+        const ionCloseIcon = closeIcon as unknown as IonIcon;
```
jkaltoft, line 64: I did not know you could chain `as`s (so to speak 😁)

I think I like it better _with_ the parenthesises.

Is there any technical difference? How are the precedence rules for multiple `as`s?

libs/designsystem/src/lib/components/flag/flag.component.ts
```diff
@@ -1,10 +1,8 @@
-import { Component, Input, ChangeDetectionStrategy, HostBinding } from '@angular/core';
+import { ChangeDetectionStrategy, Component, HostBinding, Input } from '@angular/core';
 
 @Component({
   selector: 'kirby-flag',
-  template: `
-    <ng-content></ng-content>
-  `,
+  template: ` <ng-content></ng-content> `,
```
jkaltoft, line 5: I suppose the spaces don't make any difference. Can you confirm?

apps/cookbook/src/app/iphone/iphone.component.html
```diff
@@ -1,7 +1,7 @@
 <div class="container">
   <div class="docs-demo-device ios">
     <figure>
-      <svg class="docs-demo-device__md-bar" viewBox="0 0 1384.3 40.3">
+      <svg class="docs-demo-device-md-bar" viewBox="0 0 1384.3 40.3">
```
jkaltoft, line 4: Technically this seems to be ok - since the corresponding CSS class selector is also renamed. However, it may or may not be what we want. We haven't discussed naming conventions like BEM, have we?

apps/cookbook/src/app/iphone/iphone.component.scss
```diff
@@ -1,5 +1,6 @@
 @use '~@kirbydesign/core/src/scss/utils';
 
+/* stylelint-disable custom-property-pattern */
```
jkaltoft, line 3: Why do we disable the `custom-property-pattern` rule?

apps/cookbook/src/app/page/header/header.component.scss
```diff
@@ -63,23 +63,24 @@ nav {
     height: 100%;
     width: 100%;
   }
-}
 
-.logo {
   .text {
```
jkaltoft, line 67: Wow! This is rather clever. Nice work, stylelint.

apps/cookbook/src/app/page/side-nav/side-nav.component.scss
```diff
@@ -65,7 +65,7 @@ section {
     &.is-hidden {
       opacity: 0;
       max-height: 0;
-      padding: 0 0;
+      padding: 0;
```
jkaltoft, line 68: Thank you 🙏🏻

apps/cookbook/src/app/showcase/fonts-showcase/fonts-showcase.component.scss
```diff
@@ -18,7 +18,7 @@ code {
   }
 
   code {
-    padding: 0.25rem 0 0 0;
+    padding: 0.25rem 0 0;
```
jkaltoft, line 21: Yes, please 🙏🏻

jkaltoft (changes requested):

libs/designsystem/src/lib/components/form-field/_form-field-inputs.shared.scss
```diff
@@ -61,8 +61,8 @@ $form-field-input-padding: utils.size('s');
   }
 
   // for the browsers that has partial support, see https://caniuse.com/#feat=css-placeholder
-  &::-webkit-input-placeholder,
-  &::-ms-input-placeholder {
+  /* stylelint-disable-next-line selector-pseudo-element-no-unknown */
+  &::input-placeholder {
```
jkaltoft, line 65: Removing `::-webkit-input-placeholder` is probably ok, but maybe the comment about partial support should be updated.

However, I think removing the `-ms` prefix is a mistake: There is no `::input-placeholder` pseudo element as far as I know. So the disable rule comment in line 64 should also be removed.

I'm a bit concerned, that this is another sign of stylelint being too hard on the "remove vendor specific prefixes" stuff.

One thing is to remove prefixes that are not needed. Another thing is to replace them with invalid CSS. We should try to learn how to avoid that.

libs/designsystem/src/lib/components/form-field/_form-field-inputs.shared.scss
```diff
@@ -36,7 +36,7 @@ $form-field-input-padding: utils.size('s');
   line-height: $form-field-input-line-height;
   outline: none;
   margin: 0;
-  -webkit-appearance: none;
+  appearance: none;
```
jkaltoft, line 39: I think this is safe, but see other comment below in this file.

libs/designsystem/src/lib/components/form-field/form-field.component.ts
```diff
@@ -40,9 +41,8 @@ export class FormFieldComponent
 
   @ContentChild(InputCounterComponent, { static: false }) counter: InputCounterComponent;
   @ContentChild(RadioGroupComponent) private radioGroupComponent: RadioGroupComponent;
-  @ContentChild(RadioGroupComponent, { read: ElementRef }) private radioGroupElement: ElementRef<
-    HTMLElement
-  >;
+  @ContentChild(RadioGroupComponent, { read: ElementRef })
+  private radioGroupElement: ElementRef<HTMLElement>;
```
jkaltoft, line 45: Disclaimer: I'm no Angular expert 😎

Is this correct? Is separating `private radioGroupElement` from `@ContentChild` equivalent to having them on the same line?

libs/designsystem/src/lib/components/icon/icon.component.scss
```diff
@@ -9,8 +9,8 @@
   margin-right: var(--kirby-icon-margin-right, unset);
 
   svg path {
-    fill: currentColor;
-    stroke: currentColor;
+    fill: currentcolor;
+    stroke: currentcolor;
```
jkaltoft, line 13: Also an error. The correct value is `currentColor`

libs/designsystem/src/lib/components/list/list.component.scss
```diff
@@ -169,7 +169,7 @@ ion-item-options[side='end'] {
   ion-list-header {
     border-top-left-radius: utils.$border-radius;
     border-top-right-radius: utils.$border-radius;
-    -webkit-mask-image: -webkit-radial-gradient(
+    mask-image: radial-gradient(
```
jkaltoft, line 172: Once again, we need to be completely sure that we want the prefixes removed.

jkaltoft (changes requested):

libs/designsystem/src/lib/components/modal/footer/modal-footer.component.scss
```diff
@@ -20,35 +20,38 @@ ion-footer {
...
 
-:host(.light) ion-footer {
-  background-color: utils.get-color('background-color');
-}
-
-:host(.inline) ion-footer {
-  background: transparent;
-  box-shadow: none;
+:host-context(.modal-wrapper.full-height) {
+  --kirby-modal-footer-safe-area-bottom: var(--kirby-safe-area-bottom);
```
jkaltoft, line 56: Do you know which rule triggers this reordering?

libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
```diff
@@ -57,18 +57,14 @@ export class ModalWrapperComponent implements Modal, AfterViewInit, OnInit, OnDe
...
-    HTMLIonTitleElement
-  >;
+  @ViewChild(IonContent, { static: true, read: ElementRef })
+  private ionContentElement: ElementRef<HTMLIonContentElement>;
+  @ViewChild(IonHeader, { static: true, read: ElementRef })
+  private ionHeaderElement: ElementRef<HTMLIonHeaderElement>;
+  @ViewChild(IonToolbar, { static: true, read: ElementRef })
+  private ionToolbarElement: ElementRef<HTMLIonToolbarElement>;
+  @ViewChild(IonTitle, { static: true, read: ElementRef })
+  private ionTitleElement: ElementRef<HTMLIonTitleElement>;
```
jkaltoft, line 67: Is it ok to do this?

jkaltoft (commented):
giphub

libs/designsystem/src/lib/components/form-field/form-field.component.ts
MadsBuchmann: Yep! It's the same thing 😄 

Fun fact: Actually decorators are an experimental typescript thingy and not Angular specific. 
image

MadsBuchmann (commented):
Started reviewing - but stopping my review again as i found out that @jkaltoft is on it B)

apps/cookbook/src/app/designer/card-designer/card-designer.component.scss
```diff
@@ -12,16 +12,21 @@ cookbook-card-example {
   display: inline-block;
   width: 257px;
   height: 362px;
+
   --kirby-card-background-image: url('/assets/kirby/backgrounds/kirby-card.png');
+
   &.show-background {
-    --kirby-example-wireframe-color: rgba(255, 0, 0, 0.5);
+    --kirby-example-wireframe-color: rgb(255 0 0 / 50%);
```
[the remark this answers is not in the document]
RasmusKjeldgaard: Yes, this is the default setting for `color-function-notation` https://stylelint.io/user-guide/rules/list/color-function-notation/, which stylelint apparently is quite opinionated about. At least they deem the comma-separated approach legacy.

apps/cookbook/src/app/examples/chart-deprecated-example/examples/chart-deprecated-example-areaspline.component.ts
```diff
@@ -30,29 +30,7 @@ export class ChartDeprecatedExampleAreasplineComponent {
...
-    1440.0,
-    1460.0,
-    1350.6,
-    1480.5,
-    1800.4,
-    1940.1,
-    1950.6,
-    1700.4,
+    1600.9, 1710.5, 1060.4, 1290.2, 1440.0, 1460.0, 1350.6, 1480.5, 1800.4, 1940.1, 1950.6, 1700.4,
+    1600.9, 1710.5, 1060.4, 1290.2, 1440.0, 1460.0, 1350.6, 1480.5, 1800.4, 1940.1, 1950.6, 1700.4,
```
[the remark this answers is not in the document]
RasmusKjeldgaard: It is a prettier default I believe. Set to 100, but I have not configured it myself.

apps/cookbook/src/app/iphone/iphone.component.html
RasmusKjeldgaard: This is changed for consistency across the code base. If we opt for something like BEM, I think we should revisit.

apps/cookbook/src/app/iphone/iphone.component.scss
RasmusKjeldgaard: This is because the custom properties are named something like `--text-color--dark`, with a double dash between color and dark. 

I don't think they are used anywhere though, could you double-check? Then we could just remove them completely, and get rid of the rule.

libs/designsystem/src/lib/components/slide-button/slide-button.component.scss
```diff
@@ -2,7 +2,7 @@
 @use 'slide-button.component.shared';
 
 @mixin _slider-thumb() {
-  -webkit-appearance: none;
+  appearance: none;
```
jkaltoft, line 5: I just found out, that apparently [Angular uses Autoprefixer](https://angular.io/guide/build#configuring-browser-compatibility) as part of the build, so letting stylelint remove vendor specific prefixes _should_ be safe in the Sass source files - as long as they get changed to standard properties.

✅ `-webkit-appearance` ➡️ `appearance` is ok (I think)
❌ `display: -webkit-box` ➡️ `display: box` is _not_ ok 

Any thoughts?

libs/designsystem/src/lib/components/modal/footer/modal-footer.component.scss
RasmusKjeldgaard: Good catch. 
Stylelint gave A LOT of errors because of the `no-descending-specificity` rule, but in the end I opted for not creating a too crazy diff, as I feared a lot of the actual changes would fall through the cracks in a large diff. So I disabled the rule instead, for now. 

So these changes are actually me who forgot to revert some of the initial reordering I did.

libs/designsystem/src/lib/polyfills/intersection-observer-polyfill-loader.js
```diff
@@ -1,14 +1,14 @@
...
+    script.src = 'kirby/polyfills/intersection-observer-polyfill.min.js';
+    document.head.appendChild(script);
+  }
 
-    function polyfillNeeded() {
-        return (typeof IntersectionObserver === 'undefined');
-    }
-  })();
+  function polyfillNeeded() {
+    return typeof IntersectionObserver === 'undefined';
```
jkaltoft, line 12: Are we completely sure we want to remove optional parentheses?

For: Sometimes they make things easier to read and understand
Against: Clutter. Anything else?

May I suggest that we postpone decisions on which stylelint rules to use in the config? We could spend some time getting used to the current rules and the have a meeting and decide later if we want to change anything.

apps/cookbook/src/app/showcase/showcase.component.scss
```diff
@@ -47,11 +47,12 @@ h4 {
   height: $icon-size;
   margin-right: utils.size('xxs');
   margin-left: utils.size('m');
-  color: currentColor;
-  fill: currentColor;
+  color: currentcolor;
+  fill: currentcolor;
```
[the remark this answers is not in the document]
RasmusKjeldgaard: Wait, is this really an error? https://developer.mozilla.org/en-US/docs/Web/CSS/color#syntax

jkaltoft (changes requested):
I think this was the last batch of comments - at least for now 🙂

jkaltoft:
May I suggest that we postpone decisions on which stylelint rules to use in the config? We could spend some time getting used to the current rules and the have a meeting and decide later if we want to change anything.

libs/core/src/scss/base/_line-clamp.scss
RasmusKjeldgaard: Nice. This is caused by the https://stylelint.io/user-guide/rules/list/value-no-vendor-prefix/ rule. 
Should we customise the rule to ignore specific properties, or disable it entirely?

RasmusKjeldgaard: As per your other comment, I can see that stylelint integrates with Autoprefixer by default, so maybe it should just be disabled for this specific line, if it is still needed?

libs/core/src/scss/base/_variables.scss
RasmusKjeldgaard: Yes!

libs/core/src/scss/print/generic/_colors.scss
```diff
@@ -1,9 +1,9 @@
 @media print {
   *,
-  ::after,
```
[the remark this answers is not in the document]
RasmusKjeldgaard: Done!

libs/designsystem/src/lib/components/avatar/avatar.component.scss
```diff
@@ -9,7 +9,7 @@ $diameter-large: map.get(utils.$avatar-sizes, 'l');
 $badge-diameter: utils.$avatar-badge-size;
 
 @function get-badge-position($avatar-diameter) {
-  $cos-to-45-degrees: 0.70710678118;
+  $cos-to-45-degrees: 0.7071;
```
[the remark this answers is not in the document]
jkaltoft: I think it's ok with a precision of 4 decimals. As long as we can decide the value for that setting 👍🏻

libs/core/src/scss/themes/_colors.scss
```diff
@@ -1,42 +1,39 @@
...
 );
-
 $text-colors: (
-  'white': #ffffff,
+  'white': #fff,
   'semi-dark': #707070,
   'black': #1c1c1c,
   'danger': #ee0d0d,
 );
+$focus-ring-color: rgb(77 144 254);
```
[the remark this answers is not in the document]
RasmusKjeldgaard: This configuration-point in stylelint is only related to the color function 👍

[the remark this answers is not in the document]
RasmusKjeldgaard: Fortunately, it only controls the color function https://stylelint.io/user-guide/rules/list/color-function-notation/

libs/designsystem/src/lib/components/modal/footer/modal-footer.component.scss
jkaltoft: Ah, ok. Fair reasoning. The diff is large enough as it is 🙂

That said, this is one of the rules I was looking forward to using. How about we do a second run in another PR? We can apply `no-descending-specificity`, and perhaps other rules we might want to change, if things doesn't get too crazy too review in one go.

libs/designsystem/src/lib/components/dropdown/dropdown.component.ts
RasmusKjeldgaard: This is a prettier-specific configuration related to line-breaks and/or line length, so I think we should leave it as is.

libs/designsystem/src/lib/components/fab-sheet/fab-sheet.component.ts
RasmusKjeldgaard: Yes, I thinks this is auto-fixed by the [ordered imports](https://palantir.github.io/tslint/rules/ordered-imports/) tslint rule, so we have some config options.

apps/cookbook/src/app/iphone/iphone.component.scss
jkaltoft: Ah, ok. `--text-color--dark` is being used in [one place in this file](https://github.com/kirbydesign/designsystem/blob/0e15ad9161a9e3d3c02222e25813f7545fad5b8a/apps/cookbook/src/app/iphone/iphone.component.scss#L78), so we should probably still disable the rule.

But we should also revisit the naming of our custom properties so we don't have to disable the rule here.

libs/designsystem/src/lib/components/fab-sheet/fab-sheet.component.ts
jkaltoft: Hmmm, if it's not a stylelint or prettier rule, how come the file hasn't been reformatted before 🤔? Perhaps no-one has modified the file ever 😁

apps/cookbook/src/app/showcase/showcase.component.scss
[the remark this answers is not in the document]
jkaltoft: Actually, I don't think it is. Good catch yourself.

It just looks weird when you're so used to a different casing. Image if you had to get used to iphOne 😁

> Color keywords are case-insensitive identifiers [...]
https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#color_keywords

The devtools in browsers suggest `currentColor` with a capital C, but they respect all lowercase color keywords.

libs/designsystem/src/lib/components/fab-sheet/fab-sheet.component.ts
RasmusKjeldgaard: Nice catch, the typescript docs uses parentheses, so I am unsure of the technical differences. https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions

I'll use // prettier-ignore on this specific line.

jkaltoft (changes requested):

libs/designsystem/src/lib/components/flag/flag.component.ts
RasmusKjeldgaard: Confirmed!

libs/designsystem/src/lib/components/form-field/_form-field-inputs.shared.scss
RasmusKjeldgaard: Unfortunately, I do not think stylelint is clever enough to know what property is meant in a given place, it can only identify the prefix.

But auto-prefixer should handle this, if the correct rule is used in the first place, so I'll reduce this to `::placeholder`.

libs/core/src/scss/base/_line-clamp.scss
RasmusKjeldgaard: Done!

libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.ts
RasmusKjeldgaard: "Typescript does not care about indentations" - Mads 2022

libs/designsystem/src/lib/polyfills/intersection-observer-polyfill-loader.js
RasmusKjeldgaard: This is a tslint setting, and the linter has been run when this file was staged applying prettier changes. So lets take this in another round when configuring linting.

Functionally it should make no difference.

jkaltoft (approved):
giphub

libs/designsystem/src/lib/components/modal/modal-wrapper/modal-wrapper.component.scss
```diff
@@ -83,16 +57,41 @@ ion-header ion-toolbar {
...
   }
 
-  @include phablet-toolbar-padding();
+  @include phablet-toolbar-padding;
+}
+
+:host {
+  --vh100: var(--vh, 1vh) * 100; /// Fixes an issue with vh units on iOS Safari
+  --header-height: 0;
+  --footer-height: 0;
```
RasmusKjeldgaard, line 66: Probably needs to be reverted, just to be sure.

PR #2169, opened 2022-04-19
OnPush change detection for simplest components

### LINKED ISSUE #2077: [Enhancement] Use OnPush Change Detection strategy in 'dumb' Kirby components

RasmusKjeldgaard:

This PR closes #2077 

## What is the new behavior?
The simplest of components that do not have mutable state, has been changed to use OnPush change detection by default. 

- [ ] Yes
- [x] No

While this is not a breaking change, it still targets a major release. Even though it has been tested thoroughly, there might be edge cases or unforeseen consequences related to this change, and therefore it should be announced in the release notes of v6, so consumers get a chance to verify that everything works as intended. 

## Checklist:

### Reminders

### Review  
- [ ] ~~Request that the changes are [UX reviewed](https://github.com/kirbydesign/designsystem/blob/stable/.github/CONTRIBUTING.md/#ux-review) (only necessary if your PR introduces visual changes)~~

### DIFF
--- a/libs/designsystem/src/lib/components/card/card.component.ts
+++ b/libs/designsystem/src/lib/components/card/card.component.ts
@@ -1,5 +1,4 @@
 import {
-  ChangeDetectionStrategy,
   Component,
   ElementRef,
   HostBinding,
@@ -16,7 +15,6 @@ import { ResizeObserverEntry } from '../shared/resize-observer/types/resize-obse
   selector: 'kirby-card',
   templateUrl: './card.component.html',
   styleUrls: ['./card.component.scss'],
-  changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class CardComponent implements OnInit, OnDestroy {
   @Input() title: string;
--- a/libs/designsystem/src/lib/components/card/card.component.ts
+++ b/libs/designsystem/src/lib/components/card/card.component.ts
@@ -1,4 +1,5 @@
 import {
+  ChangeDetectionStrategy,
   Component,
   ElementRef,
   HostBinding,
@@ -15,6 +16,7 @@ import { ResizeObserverEntry } from '../shared/resize-observer/types/resize-obse
   selector: 'kirby-card',
   templateUrl: './card.component.html',
   styleUrls: ['./card.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class CardComponent implements OnInit, OnDestroy {
   @Input() title: string;
--- a/libs/designsystem/src/lib/components/form-field/input-counter/input-counter.component.ts
+++ b/libs/designsystem/src/lib/components/form-field/input-counter/input-counter.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnDestroy, OnInit } from '@angular/core';
+import { ChangeDetectionStrategy, Component, Input, OnDestroy, OnInit } from '@angular/core';
 import { Subscription } from 'rxjs';
 
 import { InputComponent } from '../input/input.component';
@@ -8,6 +8,7 @@ import { TextareaComponent } from './../textarea/textarea.component';
 @Component({
   selector: 'kirby-input-counter',
   templateUrl: './input-counter.component.html',
+  changeDetection: ChangeDetectionStrategy.Default,
 })
 export class InputCounterComponent implements OnInit, OnDestroy {
   @Input() listenTo: InputComponent | TextareaComponent;
--- a/libs/designsystem/src/lib/components/form-field/input-counter/input-counter.component.ts
+++ b/libs/designsystem/src/lib/components/form-field/input-counter/input-counter.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component, Input, OnDestroy, OnInit } from '@angular/core';
+import { Component, Input, OnDestroy, OnInit } from '@angular/core';
 import { Subscription } from 'rxjs';
 
 import { InputComponent } from '../input/input.component';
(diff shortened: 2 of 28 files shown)


RasmusKjeldgaard:
Will be rebased when #2170 is merged to release/v6.0.0, but can easily be reviewed before then 👍

libs/designsystem/src/lib/components/card/card.component.ts
```diff
@@ -15,6 +16,7 @@ import { ResizeObserverEntry } from '../shared/resize-observer/types/resize-obse
   selector: 'kirby-card',
   templateUrl: './card.component.html',
   styleUrls: ['./card.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
MadsBuchmann, line 19: Depending on what constitutes a simple component in this case - is this one too complex due to the sizes input property? 
```
  @Input()
  set sizes(value: { [size: string]: number }) {
    if (typeof value === 'string') {
      console.error(
        'Sizes property cannot be a string. Please ensure the size property is bound as an expression:\n[sizes]="{...}"'
      );
    }
    this.sizesSortedByBreakpoint = this.sortSizesByBreakpoint(value);
  }
```

libs/designsystem/src/lib/components/form-field/input-counter/input-counter.component.ts
```diff
@@ -8,6 +8,7 @@ import { TextareaComponent } from './../textarea/textarea.component';
 @Component({
   selector: 'kirby-input-counter',
   templateUrl: './input-counter.component.html',
+  changeDetection: ChangeDetectionStrategy.Default,
```
MadsBuchmann, line 11: Why default?

libs/designsystem/src/lib/components/modal/alert/alert.component.spec.ts
```diff
@@ -1,150 +1,104 @@
-import { CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
-import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
-import { By } from '@angular/platform-browser';
 import { createHostFactory, SpectatorHost } from '@ngneat/spectator';
```
MadsBuchmann, line 1: Why are the changes to this test suite necessary?

libs/designsystem/src/lib/components/modal/alert/alert.component.ts
```diff
@@ -9,6 +16,7 @@ import { WindowRef } from '../../../types/window-ref';
   styleUrls: ['./alert.component.scss'],
   // tslint:disable-next-line: no-host-metadata-property
   host: { '[class.ion-page]': 'false' }, //Ensure ion-page class doesn't get applied by Ionic Modal Controller
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
MadsBuchmann, line 19: Do we know how onPush handles observables which the title input property can take?
```typescript
title$: Observable<string>;
  @Input()
  set title(title: string | Observable<string>) {
    this.title$ = typeof title === 'string' ? of(title) : title;
}
```

libs/designsystem/src/lib/components/toggle/toggle.component.spec.ts
```diff
@@ -1,63 +1,62 @@
-import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 import * as ionic from '@ionic/angular';
-import { MockComponent } from 'ng-mocks';
+import { createComponentFactory, Spectator } from '@ngneat/spectator';
```
MadsBuchmann, line 3: Why are the changes to this spec necessary?

MadsBuchmann (commented):
Good work! I got a couple of questions however. 
They mostly stem from my limited knowledge of on push 😄

libs/designsystem/src/lib/components/modal/alert/alert.component.spec.ts
RasmusKjeldgaard: I have written a short description for this in the commit-message:
"Directly setting the input on the component did not work with OnPush
change detection, so spectator is used in alert test now, while we
were at it. This way we can interact with the component via setInput()."

libs/designsystem/src/lib/components/toggle/toggle.component.spec.ts
RasmusKjeldgaard: I have written a short description for this in the commit-message:
"Directly setting the input on the component did not work with OnPush
change detection, so spectator is used in alert test now, while we
were at it. This way we can interact with the component via setInput()."

libs/designsystem/src/lib/components/form-field/input-counter/input-counter.component.ts
RasmusKjeldgaard: I think I'll remove this. It was done because I know for a fact that OnPush breaks this component, but this is a weird way of communicating that, I realise 🤔

libs/designsystem/src/lib/components/modal/alert/alert.component.ts
RasmusKjeldgaard: This should be safe. Observables are comparable to Immutable inputs in this regard. See more here: https://vsavkin.com/change-detection-in-angular-2-4f216b855d4c

libs/designsystem/src/lib/components/card/card.component.ts
RasmusKjeldgaard: Hmm, my initial hunch was that a setter would ensure change detection being run, but revisiting it now, I am not so sure. Consumers could provide a mutable object here and change it without us knowing, I guess..?

RasmusKjeldgaard: Reverted in b6d2f0ac2f1a74768a6bd8b0107c7de8da10a5b4

libs/designsystem/src/lib/components/form-field/input-counter/input-counter.component.ts
MadsBuchmann: Ah i see - then i actually think it makes sense to set the changeDetection explicitly if it is backed up by a comment (alternatively - just leave a comment).

MadsBuchmann (approved):

PR #2175, opened 2022-04-20
Remove deprecated okBtnText and cancelBtnText from AlertConfig

### LINKED ISSUE #2063: [Housekeeping] Remove deprecated okBtnText and cancelBtnText from AlertConfig
As per the deprecation notice in [AlertConfig](https://github.com/kirbydesign/designsystem/blob/main/libs/designsystem/src/lib/components/modal/alert/config/alert-config.ts#L7-L15) `okBtnText` and `cancelBtnText` can be removed. 

The target is currently Kirby v6.

RasmusKjeldgaard:
This PR closes #2063 

## What is the new behavior?
`okBtnText` and `cancelBtnText` properties of the AlertConfig interface has been removed. The AlertConfig interface is used by consumers to configure the `ModalController` with the alert flavor. 

Additionally, the input-props of the AlertComponent has been changed to reflect this. This makes it a breaking change to the alert component too, but this should have low to no impact on consumers, as the alert component should not be used on its own, but rather through the ModalController/AlertHelper configuration. 

- [x] Yes
- [ ] No

`okBtnText` and `cancelBtnText` of `AlertConfig` are removed, and the existing `okBtn` and `cancelBtn` should be used instead, as a direct replacement for these. 

If for some reason the AlertComponent is used directly in markup as `kirby-alert` instead of passing the config to the modalController, the input bindings of `kirby-alert` should be changed accordingly, to `okBtn` and `cancelBtn`. 

## Checklist:

### Reminders

### Review  
- [ ] ~~Request that the changes are [UX reviewed](https://github.com/kirbydesign/designsystem/blob/stable/.github/CONTRIBUTING.md/#ux-review) (only necessary if your PR introduces visual changes)~~

### DIFF
--- a/libs/designsystem/src/lib/components/modal/services/alert.helper.ts
+++ b/libs/designsystem/src/lib/components/modal/services/alert.helper.ts
@@ -30,7 +30,7 @@ export class AlertHelper {
   private getComponentProps(config: AlertConfig) {
     return {
       ...config,
-      okBtn: this.getOkBtnText(config),
+      okBtn: this.getOkBtn(config),
       cancelBtn: config.cancelBtn,
       okBtnIsDestructive: this.getOkBtnIsDestructive(config),
       iconName: config.icon && config.icon.name,
@@ -38,7 +38,7 @@ export class AlertHelper {
     };
   }
 
-  private getOkBtnText(config: AlertConfig) {
+  private getOkBtn(config: AlertConfig) {
     let text: string;
 
     if (config.okBtn) {
--- a/libs/designsystem/src/lib/components/modal/services/alert.helper.ts
+++ b/libs/designsystem/src/lib/components/modal/services/alert.helper.ts
@@ -31,7 +31,7 @@ export class AlertHelper {
     return {
       ...config,
       okBtn: this.getOkBtnText(config),
-      cancelBtn: this.getCancelBtnText(config),
+      cancelBtn: config.cancelBtn,
       okBtnIsDestructive: this.getOkBtnIsDestructive(config),
       iconName: config.icon && config.icon.name,
       iconThemeColor: config.icon && config.icon.themeColor,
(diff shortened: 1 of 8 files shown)


libs/designsystem/src/lib/components/modal/services/alert.helper.ts
```diff
@@ -60,11 +56,6 @@ export class AlertHelper {
   }
 
   private getCancelBtnText(config: AlertConfig) {
```
MadsBuchmann, line 58: Consider remvoing this line competely and just use `config.cancelBtn` on line 34 instead.

libs/designsystem/src/lib/components/modal/services/alert.helper.ts
```diff
@@ -39,12 +40,7 @@ export class AlertHelper {
 
   private getOkBtnText(config: AlertConfig) {
```
MadsBuchmann, line 41: Consider renaming to reflect new property name: 
```suggestion
  private getOkBtn(config: AlertConfig) {
```

MadsBuchmann (changes requested):
Burn the deprecated properties buuuurn! 
billede
 
Oh yeah and before i get too carried away - two minor comments :D

libs/designsystem/src/lib/components/modal/services/alert.helper.ts
RasmusKjeldgaard: I thought about this, but ended up not doing it because it actually returns text in the end (line 48-56). Thoughts on that?

libs/designsystem/src/lib/components/modal/services/alert.helper.ts
RasmusKjeldgaard: Fixed in a1a30352

libs/designsystem/src/lib/components/modal/services/alert.helper.ts
RasmusKjeldgaard: Renamed in d4fc2b3

MadsBuchmann (approved):

PR #2177, opened 2022-04-21
Remove deprecated dim property on ModalConfig

### LINKED ISSUE #2064: [Housekeeping] Remove deprecated dim on ModalConfig
As per the deprecation notice in [ModalConfig](https://github.com/kirbydesign/designsystem/blob/main/libs/designsystem/src/lib/components/modal/modal-wrapper/config/modal-config.ts#L22) the dim property can be removed. 

The current target is Kirby v6.

RasmusKjeldgaard:

This PR closes #2064

## What is the new behavior?
Dim is removed 🥳 

- [x] Yes
- [ ] No

Dim is no longer a configurable option, so it should be removed from ModalConfig.


MadsBuchmann:
Seems like something is up with the git history 🤔

RasmusKjeldgaard:
> Seems like something is up with the git history 🤔

I was too quick when creating the PR. It now has the correct base branch!

MadsBuchmann (approved):
Yes!

PR #2208, opened 2022-05-02
Fix slow hot reloading

### LINKED ISSUE #2203: [Housekeeping] Fix slow hot reloading on Angular 12
Hot reloading on the branch `release/v6.0.0` is slow. That has happened after the upgrade to Angular 12. 

We should fix that :o)

MadsBuchmann:

This PR closes #2203

Also i got a bit carried away and submitted #2209 which removes the e2e app. There's no reason for having it. 

## What is the new behavior?

A "production builds by default" initiative (see: https://github.com/angular/angular-cli/commit/656f8d75a3368a5affd1c55145841123dafdb007) has resulted in our development configurations being the production settings. 

I've added a 'development' configuration which can be used when developing locally; it uses the old defaults which are better for development.

Also i've tried to tidying it up a little. 

- [ ] Yes
- [X] No

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -95,6 +95,15 @@
             "scripts": []
           },
           "configurations": {
+            "development": {
+              "optimization": false,
+              "aot": false,
+              "buildOptimizer": false,
+              "sourceMap": true,
+              "extractLicenses": false,
+              "namedChunks": true,
+              "vendorChunk": true
+            },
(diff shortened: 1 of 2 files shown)


RasmusKjeldgaard (approved):

RasmusKjeldgaard:
So fast

PR #2209, opened 2022-05-02
Remove cookbook-e2e app & helm charts

MadsBuchmann:

Is an extension of #2208

## What is the new behavior?
For some reason we have a cookbook-e2e application. It is not really used for anything. Most likely it is autogenerated from when the project was set up. 

Also, as part of PR https://github.com/kirbydesign/designsystem/pull/1706 helm charts was added to the repository - before that they had been acquired via a dependency to another repository. However we do not use kubernetes for deploying anymore, therefore we can just remove them.

I think we should remove it to reduce clutter. There is no reason to have unused code and configurations in the repository. 

- [ ] Yes
- [X] No


angular.json
```diff
@@ -95,21 +95,23 @@
...
           "configurations": {
+            "development": {
+              "optimization": false,
+              "aot": false,
+              "buildOptimizer": false,
+              "sourceMap": true,
+              "extractLicenses": false,
+              "namedChunks": true,
+              "vendorChunk": true
+            },
```
MadsBuchmann, line 106: For some reason this is appearing in the diff - but it is already present on target branch

angular.json
```diff
@@ -95,21 +95,23 @@
...
                 }
               ],
-              "optimization": true,
               "outputHashing": "all",
-              "sourceMap": false,
-              "namedChunks": false,
-              "aot": true,
-              "extractLicenses": true,
-              "vendorChunk": false,
-              "buildOptimizer": true,
```
MadsBuchmann, line 112: For some reason this is appearing in the diff - but it is already present on target branch

angular.json
```diff
@@ -128,7 +130,7 @@
         "serve": {
           "builder": "@angular-builders/custom-webpack:dev-server",
           "options": {
-            "browserTarget": "cookbook:build"
```
MadsBuchmann, line 131: For some reason this is appearing in the diff - but it is already present on target branch

readme.md
```diff
@@ -209,8 +209,7 @@ A basic walkthrough is outlined in the structure below:
 ```
 @kirbydesign/designsystem
 ├── apps                    # Contains source code for applications
-|  ├── cookbook             # - Cookbook application (showcase and examples)
-|  └── cookbook-e2e         # - End-to-end tests for Cookbook application
+|  └── cookbook             # - Cookbook application (showcase and examples)
 ├── config
 |  └── helm
```
RasmusKjeldgaard, line 214: We can remove helm here too, right?

package.json
```diff
@@ -50,7 +50,7 @@
     "prettier:fix": "prettier --write \"./{libs,apps}/**/*.{ts,json,md,scss,html}\"",
     "e2e": "ng e2e",
     "dist": "npm run dist:cookbook",
-    "dist:cookbook": "ng build cookbook --prod --aot --progress false",
+    "dist:cookbook": "ng build cookbook --configuration production",
```
MadsBuchmann, line 53: For some reason this is appearing in the diff - but it is already present on target branch

angular.json
RasmusKjeldgaard: I think you would have to rebase this branch on release/v6.0.0 because the commits from the previous branch with build time improvements was squashed when it was merged into release/v6.0.0

RasmusKjeldgaard (changes requested):
Can the e2e-related scripts in our package.json (line 51 and 65) be removed too?
And possibly also the cypress dependency?

MadsBuchmann:
Good point with cypress. It has been removed together with the package.json scripts for e2e testing.

RasmusKjeldgaard (approved):
MAXIMUM APPROVE

PR #2217, opened 2022-05-04
Fix Reported Flaky Tests

### LINKED ISSUE #1836: [Housekeeping] Harden flakey test
We should consider if we can refactor either the dropdown component or the component test related to the isOpen functionality of the dropdown, as I am seeing sporadic test failures when running the test suite. The TestHelper.whenTrue function returns its Timeout Error for [this test.](https://github.com/kirbydesign/designsystem/blob/master/libs/designsystem/src/lib/components/dropdown/dropdown-popover.component.spec.ts#L311)

Could we either avoid using actual time (some fakeAsync magic or similar?) or otherwise consider to add more milliseconds to the timeout in [dropdown.popover.component.spec.ts](https://github.com/kirbydesign/designsystem/blob/master/libs/designsystem/src/lib/components/dropdown/dropdown-popover.component.spec.ts#L311 )

image

mictro:

This PR closes #1836 

## What is the new behavior?

Fix of reported flaky. tests in kirby.

- [ ] Yes
- [X] No

## Checklist:

### Reminders
- [X] Make sure you have implemented tests following the guidelines in: "[The good: Test](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Test)".
- [X] Make sure you have updated the cookbook with examples and showcases (for bug fixes, enhancements & new components).

### Review  
- [X] Do a [self-review](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Self-review).
- [X] Request that the changes are code-reviewed 

### DIFF
--- a/libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
+++ b/libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
@@ -121,6 +121,14 @@ kirby-card {
   }
 }
 
+:host(.with-popover) {
+  &.is-open {
+    kirby-card {
+      display: block;
+    }
+  }
+}
+
 :host(.is-open) {
   & > button {
     box-shadow: utils.get-elevation(8);
--- a/libs/designsystem/src/lib/components/dropdown/dropdown-popover.component.spec.ts
+++ b/libs/designsystem/src/lib/components/dropdown/dropdown-popover.component.spec.ts
@@ -261,17 +261,13 @@ describe('DropdownComponent (popover version)', () => {
     });
 
     describe('when configured with popout direction', () => {
-      /* TODO: This test has been flakey for some time. 
-         I've excluded it as it is currently worthless. 
-         For an example take a look at this test run: 
-         https://github.com/kirbydesign/designsystem/runs/5813726617
-     */
-      xit('should open card to the left when popout=left', fakeAsync(() => {
+      it('should open card to the left when popout=left', fakeAsync(() => {
         spectator.component.popout = HorizontalDirection.left;
         spectator.element.style.cssFloat = 'right';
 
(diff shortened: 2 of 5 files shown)


libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
```diff
@@ -121,6 +121,14 @@ kirby-card {
...
 }
 
+:host(.with-popover) {
+  &.is-open {
+    kirby-card {
+      display: block;
+    }
+  }
+}
+
```
MadsBuchmann, line 131: I think we should leave this as is. 

I've been playing around with refactoring `dropdown.component.scss` a bit and i think there's enough to be considered, that it should happen in a separate issue. 

For example: 
- why do we have the classes `is-opening` & `is-open` which we have to track the state of in the controller. Can we get rid of them and simplify it?
- Why do we control opacity? 
- We should find a way to invert the relationship to `kirby-calendar` such that the necessary styles are set in the calendar stylesheet instead of in this file. It gets very hard to see where rules come from when using `host-context` unless strictly necessary. 
- General cleanup; the rule on line 44 could might as well be part of the first :host selector.

I think it is a task in of itself to do this properly and refactor this file and beyond the scope of this issue. 

Perhaps leave a comment behind explaining why this rule exists :)

MadsBuchmann (approved):

libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
MadsBuchmann: Also can we find a way to invert the `:not(.with-popover)` selector? I personally think something that is NOT is harder to read/mentally grasp as opposed to just adding something on top.

PR #2276, opened 2022-05-17
Extend transition for interaction states

### LINKED ISSUE #2273: Extend transition for interaction states
- [x] Rewrite the transition mixin to _not_ accept parameters
- [x] Use `@content` block instead and use that to override `transition-property` when needed
- [x] Update those few places where we transition on specific properties (other than `all`)
- [x] Write a function that returns the transition property values in shorthand form - some Ionic components need that, e.g. ion-item has as a `--transition` custom property that controls the transition effect.

jkaltoft:

This PR closes #2273 

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  

### DIFF
--- a/libs/core/src/scss/base/_link.scss
+++ b/libs/core/src/scss/base/_link.scss
@@ -20,8 +20,9 @@ $_focus-ring-border-radius: utils.size('xxxxs');
 }
 
 a {
-  @include interaction-state.transition;
   @include interaction-state.apply-hover {
     text-decoration-color: transparent;
   }
+
+  transition: interaction-state.transition();
 }
--- a/libs/core/src/scss/interaction-state/_focus.scss
+++ b/libs/core/src/scss/interaction-state/_focus.scss
@@ -60,10 +60,9 @@
 }
 
 @mixin _focus-ring($shadow, $gap) {
-  @include interaction-state.transition;
-
   $stroke-width: utils.size('xxxxs');
 
+  transition: interaction-state.transition();
   box-shadow: #{$shadow}, 0 0 0 $gap #{utils.get-color('background-color')},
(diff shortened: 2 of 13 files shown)


RasmusKjeldgaard:
> I have replaced the `transition` mixin with a function. Reviewer, it would be great if you could check out the diff between the (currently) last two commits [da7a45c](https://github.com/kirbydesign/designsystem/commit/da7a45c8b2bd5a2e3830b4690571a8db8ca67d4a) and [6d100be](https://github.com/kirbydesign/designsystem/commit/6d100be9c2ba2bbb30a4cf78aa70823bcc5a96bf) where both mixin and function are present. I'm not completely certain what I prefer.

My take on this is that even though it is wrapped in some unnecessary code, that the mixin solution is better because it is similar to hover and active in use. But I can see how there are (dis)advantages to both, so not set in stone for me.

jkaltoft:
> > I have replaced the `transition` mixin with a function. Reviewer, it would be great if you could check out the diff between the (currently) last two commits [da7a45c](https://github.com/kirbydesign/designsystem/commit/da7a45c8b2bd5a2e3830b4690571a8db8ca67d4a) and [6d100be](https://github.com/kirbydesign/designsystem/commit/6d100be9c2ba2bbb30a4cf78aa70823bcc5a96bf) where both mixin and function are present. I'm not completely certain what I prefer.
> 
> My take on this is that even though it is wrapped in some unnecessary code, the mixin solution is better because it is similar to hover and active in use. But I can see how there are (dis)advantages to both, so not set in stone for me.

I also prefer the mixin solution, but we need to add the transition function to handle special use cases (i.e. Ionic). So we either end up with both a mixin and a function or just the function - the mixin will just use the function anyway.

RasmusKjeldgaard (approved):

PR #2304, opened 2022-06-01
Add SCSS files to exports in package.json

andershessellund:
I forbindelse med opgradering til Angular 13 og dermed et nyt pakkeformat skal eksporterede ting eksplicit med i package.json . Angular tilføjer automatisk javascript filer, men scss filer skal tilføjes manuelt.

### DIFF
--- a/libs/designsystem/package.json
+++ b/libs/designsystem/package.json
@@ -23,5 +23,9 @@
     "@angular/router": ">=9.0.0",
     "rxjs": ">=6.5.3",
     "zone.js": ">=0.10.2"
+  },
+  "exports": {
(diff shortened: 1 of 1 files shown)


MadsBuchmann:
The referenced files just forward the scss files found in the core package. Do the core files have to be referenced somehow in package.json - or is it unecessary as the core package is not dependent on angular?

andershessellund:
The core package does not have an "exports" field in package.json, and therefore everything is exported.

RasmusKjeldgaard (approved):
Thank you for doing this!
Added some migration notes, as this is a breaking change.

PR #2315, opened 2022-06-08
Use import instead of require for highcharts

andershessellund:
This fixes error "require is not defined" by using import instead of require.


RasmusKjeldgaard (approved):

MortenGregersen:
Screen Shot 2022-06-08 at 09 25 25

PR #2316, opened 2022-06-08
Add backButtonClicked event to page component

### LINKED ISSUE #2119: [Enhancement] Add "backButtonClick" event to page

## Describe the enhancement
In some cases we want to be able to override the behaviour of the back button in the `kirby-page` element. 

For example when using a micro frontend (MF). A MF is composed of several frontend applications each having their own local router. Here the back button needs to navigate to a page outside of the router using functions exposed by some MF framework. 

Our current `kirby-page` can not facilitate this at the moment. The closest we come is the `defaultBackHref` input property. But this only allows for routes within the local router. 

@NikolajSkousen and i had a meeting the 18th of March 2022 where we came up with the following solution: 
 
- A `backButtonClick` event should be added to the `kirby-page` element
- When a value is set for `backButtonClick` the default back behaviour of the back button should be disabled
    - This should be done by overwriting the `onClick` event of the `ion-back-button`. This is marked as internal by the ionic type declarations. But `onClick` is a standard html attribute and therefore we can assume it will not be removed. It is merely the attribute used by ionic to handle clicks and provide default behaviour; allowing us to overwrite it to disable that behaviour.  

### Other implementation notes
Providing a value for `backButtonClick` should not alter the default display logic. If one wish to override the default the `hideBackButton` input property should be used instead.

mictro:

This PR closes #2119

## What is the new behavior?

Add a back-button event-emitter outpput option to page component. When set the default back-button behaviour is removed. 

- [ ] Yes
- [X] No

This change allows applications to navigate past the routing restrictions present in micro-frontends.

## Checklist:

### Reminders
- [X] Make sure you have implemented tests following the guidelines in: "[The good: Test](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Test)".
- [X] Make sure you have updated the cookbook with examples and showcases (for bug fixes, enhancements & new components).

### Review  
- [X] Do a [self-review](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Self-review).
- [X] Request that the changes are code-reviewed 


MadsBuchmann (changes requested):
It seems like the default behaviour has broken somehow 🤔  

Here's how the back button behaves in our showcase without any `backButtonClicked` event handler registered on the `main` branch: 

https://user-images.githubusercontent.com/17269698/173577520-cdd70e25-2486-46c4-9d28-57280036148d.mp4

On this branch nothing happens however.

libs/designsystem/src/lib/components/page/page.component.ts
```diff
@@ -161,6 +161,7 @@ export class PageComponent
   @Output() enter = new EventEmitter<void>();
   @Output() leave = new EventEmitter<void>();
   @Output() refresh = new EventEmitter<PullToRefreshEvent>();
+  @Output() backButtonClicked = new EventEmitter<Event>();
```
MadsBuchmann, line 164: Looking at this again i think we should change the name to be in imperative form, considering other events are named like that as well :D 
```suggestion
  @Output() backButtonClick = new EventEmitter<Event>();
```

libs/designsystem/src/lib/components/page/page.component.ts
```diff
@@ -259,6 +262,16 @@ export class PageComponent
     this.windowRef.nativeWindow.addEventListener(selectedTabClickEvent, () => {
       this.content.scrollToTop(KirbyAnimation.Duration.LONG);
     });
+
+    // Intercept back-button click events, defaulting to the built-in click-handler.
+    if (this.backButtonClicked.length == 0) {
```
MadsBuchmann, line 267: Are there a specific reason for using the equality - instead of the identity operator? 

Personally i prefer the more strict identity operator as it is more predictable how it will behave 😄 
```suggestion
    if (this.backButtonClicked.length === 0) {
```

libs/designsystem/src/lib/components/page/page.component.ts
```diff
@@ -259,6 +262,16 @@ export class PageComponent
...
+      this.backButtonClicked
+        .pipe(takeUntil(this.ngOnDestroy$))
+        .subscribe(this.backButtonDelegate.onClick.bind(this.backButtonDelegate));
+    }
+    this.backButtonDelegate.onClick = (event: Event) => {
+      this.backButtonClicked.emit(event);
+    };
```
MadsBuchmann, line 274: We could consider separating all of this into a `InterceptBackButtonClicks` method to make it more clear that this code belongs together and what its responsibility is

mictro: Done

libs/designsystem/src/lib/components/page/page.component.spec.ts
```diff
@@ -141,6 +141,19 @@ describe('PageComponent', () => {
...
+      const subscriber = jasmine.createSpy();
+      spectator.output('backButtonClicked').subscribe(subscriber);
+
+      spectator.click(ionBackButton);
+      spectator.detectComponentChanges();
+
+      expect(subscriber).toHaveBeenCalledTimes(1);
+    });
+  });
+
```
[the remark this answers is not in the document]
mictro: Added

apps/cookbook/src/app/showcase/page-showcase/page-showcase.component.ts
```diff
@@ -86,6 +86,12 @@ export class PageShowcaseComponent {
       description: 'Emitted when leaving the page',
       signature: 'func',
     },
+    {
+      name: 'backButtonClicked',
```
MadsBuchmann, line 90:
```suggestion
      name: 'backButtonClick',
```

MadsBuchmann (changes requested):
One minor documentation thing other than that it looks goodie!

apps/cookbook/src/app/showcase/page-showcase/page-showcase.component.ts
mictro: Updated and regenerated mock

MadsBuchmann (approved):
LGTM!

PR #2390, opened 2022-07-04
Feature/459 affix functionality for kirby input

### LINKED ISSUE #459: [Enhancement] Ability to add prefix and suffix
It should be possible to add pre- and suffixes inside an input/form field to make it even more clear what the user is typing into that field. 

The solution should allow both icons and values, dependant on the usage-situation. The task should not be about making the icon also clickable. 

The design is either a black icon or a semi-dark text (see also screenshots)

We have considered the possibility that it should be possible to change the alignment of the text also, to make the suffix close to the the value you put inside it, but that might be too much configuration for the user using it...

See also: https://zpl.io/YYNKvmn

Pre   Suffix

See also https://getbootstrap.com/docs/4.0/components/input-group/ for inspiration

and use case should be here:

tspringborg:
## Why is this a draft?
I'd like to discuss some things I'm not completely happy with.

There are a couple of things.
1. In order to fix a compatibility issue when using the affix directive in conjunction with the date-mask directive, the affix directive simply removes the date-mask input field. Which doesn't seem to do anything atm. Is there a better way that doesn't require refactoring form field and related components? ..and is the extra date-mask input actually needed? 

2. The CSS variables had to be added in quite a few places. Is there a better approach? (While still using CSS variables)

This PR closes #459

## What is the new behavior?

Added affix directive for kirby input fields

- [x] Yes
- [ ] No

Sort of. Since there's a compatibility issue with data-mask the affix directive will remove the extra input the date-mask adds. 

## Checklist:

### Reminders

### Review  


tspringborg:
Erm I accidentally clicked ready for review. Not sure how to revert to draft status...

apps/cookbook/src/app/examples/form-field-example/examples/input/default.ts
```diff
@@ -5,7 +5,7 @@ import { InputSize } from '@kirbydesign/designsystem';
 const config = {
   selector: 'cookbook-form-field-input-example',
   template: `<kirby-form-field>
-  <input kirby-input [size]="size" placeholder="Default input with placeholder text" />
+  <input kirby-input [size]="size" placeholder="Default input with placeholder text." />
```
tspringborg, line 8: a mistake. Disregard

jkaltoft:
> Erm I accidentally clicked ready for review. Not sure how to revert to draft status...

See link just below Reviewers (top of right column on desktop sized display)

libs/core/src/scss/base/_variables.scss
```diff
@@ -188,3 +188,7 @@ $transition-easings: (
   ),
 );
 $loading-overlay-backdrop-opacity: 0.8;
+$form-field-input-font-family: var(--kirby-font-family);
+$form-field-input-line-height: 1.5;
+$form-field-input-padding: 1em;
+$form-field-label-height: 24px;
```
jkaltoft, line 194: These Sass variables should stay in `libs/designsystem/src/lib/components/form-field/_form-field-inputs.shared.scss`

libs/designsystem/src/lib/components/form-field/_form-field-inputs.shared.scss
```diff
@@ -1,9 +1,5 @@
 @use '~@kirbydesign/core/src/scss/utils';
-
-$form-field-input-font-family: var(--kirby-font-family);
-$form-field-input-line-height: 1.5;
-$form-field-input-padding: 1em;
-$form-field-label-height: 24px;
+@import '~@kirbydesign/core/src/scss/base/variables';
```
jkaltoft, line 2: We use Sass Modules, so `@import` should not be used.

See e.g. https://sass-lang.com/documentation/at-rules/use and https://css-tricks.com/introducing-sass-modules/

jkaltoft, line 2: `@kirbydesign/core/src/scss/base/variables` is already included in `@kirbydesign/core/src/scss/utils` so this line should be removed

libs/designsystem/src/lib/components/form-field/affix/input-affix.component.scss
```diff
@@ -0,0 +1,18 @@
+@import '~@kirbydesign/core/src/scss/base/variables';
```
jkaltoft, line 1: Import other Sass files with `@use` instead of `@import` (Sass modules)

```suggestion
@use '../form-field-inputs.shared' as shared;
```

(change the relative path if I got the wrong number of `..`s 🙂)

libs/designsystem/src/lib/components/form-field/affix/input-affix.component.scss
```diff
@@ -0,0 +1,18 @@
...
+
+:host {
+  --input-padding: #{$form-field-input-padding};
```
jkaltoft, line 4:
```suggestion
  --input-padding: #{shared.$form-field-input-padding};
```

libs/designsystem/src/lib/components/form-field/affix/input-affix.component.scss
```diff
@@ -0,0 +1,18 @@
...
+  position: absolute;
+  top: 50%;
+  transform: translateY(-50%);
+
+  &.default {
+    font-size: 16px;
+    line-height: 24px;
+    letter-spacing: 0;
+    color: #707070;
+  }
```
jkaltoft, line 17: Should use design tokens and utility functions - not hardcoded values. E.g. `font-size: utils.font-size('n');`. The same goes for colors.

jkaltoft (changes requested):
I've done a quick pre-review on the stylesheets (mostly about using Sass Modules).

Regarding...

> The CSS variables had to be added in quite a few places. Is there a better approach? (While still using CSS variables)

...I'm convinced there may be a better approach, but I'll need some time to dig a bit deeper to come up with an idea for a solution.

libs/core/src/scss/base/_variables.scss
tspringborg: If they stay there. Then how can I import them without all the styles getting applied as well?

jkaltoft: See https://github.com/kirbydesign/designsystem/pull/2390#discussion_r913591504

RasmusKjeldgaard:
Replaced by https://github.com/kirbydesign/designsystem/pull/2406

PR #2380, opened 2022-06-28
Upgrade to Ionic 6

RasmusKjeldgaard:
## Which issue does this PR close?
Fixes #1932, fixes #1009, fixes #2260

## What is the new behavior?
Upgraded to the newest Ionic version at the time of speaking, and made the necessary changes as per the [upgrade guide](https://ionicframework.com/docs/intro/upgrading-to-ionic-6).

This is mostly style refactorings to the modal to account for the new shadow root. 
Furthermore, there are some changes to the Modal Animation Builder, as the modal was no longer animating correctly. 
The animation builder is essentially just holding a copy of the [enter/leave animations from Ionic](https://github.com/ionic-team/ionic-framework/tree/main/core/src/components/modal/animations), with some adjustments to account for removing the opacity on stacked backdrops. 

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Are there any additional context?

With the new update we are seeing a boatload of these warnings in the console whenever a modal is opened: 

```
[Ionic Warning]: swipeToClose has been deprecated in favor of canDismiss.

If you want a card modal to be swipeable, set canDismiss to `true`. In the next major release of Ionic, swipeToClose will be removed, and all card modals will be swipeable by default.
```

I tried switching to canDismiss, but this does not seem to work for our drawer, as it is not a card modal. We will have to leave it to a followup issue to remove the deprecated functionality, before it is completely removed from Ionic. 

## Checklist:

The following tasks should be carried out in sequence in order to follow [the process of contributing](https://github.com/kirbydesign/designsystem/blob/main/.github/CONTRIBUTING.md/#the-process-of-contributing) correctly.

### Reminders
- [x] Make sure you have implemented tests following the guidelines in: "[The good: Test](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Test)".
- [x] Make sure you have updated the cookbook with examples and showcases (for bug fixes, enhancements & new components).

### Review  
- [x] Do a [self-review](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Self-review).
- [x] Request that the changes are code-reviewed 
- [ ] Request that the changes are [UX reviewed](https://github.com/kirbydesign/designsystem/blob/main/.github/CONTRIBUTING.md/#ux-review) (only necessary if your PR introduces visual changes)

When the pull request has been approved it will be merged to `develop` by Team Kirby.

Issue #2341, opened 2022-06-15
Deprecate the standalone chip component


RasmusKjeldgaard:
Chip should not be part of the public API. We should announce the deprecation in an early ~~v6~~ v7 feature release, and then remove it completely in ~~v7~~ next major version. 

Remember to remove it from the cookbook - and use a button in the Segmented Control showcase for the grouped example "show all" chip.

- [x] Add deprecation warning
- [x] Remove from cookbook
- [x] Update Segmented Control showcase
- [x] Create issue for removing standalone chip component (next major version)

PR #2417, opened 2022-08-12
Align behavior for all Segmented Control modes

### LINKED ISSUE #2347: Segmented control - Align keyboard interactions between the different modes
The arrow key functionality for the default Segmented Control has been introduced by the Ionic 6 upgrade. Chip mode should use the same method as default mode (arrow keys once it has received focus).

RasmusKjeldgaard:

This PR closes #2347

## What is the new behavior?
To implement similar keyboard interactions and behavior for all modes of the Segmented Control, it has been refactored to use `ion-segment` and `ion-segment-button` in all use cases.

This should also pave the way for removing the chip entirely, as it is no longer used internally, and was deprecated in #2341.  

🚨 Please note: The intented keyboard-interactions will only take effect when merged with the #2380. 

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


libs/designsystem/src/lib/components/segmented-control/segmented-control.component.spec.ts
```diff
@@ -70,61 +70,11 @@ describe('SegmentedControlComponent', () => {
...
-
-        const clonedItems = JSON.parse(JSON.stringify(items));
-        spectator.setHostInput({ items: clonedItems });
-        expect(onSegmentSelectSpy).not.toHaveBeenCalled();
-      });
-    });
-  });
-
-  const testModes = [SegmentedControlMode.chip, SegmentedControlMode.compactChip];
+  const testModes = [
```
RasmusKjeldgaard, line 73: This whole files diff makes the changes look more advanced than they actually are. 

What actually happens is that default mode is added to the testModes array, and then the whole `in 'default' mode` that looks deleted here is modified slightly to fit all modes. 

Then the other part of the test suite testing the chip mode is deleted.

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
```diff
@@ -21,32 +21,62 @@
...
   }
 
+  &.default-mode {
+    ion-segment {
+      --background: #{utils.get-color('white')};
+
+      display: inline-flex; // The segmented control itself is block-level, but we don't want the background to stretch
+      width: auto; // The segmented control itself is block-level, but we don't want the background to stretch
+      overflow: visible; // Ensures the badge is not cut off
+      contain: unset; // Ensures the badge is not cut off
```
jkaltoft, line 31: In general one of the only good use cases for using `unset` is when assigned to the `all` property. In all other cases I prefer to just assign the value that `unset` will result in. That would be either `initial` or `inherit` depending on the property.

In this case I would suggest assigning either `initial` or explicitly `none` (the initial value of `contain`)

```suggestion
      contain: initial; // Ensures the badge is not cut off
```

or 

```suggestion
      contain: none; // Ensures the badge is not cut off
```

https://developer.mozilla.org/en-US/docs/Web/CSS/unset
https://developer.mozilla.org/en-US/docs/Web/CSS/contain#formal_definition

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
```diff
@@ -21,32 +21,62 @@
     }
   }
 
+  &.default-mode {
+    ion-segment {
+      --background: #{utils.get-color('white')};
+
+      display: inline-flex; // The segmented control itself is block-level, but we don't want the background to stretch
+      width: auto; // The segmented control itself is block-level, but we don't want the background to stretch
```
jkaltoft, line 29: Just out of curiosity: Could the same be achieved by using `width: fit-content`?

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
```diff
@@ -21,32 +21,62 @@
...
+        scrollbar-width: none; /* Firefox */
+        &::-webkit-scrollbar {
+          display: none;
+        }
+      }
+
+      --background: none;
+
+      border-radius: 0;
+      padding: #{utils.size('xxxs')} #{utils.size('xxxs')};
```
jkaltoft, line 47: If you are declaring the same value for both axises anyway then:

```suggestion
      padding: #{utils.size('xxxs')};
```

jkaltoft (changes requested):
Made a quick review of the style changes. Will follow up on the rest later 🙂

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.html
```diff
@@ -1,15 +1,13 @@
...
   [value]="value?.id"
-  [scrollable]="disableChangeOnSwipe"
+  [scrollable]="disableChangeOnSwipe || mode !== 'default'"
   (ionChange)="onSegmentSelect($event.detail.value)"
   (click)="preventOutsideClick($event)"
 >
   <ng-container *ngFor="let item of items">
     <ion-segment-button [value]="item.id">{{ item.text }} </ion-segment-button>
-    <div class="badge-container">
+    <div class="badge-container" *ngIf="item.badge">
```
jkaltoft, line 9: 🙌🏻

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
jkaltoft: I just realised this may be code you just moved - not something you wrote, right?

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
jkaltoft, line 29:
```suggestion
      width: fit-content;
```

You should of course verify that my suggestion works as intended 😁

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
```diff
@@ -21,32 +21,62 @@
...
+    ion-segment {
+      --background: transparent;
+
+      gap: #{utils.size('xxxs')};
+    }
+
+    ion-segment-button {
+      --background: transparent;
+      --padding-start: #{utils.size('xs')};
+      --padding-end: #{utils.size('xs')};
```
jkaltoft, line 74: If you're having issues with compact mode being too compact, you may be able to hit approximately the same amount of padding as before by changing these two custom properties. I think the closest you get is by using `15px` but since that is not one of our size tokens `size('s')` is probably an ok size.

If you choose `size('s')` you could even just remove both declarations, since it's the same as declared in [lines 19-20](https://github.com/kirbydesign/designsystem/blob/7e7c9e4da277169cd466b267501d6be102700f84/libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss#L19-L20). This of course requires that  `.sm` is used with compact mode.

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
jkaltoft: > Just out of curiosity: Could the same be achieved by using `width: fit-content`?

https://github.com/kirbydesign/designsystem/pull/2417#discussion_r949005332

RasmusKjeldgaard: It indeed does, much more readable!

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
RasmusKjeldgaard: Yeah, the reason for the compactness right now is actually that I wanted to propose a slightly _compacter_ mode, as there was very little difference before. But right now, I think it is just too compact.

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
RasmusKjeldgaard: Great, didn't catch that when fiddling with the padding!

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.scss
RasmusKjeldgaard: You are right, just moved it, but lets change it anyways now that we have a good description of the context here! I have set it to `none`.

jkaltoft (approved):
Great work!

Let me know if you need a second review, e.g., if you make additional changes to the size of compact mode.

RasmusKjeldgaard:
> Great work!
> 
> Let me know if you need a second review, e.g., if you make additional changes to the size of compact mode.

Only a very minor detail change, so merging this 🎉

PR #2480, opened 2022-09-16
Update changelog with v7 entry

RasmusKjeldgaard:
Changelog for v7.


jkaltoft (approved):
We could consider moving the _Features_ section below _Breaking Changes_ and _Deprecations_. At least I think we should try to avoid having _**card** - make it possible to use alternative hover and active on dark cards_ mentioned as the first item

PR #2501, opened 2022-09-21
Feature/local navigation

### LINKED ISSUE #2186: [COMPONENT] Local navigation bar

This issue is for Pre tech-refinement of a new component that we should evaluate.

I will call in for a short presentation...

Skærmbillede 2022-04-26 kl  14 36 19
Skærmbillede 2022-04-26 kl  14 37 07
Skærmbillede 2022-04-26 kl  14 39 02
Skærmbillede 2022-04-26 kl  14 39 24

https://material.io/components/tabs#usage

<hr />

## Checklist:

### Refinement

- [ ] Request that the issue is [UX refined](https://github.com/kirbydesign/designsystem/blob/stable/.github/CONTRIBUTING.md/#ux-refinement); do not proceed until this is done.
- [ ] Request that the issue is [tech refined](https://github.com/kirbydesign/designsystem/blob/stable/.github/CONTRIBUTING.md/#tech-refinement); do not proceed until this is done.

### Implementation 

- [ ] Make sure you have read: "[Before you get coding](https://github.com/kirbydesign/designsystem/blob/stable/.github/CONTRIBUTING.md/#before-you-get-coding)".
- [ ] Create a branch from the [stable branch](https://github.com/kirbydesign/designsystem/tree/stable) following our [branch naming convention](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Branch). 

### Review

RasmusTraeholt:

This PR closes #2186 

## What is the new behavior?

Added two new experimental features.

1. A new kirby-page directive `[kirbyPageStickyContent]` which allows custom content to be inserted between the kirby-page-title and the kirby-page-content. That will become fixed beneath the ion-header, when scrolling to its intersection point.
2. A new local-navigation component `<kirby-page-local-navigation>` that can be used in conjunction with the new kirby-page directive.

Other:
- A new `KirbyExperimentalModule` in the designsystem has been introduced in which experimental components can be declared and exposed.
- A new `ExperimentalExamplesModule` along with a new route `/examples/experimental/[path-to-experiment]` has been introduced in the cookbook app in which experimental features can be created and viewed (the new route does not have a ui link)

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


libs/designsystem/src/lib/components/index.ts
```diff
@@ -63,3 +63,5 @@ export { ToggleComponent } from './toggle/toggle.component';
 
 export { ProgressCircleComponent } from './progress-circle/progress-circle.component';
 export * from './toggle-button';
+
+export * from './page-local-navigation';
```
mictro, line 67: Always prefer explicit export, because implicit export is one of the most prominent factors for circular dependency error

libs/designsystem/src/lib/components/page-local-navigation/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './page-local-navigation.component';
+export * from './page-local-navigation-item';
```
mictro, line 2: Always prefer explicit export, because implicit export is one of the most prominent factors for circular dependency error

libs/designsystem/src/lib/components/page-local-navigation/page-local-navigation.component.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (value > -1 && value !== this._selectedIndex) {
+      this._selectedIndex = value;
+      this.selectedIndex$.next(value);
+      this.scrollToSelectedTab(value);
+    }
+  }
+  get selectedIndex(): number {
+    return this._selectedIndex;
+  }
+
```
mictro, line 36: Make it a two-way binding  
@Input  set selectedIndex(value: number)
@Output() selectedIndexChange = new EventEmitter)<number>();
Client two-way-binds with '[(selectedIndex)]=...' avoiding the (ugly) implicit requirement to listen to and update the index through itemSelect (done elsewhere but not nice).

libs/designsystem/src/lib/components/page-local-navigation/page-local-navigation-item.ts
```diff
@@ -0,0 +1,26 @@
...
+export interface LocalNavigationItem {
+  /**
+   * Text of the item
+   */
+  text: string;
+
+  /**
+   * Id of the item. Must be unique!
+   */
+  id: string;
```
mictro, line 20: This component doesn't seem to require a unique id - remove and let the client decide it if it needs one.

libs/designsystem/src/lib/components/page-local-navigation/page-local-navigation-item.ts
```diff
@@ -0,0 +1,26 @@
+import { ThemeColor } from '@kirbydesign/core';
+
+export type LocalNavigationItemBadge = {
+  icon?: string;
+  content?: string;
+  description?: string;
+  isCustomIcon?: boolean;
+  themeColor: ThemeColor;
+};
```
mictro, line 9: Loose/fluffy typing - consider something along the lines:
type BadgeIcon = {
  name: string;
  isCustom: boolean;
};
type BadgeText = {
  text: string;
};

type TabBadge = {
  content: BadgeIcon | BadgeText;
  description: string;
  color?: ThemeColor;
};

export interface NavigationTab {
  text: string;
  badge?: TabBadge;
}

libs/designsystem/src/lib/components/page-local-navigation/page-local-navigation.component.ts
mictro, line 36: Decide to use _selectedIndex or selectedIndex$ not both. 
If choosing the latter you may subscribe to it in order to update the selected tab - simplifying the busy setter method.

libs/designsystem/src/lib/components/page/page.component.ts
```diff
@@ -256,6 +276,15 @@ export class PageComponent
 
   ngOnInit(): void {
     this.removeWrapper();
+    this.refresherActive$
+      .pipe(
+        takeUntil(this.ngOnDestroy$),
+        filter((value) => value === true || value === false),
+        distinctUntilChanged()
+      )
```
mictro, line 284: takeUntil should be the last - see https://ncjamieson.com/avoiding-takeuntil-leaks/

mictro (changes requested):
Seen from my review corner - nice work; a good test for the component is missing.

RasmusTraeholt:
Hey @mictro thank you for the great feedback. But because i just realized that position: sticky works flawlessly when used in root of ion-content (i did not work in previous iterations because i was working from drb) i have done some major refactoring, since it removes the need for intersection and mutation observers. The code should be muck simpler and cleaner now, with fewer additions.

mictro:
> 

What's there not to like :=)

mictro:
LGTM; it remains for @jkaltoft to review the styling and let @henrikvoetmand take a final look at the UX.

jkaltoft (changes requested):
Thank you for your contribution, Rasmus.

I've reviewed mainly the stylesheets and added some comments. Most of them are suggestions for improvements - a few things need to be changed.

In general:

- Add a link from the PR to the issue it solves
- When using a solution based more or less entirely on a blog post or similar add a reference to the source(s). They can provide context and explain the solutions and the problems they solve in more detail than what fits in a comment 🙂
- Use generic and english terms in the examples (and tests) to align with the rest of the cookbook

I have not addressed accessibility issues in this review (except a single comment about using `tabindex` 🙂).

jkaltoft:
Oh, it seems that several comments have been marked as outdated due to updates being made before I got to submit the review. The comments still apply. I'll see if I can find the time to go through them again tomorrow.

RasmusTraeholt:
@jkaltoft Thank you for your feedback. I have made changes where it was needed. I don't know what to do with fluffy comments like 'it's a complicated solution for something seemingly simple'. You are welcome to make suggestions if you have a different solution in mind.

libs/designsystem/src/lib/components/page-local-navigation/page-local-navigation.component.scss
```diff
@@ -0,0 +1,117 @@
...
+    content: '';
+    background-color: $border-color-standard;
+    position: absolute;
+    height: $bottom-border-height;
+    bottom: 0;
+    z-index: 1;
+    left: 0;
+    width: 100%;
+
+    @media only screen and (max-width: $divider-max-width-breakpoint) {
```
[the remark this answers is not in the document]
RasmusTraeholt: the function does not support calculated values

libs/designsystem/src/lib/components/page-local-navigation/page-local-navigation.component.scss
```diff
@@ -0,0 +1,117 @@
...
+      left: -#{utils.size('s')};
+      width: calc(100% + (#{utils.size('s')} * 2));
+    }
+  }
```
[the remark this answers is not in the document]
RasmusTraeholt: It's the solution i came up with. You are welcome to make suggestions if you have a different one in mind.

RasmusTraeholt:
>     * When using a solution based more or less entirely on a blog post or similar add a reference to the source(s). They can provide context and explain the solutions and the problems they solve in more detail than what fits in a comment 🙂

What solutions are you referring too?

libs/designsystem/src/lib/components/page-local-navigation/page-local-navigation.component.scss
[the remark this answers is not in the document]
jkaltoft: My suggestion probably requires too many changes at this stage - considering it's an experimental feature - and it also challenges the design decisions. I think this is fine for now 🙂

libs/designsystem/src/lib/components/page-local-navigation/page-local-navigation.component.scss
```diff
@@ -0,0 +1,111 @@
+@use '~@kirbydesign/core/src/scss/utils';
+
+$tab-item-text-max-width: 100px;
+$bottom-border-height: 1px;
+$border-color-standard: utils.get-color('medium');
+$border-color-selected: utils.get-color('dark');
+$divider-max-width-breakpoint: calc(#{utils.$page-content-max-width} + (#{utils.size('s')} * 2));
```
jkaltoft, line 7:
```suggestion
$divider-max-width-breakpoint: utils.$page-content-max-width + utils.size('s') * 2;
```

jkaltoft:
> > ```
> > * When using a solution based more or less entirely on a blog post or similar add a reference to the source(s). They can provide context and explain the solutions and the problems they solve in more detail than what fits in a comment 🙂
> > ```
> 
> What solutions are you referring too?

For example, https://css-tricks.com/bold-on-hover-without-the-layout-shift/ has a good description of the same technique that you're using to avoid layout shifts when using bold text as hover effect.

jkaltoft:
> @jkaltoft Thank you for your feedback. I have made changes where it was needed. I don't know what to do with fluffy comments like 'it's a complicated solution for something seemingly simple'. You are welcome to make suggestions if you have a different solution in mind.

Nice work, Rasmus 👍🏻

Sorry for the fluffyness. I began writing the comment before you changed the `/* HAX */` bit, but submitted it after 🙂. It still makes sense to consider, though.

I don't think you need a different solution for this. The solution is understandable - it's just more complex than using a simple border.

I'm guessing this solution was chosen because we need the border to exceed the width of `.container` on narrow viewports, but I'm not completely sure. By looking at the code itself it's not immediately obvious why it can't be done with just a 1-line `border-bottom` declaration on `.container`.

So, my suggestion is that you consider writing a short comment explaining why we need a pseudo element and several lines of code to create a border.

jkaltoft (dismissed):

jkaltoft:
I'm on vacation next week so I've approved the changes in advance 🙂

If you have any questions or comments we can follow up on it the week after next week.

RasmusTraeholt:
> I'm guessing this solution was chosen because we need the border to exceed the width of `.container` on narrow viewports, but I'm not completely sure. By looking at the code itself it's not immediately obvious why it can't be done with just a 1-line `border-bottom` declaration on `.container`.

Theres a bit more too it than just breaking constraints. The selected bottom border itself needs to have rounded corners (on all four), which is also part of the reason that it is not "just" a bottom border.

RasmusTraeholt:
> > > ```
> > > * When using a solution based more or less entirely on a blog post or similar add a reference to the source(s). They can provide context and explain the solutions and the problems they solve in more detail than what fits in a comment 🙂
> > > ```
> > 
> > 
> > What solutions are you referring too?
> 
> For example, https://css-tricks.com/bold-on-hover-without-the-layout-shift/ has a good description of the same technique that you're using to avoid layout shifts when using bold text as hover effect.

Nice find. I won't reference something i did not see or base my solution on though. But i can extend the explanation a bit :)
I believe i saw a comment on stackoverflow with a suggestion to insert a pseudo-element to avoid layout shifts. But the question it was posted to had very little in common with what i wanted to achieve. So i wont reference that either, since it could cause more confusion than relevant context.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "designsystem",
-  "version": "7.1.1",
+  "version": "7.1.2",
```
jkaltoft, line 3: Bumping Kirby version should only be done when we publish, not as part of implementing a feature.

```suggestion
  "version": "7.1.1",
```

jkaltoft (changes requested):
Good work, Rasmus 👍🏻

The only thing left is to undo bumping the version number.

henrikvoetmand:
UX/UI review is OK!
There might be adjustments when we align with Toolbar component.

jkaltoft (approved):

mictro (approved):

mictro:
It looks good to all of us....
download

PR #2532, opened 2022-10-06
POC: Change fullscreen modal to card modal

mark-drastrup:

This PR closes no issue

## What is the new behavior?

The default modal is now a card modal instead of a fullscreen modal.

- [ ] Yes
- [ ] No

## Checklist:

### Reminders

### Review  

### DIFF
--- a/libs/designsystem/src/lib/components/modal/services/modal.helper.ts
+++ b/libs/designsystem/src/lib/components/modal/services/modal.helper.ts
@@ -7,6 +7,7 @@ import { ModalCompactWrapperComponent } from '../modal-wrapper/compact/modal-com
 import { ModalConfig, ModalFlavor, ModalSize } from '../modal-wrapper/config/modal-config';
 import { ModalWrapperComponent } from '../modal-wrapper/modal-wrapper.component';
 
+import { PlatformService } from '../../../helpers/platform.service';
 import { ModalAnimationBuilderService } from './modal-animation-builder.service';
 import { Overlay } from './modal.interfaces';
 
@@ -19,7 +20,8 @@ export class ModalHelper {
   constructor(
     private ionicModalController: ModalController,
     private modalAnimationBuilder: ModalAnimationBuilderService,
-    private windowRef: WindowRef
+    private windowRef: WindowRef,
+    private platform: PlatformService
   ) {}
 
   public async showModalWindow(config: ModalConfig): Promise<Overlay> {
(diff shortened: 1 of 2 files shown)


libs/designsystem/src/lib/components/modal/services/modal.helper.ts
```diff
@@ -52,8 +52,8 @@ export class ModalHelper {
     const ionModal = await this.ionicModalController.create({
       component: config.flavor === 'compact' ? ModalCompactWrapperComponent : ModalWrapperComponent,
       cssClass: [
-        'kirby-overlay',
-        'kirby-modal',
+        config.flavor !== 'modal' ? 'kirby-overlay' : null,
```
RasmusKjeldgaard, line 55: The modal seems to break now on desktop size screens, where it will not resize to fit at the bottom of the screen anymore. 
This is fine when testing on mobile, but we probably need to preserve the classes, and just only add custom behavior only for mobile screens.

libs/designsystem/src/lib/components/modal/services/modal.helper.ts
```diff
@@ -52,8 +52,8 @@ export class ModalHelper {
...
+        config.flavor !== 'modal' ? 'kirby-modal' : null,
```
RasmusKjeldgaard, line 56: @henrikvoetmand another thing to think about: When using the card modal, it will also remain swipeable on desktop, and it will be difficult for us to change that behavior. 

We should think of a good way to handle how to open different types of modals based on the platform. And consider if that is our job or an option exposed by the modal.

libs/designsystem/src/lib/components/modal/services/modal.helper.ts
RasmusKjeldgaard: Can we add back the styles for the modals background color, and other stuff that is not related to sizing?

mark-drastrup: I have made some logic to conditionally add the styles, depending on if the modal is displayed on desktop og mobile. Now the only affected modal should be the default modal on mobile. 
The solution could be prettier, but I think it's good enough for POC purposes. When we know whether we should use the fullscreen modal or the card modal, then I will make a better solution :)

RasmusKjeldgaard:
@mark-drastrup we can close this one now, right?

mark-drastrup:
@RasmusKjeldgaard yeah we don't need this POC anymore :)

PR #2550, opened 2022-10-18
Refactor dropdown styles

### LINKED ISSUE #2230: [Housekeeping] Refactor Dropdown Styles
The styles file for the dropdown component has become a mess after introducing popover. 

We should refactor it.

See this comment for suggestions: [link to comment](https://github.com/kirbydesign/designsystem/pull/2217#discussion_r865802488).

Keep as is and leave unreadable code for future generations of Kirby maintainers. Think of the children!

mark-drastrup:

This PR closes #2230

## What is the new behavior?
This PR doesn't introduce new behaviour. Its purpose is to make the dropdown styles more simple and therefore easier to maintain in the future.

- [ ] Yes
- [x] No

The issue refers to this [comment](https://github.com/kirbydesign/designsystem/pull/2217#discussion_r865802488), which has some suggestions on things that could be improved. There is a couple of things I wasn't able to do:

1. Removing the `is-opening` class. Without it the logic to change the horizontal or vertical direction of the dropdown breaks. We need to refactor the logic for the dropdown directions, if we want to remove this class. That is out of scope for this issue and I think that the time is better spent on other issues. 
2.  Move `kirby-calendar` specific styles to the calendar's stylesheet. The dropdown is used for the `year navigator` in the calendar, which appears in the top right of the calendar when `Show year navigator` is checked. As far as I understand, then the issue is that from the stylesheet of the calendar, it is not possible to write styles that reaches into the child component (in this case the dropdown), without using the deprecated `::ng-deep` selector.

Notes:
The comment also asks why we are controlling opacity. If the opacity is not controlled like we do, then the user will see a noticeable flicker, when the dropdown opens and has to change direction. E.g. the user will see the dropdown open downwards and then open upwards, after it calculates that it is outside the viewport, when opening downwards.

## Checklist:

### Reminders

### Review  

### DIFF
--- a/libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
+++ b/libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
@@ -37,7 +37,7 @@ $min-screen-width: 375px;
     }
   }
 
-  & button {
+  button {
     position: relative;
     margin: 0;
     outline: none;
@@ -77,7 +77,7 @@ $min-screen-width: 375px;
       opacity: 1;
     }
 
-    & > button {
+    > button {
       box-shadow: utils.get-elevation(8);
     }
   }
@@ -110,8 +110,7 @@ kirby-popover {
 
 kirby-card {
   max-height: $dropdown-max-height;
-  margin-top: utils.size('xxxs');
-  margin-bottom: utils.size('xxxs');
+  margin-block: utils.size('xxxs');
   overflow-y: auto;
   box-shadow: utils.get-elevation(8);
   min-width: $min-screen-width-small - $margin-horizontal-total;
--- a/libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
+++ b/libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
@@ -37,7 +37,7 @@ $min-screen-width: 375px;
     }
   }
 
-  button {
+  > button {
     position: relative;
     margin: 0;
     outline: none;
--- a/libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
+++ b/libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
@@ -57,7 +57,6 @@ $min-screen-width: 375px;
 
   kirby-card {
     display: none;
-    opacity: 0;
     position: absolute;
     z-index: utils.z('dropdown');
   }
@@ -71,7 +70,6 @@ $min-screen-width: 375px;
   &.is-open {
     kirby-card {
       display: block;
-      opacity: 1;
     }
 
     & > button {
@@ -118,12 +116,6 @@ kirby-card {
 
   max-width: calc(100vw - #{$margin-horizontal-total});
 }
-
-// :host(.is-open) {
-//   & > button {
-//     box-shadow: utils.get-elevation(8);
-//   }
-// }
 
 :host-context(kirby-calendar) {
   > button {
--- a/libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
+++ b/libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
@@ -57,6 +57,10 @@ $min-screen-width: 375px;
 
   kirby-card {
     display: none;
+
+    // Opacity is used to prevent visible flicker,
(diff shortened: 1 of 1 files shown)


jkaltoft (changes requested):
Good work.

I only have requests for minor changes (see comments).

In general it would be nice if we could move rules outside of `:host`. I think this will lower the specificity (which is good) but may be overwritten by more specific rules (which in this case is not good) 🙂. However, it may not be a feasible solution, so only change anything if it's a simple "move code from here to here" 🙂

It would also be great if the rules for `kirby-card` in different contexts/states could be moved closer together - especially the declarations that get overridden.

Unfortunately, I don't think something like this simplified example would work, but hopefully you get the idea:

```css
 kirby-card {
    display: none;
    opacity: 0;

    .is-opening & {
        display: block;
    }

    .is-open & {
        display: block;
        opacity: 1;
    }
}
```

Don't spend too much effort trying to change it 🙂

mark-drastrup:
> In general it would be nice if we could move rules outside of `:host`. I think this will lower the specificity (which is good) but may be overwritten by more specific rules (which in this case is not good) 🙂. However, it may not be a feasible solution, so only change anything if it's a simple "move code from here to here" 🙂
> 
> It would also be great if the rules for `kirby-card` in different contexts/states could be moved closer together - especially the declarations that get overridden.
> 
> Unfortunately, I don't think something like this simplified example would work, but hopefully you get the idea:
> 
> ```css
>  kirby-card {
>     display: none;
>     opacity: 0;
> 
>     .is-opening & {
>         display: block;
>     }
> 
>     .is-open & {
>         display: block;
>         opacity: 1;
>     }
> }
> ```
> 
> Don't spend too much effort trying to change it 🙂

@jkaltoft I totally agree with both points, but unfortunately it wasn't as simple "move code from here to here" for the first point 😅 And for the second point I wasn't able to find a solution to bring them closer together, than they already are. There must be a way to do it, but I wasn't able to do it in this round. I would be happy to try again or maybe work together with you on a call. I'm sure you have a better grasp on how it can be done 😄

jkaltoft:
> > In general it would be nice if we could move rules outside of `:host`. I think this will lower the specificity (which is good) but may be overwritten by more specific rules (which in this case is not good) 🙂. However, it may not be a feasible solution, so only change anything if it's a simple "move code from here to here" 🙂
> > It would also be great if the rules for `kirby-card` in different contexts/states could be moved closer together - especially the declarations that get overridden.
> > Unfortunately, I don't think something like this simplified example would work, but hopefully you get the idea:
> > ```css
> >  kirby-card {
> >     display: none;
> >     opacity: 0;
> > 
> >     .is-opening & {
> >         display: block;
> >     }
> > 
> >     .is-open & {
> >         display: block;
> >         opacity: 1;
> >     }
> > }
> > ```
> >   
> > Don't spend too much effort trying to change it 🙂
> 
> @jkaltoft I totally agree with both points, but unfortunately it wasn't as simple "move code from here to here" for the first point 😅 And for the second point I wasn't able to find a solution to bring them closer together, than they already are. There must be a way to do it, but I wasn't able to do it in this round. I would be happy to try again or maybe work together with you on a call. I'm sure you have a better grasp on how it can be done 😄

Actually, I haven't got any better ideas than the example above 🙂

libs/designsystem/src/lib/components/dropdown/dropdown.component.scss
```diff
@@ -36,47 +36,30 @@ $min-screen-width: 375px;
...
-  &.attention-level2 {
-    border: 1px solid transparent;
-  }
 
-  .text {
-    overflow: hidden;
-    text-overflow: ellipsis;
-  }
-}
+  button {
```
jkaltoft, line 40: Strictly speaking, this selector should be `> button `, shouldn't it? In the "before refactor" it uses a [child combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator) ➡️ "buttons that are _direct_ children of :host"

jkaltoft (changes requested):
I only have a single request for change left 🙂

jkaltoft (dismissed):
giphub

jkaltoft (approved):
giphub

PR #2554, opened 2022-10-19
Add eslint config to enable autoformatting of HTML files on save

mark-drastrup:

This PR closes no issue.

## What is the new behavior?
HTML files are automatically formatted on save.

- [ ] Yes
- [x] No


RasmusKjeldgaard:
Can we consider adding this somewhere in a readme instead? I know we have the extension recommendations in code, but I feel managing vscode specific settings in git is not the best pattern. If I want my editor to behave differently in this project, `settings.json` would suddenly be checked in and recognize my changes, and I would have to work my way around that right?

mark-drastrup:
Yeah that would be no problem :) If we do it this way, then we have 2 settings.json:
1. Workspace settings - The file in the PR.
2. User settings - The file that is updated if you go into preferences and change something. This file won't be checked into git.

The two files can live together side by side, but Workspace Settings file will have precedence if the same setting is configured in both files :)

RasmusKjeldgaard:
> @RasmusKjeldgaard yeah that would be no problem :) If we do it this way, then we have 2 settings.json:
> 
> 1. Workspace settings - The file in this PR.
> 2. User settings - The file that is updated if you go into preferences and change something. This file won't be checked into git.
> 
> The two files can live together side by side, but the Workspace Settings file will have precedence if the same setting is configured in both files :) https://code.visualstudio.com/docs/getstarted/settings#_settings-precedence

Good point, no harm done in including it then.

RasmusKjeldgaard (approved):

PR #2598, opened 2022-11-11
Prevent instantiation of multiple modals, when another modal is being instantiated but not completed

### LINKED ISSUE #1733: [BUG] Modal overlay is activated too late

## Describe the bug
When opening a modal using modalController.showModal(), an overlay is rendered to prevent the underlying UI from receiving further mouse clicks. However, in some cases this appears too late. This may, if the user double clicks, lead to more than one click registering, potentially causing multiple dialogs to be opened or other kinds of unexpected behaviour. This has been an issue for a while and in some places the DRB code has implemented its own workarounds, none of which are pretty.

 

The behaviour can be reproduced by opening the Kirby cookbook, going to https://cookbook.kirby.design/home/showcase/modal, and double clicking the "Show Modal" button. This will result in two modals being opened on top of each other. It will only work the first time, and it will only work if you click quickly enough.

Note: this *may* not work on faster machines. On my Macbook Pro, there is a ~140 mSec delay during which the button is still active, and if I double click quickly, I can manage to hit the button twice within ~110 mSec.

3.0.8

I would expect the modal overlay to appear before I could put in my second click, blocking mouse clicks.

- OS: MacOS
- Browser: Chrome
- Browser version: 91.0.4472.114 

I've investigated a bit without yet having fully understood it. At the core of the problem, it seems like the ion-modal needs to have the class `show-modal` in order for the overlay to be active. This is set by Ionic, but while opening the modal for the first time; alas, this happens a bit late.

I've created a branch `show-modal-experiments` with debug logging in Kirby and in Ionic. To see the same logging as in the subsequent examples, check out the branch and:
```
npm run patch-ion-modal-js
npm start
```

The following is the typical output when *initially* opening a modal:
image

A couple of observations:
1) the call this.ionicModalController.create does create an ion-modal i DOM, but as it lacks the show-modal class, the overlay is not initially active.
2) After about 140 mSecs, the ionic code attaches this class to the existing element. This is sufficient to put in a second click if you click quickly.

Here is an example where I double click as fast as I can (but I've added a guard so that I log but don't process it):
image

This is how it looks if I have opened and closed the modal, and then open it a *second* time:

image

In this case, the logged time intervals are all shorter, and the ion-modal already has the desired class when it's logged the first time. So some kind of reuse is going on. When I open a modal and then open a nested modal, the nested modal also opens pretty quickly.

I've found a fix that prevents the bug on my machine, but TBH it feels more like a workaround and a hack. Anyway, once  this.ionicModalController.create returns, and the ion-modal thus exists in the DOM, I can look it up and manually attach the class. As this happens after about 70 mSecs it's soon enough to prevent at least me from double clicking.

Still, the 70 mSecs will probably vary across machines and browsers so I don't trust the fix to work in the general case.

Next steps in my investigation:
- given that the show-modal class is present on the second run at that seems to do no harm then why doesn't ionic add it initially?
- what is taking the initial call to ionicModalController.create so long? Could we "warm it up" by calling it at an earlier point in time? 

## Checklist:

The following tasks should be carried out in sequence in order to follow [the process of contributing](../CONTRIBUTING.md/#the-process-of-contributing) correctly.

### Verification

- [ ] Make sure the `NOT verified` label has been removed by a member of [team Kirby](../SUPPORT.md/#team-kirby); do not proceed until this is done

### Implementation 

### Review

- [ ] Create a pull-request.

mark-drastrup:

This PR closes #1733

## What is the new behavior?
When the button to open a modal is clicked twice rapidly, then only one modal will open, instead of two.

- [ ] Yes
- [x] No

This fix is implemented with a boolean (`isModalOpening`) that controls whether the `ionicModalController.create` method should be called or not, based on if the promise of the previous call is resolved. This approach is recommended by one of the maintainers of Ionic: https://github.com/ionic-team/ionic-framework/issues/23327#issuecomment-847028058

## Checklist:

### Reminders

### Review  

### DIFF
--- a/libs/designsystem/src/lib/components/modal/services/modal.helper.ts
+++ b/libs/designsystem/src/lib/components/modal/services/modal.helper.ts
@@ -25,7 +25,17 @@ export class ModalHelper {
     private alertHelper: AlertHelper
   ) {}
 
+  /* 
+    isModalOpening is used to prevent additional instantiations
+    of modals, while a modal is already being instatiated, but not completed.
+    This is the recommended approach by one of the maintainers of Ionic:
+    https://github.com/ionic-team/ionic-framework/issues/23327#issuecomment-847028058
+  */
+  private isModalOpening = false;
+
   public async showModalWindow(config: ModalConfig, alertConfig?: AlertConfig): Promise<Overlay> {
+    if (this.isModalOpening) return;
+
     config.flavor = config.flavor || 'modal';
     const modalPresentingElement = await this.getPresentingElement(config.flavor);
 
@@ -60,6 +70,8 @@ export class ModalHelper {
       };
     }
 
+    this.isModalOpening = true;
+
     const ionModal = await this.ionicModalController.create({
       component: config.flavor === 'compact' ? ModalCompactWrapperComponent : ModalWrapperComponent,
       cssClass: [
@@ -89,6 +101,8 @@ export class ModalHelper {
     }
 
     await ionModal.present();
+
+    this.isModalOpening = false;
 
     return {
       dismiss: ionModal.dismiss.bind(ionModal),


RasmusKjeldgaard (approved):
Great finding 💪

AndreasPetersen:
What version is this bug fix included in, or will be included in?

RasmusKjeldgaard:
@AndreasPetersen it will land in our next release on friday 18/11. 
It will be v7.2.2, or alternatively v7.3.0 if we manage to get any features in before deadline.

PR #2624, opened 2022-11-24
Added new css based Grid

thupi:

This PR closes # (insert issue number here)

## What is the new behavior?

This adds a new set of css clases that provides grid with breakpoint support and deprecates the current Grid component. The css utilities is made as an replacement for the grid component.

It uses display flex as a solution as the grid is intended to only layout horizontally and shouldn't in generel handle vertical layouts.  The implementation is made very isolated with the option of moving the scss implementation to a different package without a huge effort. The implementation takes great inspiration from twitter bootstrap and mui(an material design implementation).

- [ ] Yes
- [x] No

This uses utilizes css attribute selectors. This is done with the only goal of improving the api. However, we should be aware that custom html attibutes are noramlly prefixed with `data-*` and this MR doesn't do that. The benefits of that is a future migration to angular components instead of css would be simpler. Some of the downsides are described here [stack-overflow](https://stackoverflow.com/questions/24932391/are-custom-html-attributes-without-data-prefix-a-valid-attribute). In short the custom attributes might be used in future versions of HTML and that could collide with our usage.

Design specs for the Grid: https://www.sketch.com/s/c4f143d5-8cff-412b-9a55-0dcb29414fba

## Checklist:

### Reminders
- [ ] ~~Make sure you have implemented tests following the guidelines in: "[The good: Test](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Test)"~~

### Review  


apps/cookbook/src/app/examples/grid-example/grid-example.component.html
```diff
@@ -1 +1,29 @@
-<kirby-grid [maxColumns]="2" [cardConfigurations]="cardConfigurations"></kirby-grid>
+<div class="kirby-grid" spacing="3">
```
RasmusKjeldgaard: Consider dividing this example up, and add a small write up/explanation of both a standard grid, and then a grid with multiple breakpoints to the showcase - similar to https://mui.com/material-ui/react-grid2/#basic-grid. 

I love smaller examples with descriptive text in between.

RasmusKjeldgaard: Possibly also add a small example that just showcases setting the spacing? 

I know it is trivial, but would be great to make it very easy for the consumer to understand.

apps/cookbook/src/app/showcase/grid-showcase/grid-showcase.component.html
```diff
@@ -1,4 +1,39 @@
...
+  <h2>Grid Item</h2>
+  <cookbook-api-description-properties
+    [properties]="gridItemProperties"
+  ></cookbook-api-description-properties>
+  <p>
+    * If you have concerns about HTMl validity, you can prefix the attributes above with
+    <code>data-*</code>. That way your HTML will be valid and it will function exactly the same.
+    Here is an example:
+    <code>&lt;div class="kirby-grid-item" data-xs="12" data-sm="6"&gt;&lt;/div&gt;</code>
+  </p>
```
RasmusKjeldgaard: Consider moving this section up in relation to one of the code snippets that showcases the non data-* approach, as a more general take away. I think it is great insight.

apps/cookbook/src/app/showcase/grid-showcase/grid-showcase.component.ts
```diff
@@ -1,11 +1,52 @@
...
+      defaultValue: '3',
+      type: ['0', '1', '2', '3', '4', '5'],
+    },
+  ];
+  gridItemProperties: ApiDescriptionProperty[] = [
+    {
+      name: 'xs',
+      description: 'Defines the column span for xtra-small screen sizes.',
+      defaultValue: '',
+      type: ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12'],
```
RasmusKjeldgaard: Can we do something like type: ['0' - '12'] for readability?

apps/cookbook/src/app/showcase/grid-showcase/grid-showcase.component.html
```diff
@@ -1,4 +1,39 @@
...
+  <cookbook-api-description-properties
+    [properties]="gridProperties"
+  ></cookbook-api-description-properties>
+
+  <h2>Grid Item</h2>
+  <cookbook-api-description-properties
+    [properties]="gridItemProperties"
+  ></cookbook-api-description-properties>
+  <p>
+    * If you have concerns about HTMl validity, you can prefix the attributes above with
```
RasmusKjeldgaard:
```suggestion
    * If you have concerns about HTML validity, you can prefix the attributes above with
```

libs/core/src/scss/components/_grid.scss
```diff
@@ -0,0 +1,97 @@
...
+  'xs': 0,
+  'sm': 632px,
+  'md': 768px,
+  'lg': 992px,
+  'xl': 1200px,
+);
+
+.kirby-grid {
+  --kirby-grid-columns: 12;
+  --kirby-grid-columnSpacing: 24px;
```
RasmusKjeldgaard: Consider sticking to kebab case throughout, to align better with existing custom properties. 
```suggestion
  --kirby-grid-column-spacing: 24px;
```

RasmusKjeldgaard: Goes for this whole file 😃

libs/core/src/scss/components/_grid.scss
```diff
@@ -0,0 +1,97 @@
...
+  --kirby-grid-rowSpacing: 24px;
+
+  min-width: 0;
```
RasmusKjeldgaard: Why `min-width: 0` here? Is that not the initial value?

libs/core/src/scss/components/_grid.scss
```diff
@@ -0,0 +1,97 @@
...
+  box-sizing: border-box;
+  display: flex;
+  flex-flow: row wrap;
+  margin: calc(var(--kirby-grid-rowSpacing) / -2) calc(var(--kirby-grid-columnSpacing) / -2);
```
RasmusKjeldgaard: Clever! The way I understand this, we use negative margins to be able to add column and row padding consistently around every element even when it is adjacent to the edge of the grid? 

Could we add a comment explaining that, for future us?

libs/core/src/scss/components/_grid.scss
```diff
@@ -0,0 +1,97 @@
...
+
+  &[spacing='1'],
+  &[data-spacing='1'] {
+    --kirby-grid-columnSpacing: 8px;
+    --kirby-grid-rowSpacing: 8px;
```
RasmusKjeldgaard: UI/UX question: Do you know if it has been considered if we want the spacing to scale together with text, if users use bigger text sizes in their browser? 

Would be trivial to base spacing on rem units instead, if that is desired behavior. 
If we want the spacing to be consistent across any font-size, discard this comment.

libs/designsystem/src/lib/components/grid/grid.component.ts
```diff
@@ -5,6 +5,9 @@ import { ScssHelper } from '../../scss/scss-helper';
 import { BreakpointHelperService } from './breakpoint-helper.service';
 import { GridCardConfiguration } from './grid-card-configuration';
 
+const GRID_DEPRICATION_WARNING =
+  'Deprecation warning: The "kirby-grid" component will be removed in a future release of Kirby designsystem. Consider using kirby-grid, kirby-grid-item css-classes as an alternative solution.';
```
RasmusKjeldgaard: Let us link to grid docs here too (https://cookbook.kirby.design/#/home/showcase/grid)

libs/core/src/scss/components/_grid.scss
```diff
@@ -0,0 +1,97 @@
...
+    @include column-style($breakpoint, 12);
+  }
+}
+
+.kirby-grid-item {
+  flex-grow: 1;
+  flex-basis: 0;
+  min-width: 0;
+  max-width: 100%;
+  box-sizing: border-box !important;
```
RasmusKjeldgaard, line 101: Why is important needed here?

RasmusKjeldgaard (changes requested):

libs/core/src/scss/components/_grid.scss
thupi: Nope, there is an issue with the flexbox spec(https://github.com/philipwalton/flexbugs#flexbug-1) but you are right, it is only needed on the items not the container :-)

libs/core/src/scss/components/_grid.scss
thupi: I would expect no as the definitions of the breakpoint is made to respect desired min width for columns. I also think this would add complexity for the component as on small devices where we would need more space for the bigger text we would now also use more space for the gap :-) ?

libs/core/src/scss/components/_grid.scss
thupi: I did this as the grid system relies on the box-sizing strategy. If anyone were to change this it would cause the entire grid to fail. I wanted to ensure this wouldn't happen even if someone would add something like that wouldn't happen if some projects were to include something that rewrites box-boxing with a "*" css selector :-) 

But I could remove it if you would prefer :-) ?

RasmusKjeldgaard: Good call! Consider adding a comment explaining that.

libs/core/src/scss/components/_grid.scss
```diff
@@ -0,0 +1,107 @@
...
+  &[data-spacing='0'] {
+    --kirby-grid-column-spacing: 0px;
+    --kirby-grid-row-spacing: 0px;
+  }
+
+  &[spacing='1'],
+  &[data-spacing='1'] {
+    --kirby-grid-column-spacing: 8px;
+    --kirby-grid-row-spacing: 8px;
+  }
```
RasmusKjeldgaard: Remember to use our size tokens for all these pixel values, e.g. 
`utils.size('xxs')` here

libs/core/src/scss/components/_grid.scss
RasmusKjeldgaard: Good point, lets leave it at using our existing design tokens here 👍

RasmusKjeldgaard (dismissed):

RasmusKjeldgaard:
Crocodile Hunter Yes

RasmusKjeldgaard (approved):

PR #2627, opened 2022-11-25
Bugfix: 'first' and 'last' not correctly applied when using Kirby list with sections

### LINKED ISSUE #1895: [BUG] Postings - Check swipe has squared corners in the first row in a section

1) Open the postings/transaction list
2) Swipe on the first row in a section
3) The check button is not with round corners - it is also squared on the right side of the list

4.0.8

Corners are round.

## Add any screenshots
IMG_0097

- OS: iOS (iPad Air iOS 15.1)

<hr />

## Checklist:

### Verification

- [ ] Make sure the `NOT verified` label has been removed by a member of [team Kirby](https://github.com/kirbydesign/designsystem/blob/master/.github/SUPPORT.md#team-kirby); do not proceed until this is done

### Implementation 

### Review

- [ ] Create a pull-request.

AgreSanGit:

This PR closes #1895 #

## What is the new behavior?
Refactoring of _getBoundaryClass in list component which correctly applies the classes 'first' and 'last' on elements in the list, when using kirby list with sections.

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


libs/designsystem/src/lib/components/list/list.component.spec.ts
```diff
@@ -222,4 +222,59 @@ describe('ListComponent', () => {
...
+      expect(list[0].classList).toContain('first');
+      expect(list[0].classList).not.toContain('last');
+
+      //Last element
+      expect(list[1].classList).toContain('last');
+      expect(list[1].classList).not.toContain('first');
+    });
+  });
+
+  describe('List have 3 elements', () => {
```
mark-drastrup: You could also divide it up into separate `it` blocks, so the first `it` block would test the first element, the second `it` block the second element and I think you get the idea 😄 It could look something like this:

```
  describe('List have 3 elements', () => {
    beforeEach(() => {
      spectator.setInput({
        items: [transactions[0], transactions[0], transactions[0]],
      });
      
      it('should apply the CSS class 'first' on the first element.', () => {
        const list = spectator.queryAll('ion-item-sliding');

        expect(list[0].classList).toContain('first');
        expect(list[0].classList).not.toContain('last');
      })
    })
  ... and similar approach for testing the middle and last item
  }
```

The idea here is just to let each `it` block be more succinct and test one very specific thing. But this is just a suggestion and it's totally up to you if you want to keep it the way you already did it, which is also perfectly fine :)

mark-drastrup (changes requested):
@AgreSanGit it looks really good! But there is a small bug in the "List with header and footer" example. When a list has a header and a footer, then the first and last item shouldn't have rounded corners:
Screenshot 2022-11-29 at 15 55 46

Screenshot 2022-11-29 at 15 56 32

libs/designsystem/src/lib/components/list/list.component.ts
```diff
@@ -243,12 +243,15 @@ export class ListComponent implements OnInit, AfterViewInit, OnChanges {
   }
 
   _getBoundaryClass(index: number, section?: any[]): BoundaryClass {
-    const _items = section || this.items;
+    if (index === 0) {
+      return section.length === 1 ? 'first last' : 'first';
+    }
 
-    if (index === 0 || _items[index - 1]?.headingName) return this.headerTemplate ? null : 'first';
```
mark-drastrup, line 248: I suspect that the issues with rounded corners, when there is a header/footer is caused, because the checks for `headingName` and `footerTemplate` were removed :)

libs/designsystem/src/lib/components/list/list.component.spec.ts
AgreSanGit: Thanks I have tought about it and I will keep my implementation

libs/designsystem/src/lib/components/list/list.component.ts
AgreSanGit: Very good spot on the bug I have refactored so it solves this bug now 🗡️

libs/designsystem/src/lib/components/list/list-item/list-item.component.ts
```diff
@@ -15,7 +15,7 @@ import { ThemeColor } from '@kirbydesign/core';
 import { PlatformService } from '../../../helpers/platform.service';
 import { ListSwipeAction, ListSwipeDirection, ListSwipeEnd } from '../list-swipe-action.type';
 
-export type BoundaryClass = 'first' | 'last';
+export type BoundaryClass = 'first' | 'last' | 'first last';
```
RasmusKjeldgaard: Just a quick thought. Could the logic for applying these classes be modified so whenever an item is both first and last, it just gets both the type 'first' and 'last' instead of 'first last'?

libs/designsystem/src/lib/components/list/list.component.spec.ts
RasmusKjeldgaard: Would be great for readability to implement Marks suggestion. 
This is hard to digest: 
'should apply the CSS class 'first' on the first element and apply 'last' on last element, the element(s) between first and last should neither have 'first' or 'last' applied'

RasmusKjeldgaard: Whenever that test fails, it takes some time to figure out exactly what failed, whereas if it was just `should apply the CSS class 'first' on the first element` that fails I know instantly what is wrong

libs/designsystem/src/lib/components/list/list-item/list-item.component.ts
AgreSanGit: I was not able to find a logic that would be able to apply 'first' and 'last' on one element without using 'first last' because of the method return type.

libs/designsystem/src/lib/components/list/list.component.spec.ts
AgreSanGit: Okay noted I can see the logic behind that! I will change it

libs/designsystem/src/lib/components/list/list.component.integration.spec.ts
```diff
@@ -174,4 +174,107 @@ describe('ListComponent', () => {
       expect(lastItem).toHaveComputedStyle({ 'padding-bottom': '0px' });
     });
   });
+
+  fdescribe('when a list have 1 element', () => {
```
RasmusKjeldgaard: Beware of all the fdescribes! Needs to be removed before merge.

AgreSanGit: removed

mark-drastrup (approved):
Great job @AgreSanGit! It loooks really clean now 😄 Now all that's left is to push to button!

Alt Text

PR #2636, opened 2022-11-29
Replace avatar with image shadow with stroke

### LINKED ISSUE #2622: [Enhancement] Replace avatar shadow with stroke
In the process of streamlining elevation, we will remove the shadow (z2 elevation) on the component "Avatar with Image".

We will **add** an inside stroke instead.

Stroke spec:
Inside stroke: #ebebeb — (Kirby / Style 01 / System / grey-02 (Semi-light))

Avatar-with-image

AgreSanGit:

This PR closes #2622 

## What is the new behavior?

Replaces avatar with image shadow with inside stroke instead

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  

### DIFF
--- a/apps/cookbook/src/app/showcase/avatar-showcase/avatar-showcase.component.ts
+++ b/apps/cookbook/src/app/showcase/avatar-showcase/avatar-showcase.component.ts
@@ -36,8 +36,14 @@ export class AvatarShowcaseComponent {
       type: ['true', 'false'],
     },
     {
-      name: 'shadow',
+      name: 'shadow (deprecated - use stroke instead)',
       description: 'Adds a box-shadow to the avatar.',
+      defaultValue: 'false',
+      type: ['true', 'false'],
+    },
+    {
+      name: 'stroke',
+      description: 'Adds a inside stroke to the avatar.',
       defaultValue: 'false',
       type: ['true', 'false'],
     },
--- a/apps/cookbook/src/app/showcase/avatar-showcase/avatar-showcase.component.ts
+++ b/apps/cookbook/src/app/showcase/avatar-showcase/avatar-showcase.component.ts
@@ -43,7 +43,7 @@ export class AvatarShowcaseComponent {
     },
     {
       name: 'stroke',
-      description: 'Adds a inside stroke to the avatar.',
+      description: 'Adds an outline to the avatar.',
       defaultValue: 'false',
       type: ['true', 'false'],
     },
--- a/libs/designsystem/src/lib/components/avatar/avatar.component.html
+++ b/libs/designsystem/src/lib/components/avatar/avatar.component.html
@@ -1,4 +1,4 @@
-<div class="avatar" [ngClass]="{ shadow: shadow, overlay: overlay }">
+<div class="avatar" [ngClass]="{ shadow: shadow, overlay: overlay, stroke: stroke }">
   <img *ngIf="imageSrc" [src]="imageSrc" [attr.alt]="altText" />
   <ng-content *ngIf="!text" select="kirby-icon"></ng-content>
   <span class="avatar-text" *ngIf="text">{{ text }}</span>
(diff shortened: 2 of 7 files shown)


libs/designsystem/src/lib/components/avatar/avatar.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="avatar" [ngClass]="{ shadow: shadow, overlay: overlay }">
+<div class="avatar" [ngClass]="{ shadow: shadow, overlay: overlay, stroke: stroke }">
```
mark-drastrup: You could also apply the `stroke` class if `shadow` is true, and then remove `shadow` from this class binding. Remember to remove the class from the styles as well.

libs/designsystem/src/lib/components/avatar/avatar.component.scss
```diff
@@ -46,7 +46,11 @@ $badge-diameter: utils.$avatar-badge-size;
   --kirby-icon-font-size: #{utils.icon-font-size('sm')};
 
   &.shadow {
```
mark-drastrup: This is the class you should remember to remove, when you have refactored the class binding in `avatar.component.html` :)

mark-drastrup (changes requested):
Really nice work @AgreSanGit! 😄 I have added a few comments to the code. And I think that we should remove the shadow example from the "Avatar with image" section, plus mark the `shadow` input as deprecated in the properties table, and tell the users to use `stroke` instead. Lastly, the `stroke` property should be added to the table as well.

libs/designsystem/src/lib/components/avatar/avatar.component.html
AgreSanGit: Updated!

libs/designsystem/src/lib/components/avatar/avatar.component.scss
AgreSanGit: removed!

apps/cookbook/src/app/showcase/avatar-showcase/avatar-showcase.component.ts
```diff
@@ -36,10 +36,16 @@ export class AvatarShowcaseComponent {
...
     {
-      name: 'shadow',
+      name: 'shadow (deprecated - use stroke instead)',
       description: 'Adds a box-shadow to the avatar.',
       defaultValue: 'false',
       type: ['true', 'false'],
     },
+    {
+      name: 'stroke',
+      description: 'Adds a inside stroke to the avatar.',
```
mark-drastrup:
```suggestion
      description: 'Adds an outline to the avatar.',
```

mark-drastrup (changes requested):
I just have 1 more really small comment, with a suggestion to a text correction :)

libs/designsystem/src/lib/components/avatar/avatar.component.ts
```diff
@@ -29,4 +30,13 @@ export class AvatarComponent {
   get _cssClass() {
     return [this.themeColor, this.size].filter((cssClass) => !!cssClass);
   }
+
+  ngOnInit(): void {
+    if (this.shadow) {
```
kodeaben: Consider a "return early" approach here: https://medium.com/swlh/return-early-pattern-3d18a41bba8
This makes the code easier to read IMO.
So it will look like this: 

```javascript
public ngOnInit(): void {
    if (!this.shadow) {
     return;
    }
    
    this.stroke = true;
    console.warn(
      'Shadow input binding on avatar will be deprecated next major. Use stroke instead'
      );
}
```

AgreSanGit: It does look better! Thank you!

mark-drastrup (approved):
Nice work! Ship it! alt text

kodeaben (approved):

mark-drastrup:
@AgreSanGit remember to add a label! :)

PR #2669, opened 2022-12-13
Prepare release v7.4.0

RasmusKjeldgaard:


mark-drastrup (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "designsystem",
-  "version": "7.3.0",
+  "version": "7.4.0",
```
kodeaben, line 3: Hvorfor bumper du en minor på designsystem når du kun patcher på core?

RasmusKjeldgaard: Core is experimental, and hence in v0.x.y - everything is a patch because the project is in flux and does not yet adhere to SemVer. This is standard practice.

RasmusKjeldgaard: The designsystem is minor because the new release contains both features and fixes.

PR #2698, opened 2022-12-21
Refactor how kirby list items receive hover effect

### LINKED ISSUE #2638: [BUG] Kirby list pointer and hightlight sometimes works
Hi Kirby

The pointer and hightlight on hover on Kirby list sometimes works and sometimes It does not work. I can't fingure out what triggers the pointer and on hover hightligt however I my best guess is that the ```(itemSelect)="onItemSelect($event)"``` is the trigger. The code below is a anonymized version of my actual code. In the real code of my project the ```[disableSelectionHighlight]``` does not work and it seems like it triggers the pointer and on hover hightligt sometimes. It could however also just be a coincidence. Below are two pictures of the same code runned mutipletimes where it works and then it does not work. 

I would like it to always show the pointer and highlight on hover. 

HTML
```
<kirby-page>
  <kirby-page-content>
    <kirby-list [items]="testdata" [showDivider] = "true" (itemSelect)="onItemSelect($event)" *ngIf="HasLoaded$() | async; else spinner">
      <kirby-list-header *kirbyListHeader class="header-size"
      ><h3>
        Header
      </h3>
      </kirby-list-header>
      <kirby-item *kirbyListItemTemplate="let item">
        <p>{{item.label}}</p>
      </kirby-item>
    </kirby-list>
  </kirby-page-content>
  <ng-template #spinner>
    <kirby-spinner></kirby-spinner>
  </ng-template>
</kirby-page>
``` 

TS
```
export class TestComponent implements OnInit {
  testdata: any = [];
  constructor(
    public outstandingsService: OutstandingsService,
  ) {}

  ngOnInit(): void {
    this.Service.update();
    this.Service.getSubject().subscribe((data) => {
      this.testdata = [
        { label: "mock", status: "ok"},
        { label: "mock", status: "ok"},
        { label: "mock", status: "ok"},
        { label: "mock", status: "ok"}
      ];
    });
  }

  onItemSelect($event: any) {

  }

}
```

image
image

mark-drastrup:

This PR closes #2638

## What is the new behavior?
Previously `kirby-items` in a list would automatically receive a hover effect, when a callback were provided to `itemSelect`. The internal logic in `kirby-list` that is responsible for marking `kirby-items` as selectable in a list breaks, if the list is conditionally rendered using an `observable` with the `async pipe`. This issue is easily fixed by providing `[selectable]="true"` to the `kirby-item` component in the template, if the user wishes to give the item a hover effect

- [x] Yes
- [ ] No

### Migration notes
It is now required to pass `[selectable]="true"` to `kirby-item` in `kirby-list` if items in the list should have a hover effect.

## Checklist:

### Reminders

### Review  


MIGRATION.md
```diff
@@ -2,6 +2,20 @@
...
 
+## Version 8
+
+- [Version 8](#version-8)
+  - [Components](#components-v8)
+    - [List](#list-v8)
+
+<h3 id="components-v8">Components</h3>
+
+<h4 id="list-v8">List</h4>
```
RasmusKjeldgaard, line 13: What is the reason for using html here instead of markdown (`<h3>` vs `###`)?

mark-drastrup: It's because of the links in the table of contents for each version. It was working correctly while we only had a table of contents for version 6, where we fx. created a link like this `[Components](#components)`. This part `#components` automatically creates a link to a heading with the title "components". But that becomes a problem when we have multiple table of contents in the file, as the link will only point to the first instance of the title. Fx. `[Components](#components)` in the table of contents for version 8 will point to the "components" title in version 6. By using HTML we are able to make links using id's instead.

There is a markdown solution that also works, but I think the HTML solution is preferable. To achieve the same effect in markdown, then we would need to create the links by adding the index of the title as a suffix like this: [Components](#components-0). But that becomes messy if there is a title that is not included in a table of contents. Fx:

```
- [Version 8](#version-8)
  - [Platform Support](#platform-support-1)
  - [Components](#components-2)
    - [Alert](#alert-1)
    - [List](#list-2)
    - [Page](#page-0)

- [Version 7](#version-7)
  - [Components](#components-1)
    - [List](#list-1)

- [Version 6](#version-6)
  - [Platform Support](#platform-support-0)
  - [Components](#components-0)
    - [Alert](#alert-0)
    - [List](#list-0)
```

It starts to become tricky in version 8 in the example above, because some titles have already appeared multiple times and some appears for the first time, which results in a different index, so we have to keep track of how many times a title appears to make a link, instead of just using an HTML tag and link to the specified id.

That was a long response  😂

RasmusKjeldgaard: That makes a lot of sense. Would a markdown solution where we just append the version work?
```
- [Version 6](#version-6) 
   - [Platform Support](#platform-support-6) 
   - [Components](#components-6)
```

Or will there still be instances where that becomes messy? Otherwise I think it is the most straight forward approach that makes sense for whoever will have to do it next time.

RasmusKjeldgaard: I'll approve and you can just use your best judgement on what you like the most.

RasmusKjeldgaard (approved):
Solid stuff! Very nice to have that awful foreach loop that _someone_ introduced removed again 😂

MIGRATION.md
mark-drastrup: That could work, but it then we would have to include the number in the actual title like this:

```
- [Version 6](#version-6) 
   - [Platform Support](#platform-support-6) 
   - [Components](#components-6)

Platform support 6
Lorem ipsum......

Components 6
Lorem ipsum.....
```
I think the output of the HTML solution looks cleaner :)

PR #2742, opened 2023-01-11
Only add API table scrollbar when content overflows

RasmusKjeldgaard:

This PR closes no issue

## What is the new behavior?
During a review I noticed that our API tables always have a horizontal scrollbar even if content is not scrollable. 
I suggest to make it only appear when content is scrollable, e.g. on mobile, so the scroll track disappears on larger screens.
 
image

- [ ] Yes
- [x] No


mark-drastrup (approved):
Nice catch!

PR #2747, opened 2023-01-11
Fix issue where iphone overscrolls and hides popout in cookbook

RasmusKjeldgaard:

This PR closes no issue

## What is the new behavior?
The iPhone has position: sticky which causes an annoying issue with unclickable pop out links.

https://user-images.githubusercontent.com/42470636/211847208-46849eef-9332-4f6f-ba15-232e5c3de081.mov

This is fixed, and we now also use our own kirbified pop out link styling class instead of a homebrew solution.

- [ ] Yes
- [x] No


mark-drastrup (approved):
Amazing 🥲

Issue #2744, opened 2023-01-11
[COMPONENT] Menu


NillerW:
- [x] I have written a [good issue](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Issue) 

## Describe the solution you'd like
For a Bankdata project, we need a "popup" menu - see mock up below. 

In the context of the mock up, the menu is appearing after click on the button with the 3 dots.

![image](https://user-images.githubusercontent.com/107917716/211763069-fbf55196-d8dd-4c7d-8532-5a4131d47ec1.png)

**Requirements** for the menu:
- [ ] Supports multiple actions (items)
- [ ] **Not** close on action click 
- [ ] Trigger event on action click, with option to include data with the event
- [ ] Support custom content for each action
- [ ] Work as a popup/overlay
- [ ] Display/Hide functionality supported from outside of the menu it self. For example on button click as shown in mock-up
- [ ] Supports left/right alignment. One would expect the menu to have it's top left corner to be directly below the button that activated the menu, but for our use case, we need the top right corner to be directly below the button.

## Have you considered any alternatives?
Action sheet was considered, but not suitable due to:
- Taking over whole screen
- Closes on click
- Doesn't support custom content

Fab sheet was considered, but not suitable due to:
- Closes on click
- Doesn't support custom content
- Is fixed to bottom right corner of page

No other components seemed relevant. 

## Are there any additional context?
As we need this for an ongoing project, we would prefer to get the component done sooner rather than later.

<hr />

## Checklist:

The following tasks should be carried out in sequence in order to follow [the process of contributing](https://github.com/kirbydesign/designsystem/blob/main/.github/CONTRIBUTING.md/#the-process-of-contributing) correctly.

### Refinement

- [x] Request that the issue is [UX refined](https://github.com/kirbydesign/designsystem/blob/main/.github/CONTRIBUTING.md/#ux-refinement); do not proceed until this is done.
- [x] Request that the issue is [tech refined](https://github.com/kirbydesign/designsystem/blob/main/.github/CONTRIBUTING.md/#tech-refinement); do not proceed until this is done.

### Implementation 
The contributor who wants to implement this issue should: 

- [x] Make sure you have read: "[Before you get coding](https://github.com/kirbydesign/designsystem/blob/main/.github/CONTRIBUTING.md/#before-you-get-coding)".
- [x] Signal to others you are working on the issue by assigning yourself.
[...]

PR #2805, opened 2023-01-24
Component/2744 popup component

### LINKED ISSUE #2721: [COMPONENT] Multi select

For use in an ongoing Bankdata project, we need a searchable dropdown, that supports multiple selections.

Since there will be upwards of 500 items in the dropdown for our use case, it is crucial that the dropdown is searchable.
Additionally, the requirement of selecting multiple values from the dropdown means that the current components are not sufficient.

Some information from UX is provided below.

### Mock-up
mockup

### Requirements

- [ ] When no items are selected, and the input is not focused, a placeholder text should be shown (Example 01)
- [ ] When the input is focused, the dropdown should start showing (Example 02).
  * This behaviour may have to change slightly in case of async search.
- [ ] The items shown in the dropdown should be filtered as the user types input. (Example 03)
  * Techincal note: A debounce should probably be forced here. UX previously suggested 500 ms, but this is subject to change.
- [ ] The input should show number of selected items when not focused. (Example 06)
- [ ] It should be possible to clear all selections when the input is not focused. (Example 06)
- [ ] When the dropdown is open, selected items should be placed first in the list. (Examples 05 and 07)
- [ ] There should be an option for displaying the total number of items matching a dropdown item. (Example B01)
  * I.e. an outer list with duplicates is filtered based on the selection in the multi select component.
    Each option in the dropdown shows the number of outer list elements with that value.
  * This would probably be achieved by allowing custom item label rendering.
- [ ] It should be possible to mark an item as disabled. (Example B01)
- [ ] It should be possible to mark the entire component as disabled.
- [ ] Provide an async search option, using `Observable` as suggested by @kodeaben [here](https://github.com/kirbydesign/designsystem/issues/2721#issuecomment-1372313120)

#### Single select mode

- [ ] The multi select functionality should be enabled if the `multiple` Input flag is set to true, otherwise the component should function as a searchable dropdown (Only one item can be selected)
- [ ] When in single select mode, the text / label of selected item should be shown instead of the number of selected items
- [ ] When in single select mode, the clear button should still appear when the input is not focused.

### Other

It is not yet clear if scroll virtualization is desirable.

<hr />

## Checklist:

### Refinement

- [x] Request that the issue is [UX refined](https://github.com/kirbydesign/designsystem/blob/main/.github/CONTRIBUTING.md/#ux-refinement); do not proceed until this is done.
- [x] Request that the issue is [tech refined](https://github.com/kirbydesign/designsystem/blob/main/.github/CONTRIBUTING.md/#tech-refinement); do not proceed until this is done.

### Implementation 

- [x] Make sure you have read: "[Before you get coding](https://github.com/kirbydesign/designsystem/blob/main/.github/CONTRIBUTING.md/#before-you-get-coding)".

### Review

NillerW:

This PR doesn't directly close any issues, but it is needed to close #2721 and #2744 down the road and provide general functionality for popup/floating components. 

The PR Adds a new and better way to handle floating/popup elements and is meant to replace the current popup component. 

This directive is meant for internal use, to create Kirby components such as Dropdown, Tooltip, Popup. etc. and not to be exposed outside the Kirby design system.

## What is the new behavior?

This PR introduces the KirbyFloating directive. This directive can be added directly to HTML-tags to make the tag float relative to another tag. 

The KirbyFloating directive implementation provides the following benefits:
- Ease-of-use: It's simply to attach two attributes to an existing element.
- Reference element and floating element does not have to be placed directly next to each other.
- Not introducing an unnecessary component tag to make the actual content float.
- Trigger events to toggle visibility of the floating element, like click and hover, are handled automatically by the directive.
- Automatic handling of expand direction, based on available size.
- Automatic positioning of floating content based on available size. 

With below image as reference, it can be seen how the new directive works. 

image

Required attributes:
- KirbyFloating: Provides the functionality.
- Reference: Element for which the floating content should be placed next to.

Optional attributes:
- Triggers: How the floating element should be set to show/hide. "Click" will attach a click event to the reference element, which toggles display of the floating element. Supports single value and array.
- Placement: How the element should be placed. Example top, down (under reference element centered), start-down (Under reference element aligned with start of element) etc.

Further inputs to the floating directive is likely to come, based on use-cases and requirements.

There are currently no tests in agreement with @RasmusKjeldgaard, until we conclude what we actually want to test for a general directive like this.

- [ ] Yes
- [X] No

## Checklist:

### Review  


NillerW:
Will be adding disabled input, given that trigger functionality is placed in the popup and not the element one would expect the trigger functionality to be placed in.

libs/designsystem/src/lib/directives/floating/floating.directive.ts
```diff
@@ -0,0 +1,262 @@
...
+    this.renderer.setStyle(
+      this.elementRef.nativeElement,
+      'z-index',
+      DesignTokenHelper.zLayer('popover')
+    );
+  }
+
+  private updateHostElementPosition(): void {
+    this.config = {
+      placement: this.placement ?? 'bottom-start',
```
RasmusKjeldgaard: Consider replacing this with `placement: this.placement`, then combined with a default value for the property: 
`private _placement: Placement = 'bottom-start';`

Then it is much easier to understand the default behavior at a glance.

libs/designsystem/src/lib/directives/floating/floating.directive.ts
```diff
@@ -0,0 +1,262 @@
...
+      middleware: this.getMiddlewareConfig(),
+      strategy: 'absolute',
```
RasmusKjeldgaard: Would love if we could already now expose this via an input. 
The dropdown will for sure need to have an option for fixed positioning.

libs/designsystem/src/lib/directives/floating/floating.directive.ts
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  @HostListener('document:mousedown', ['$event'])
+  public onMouseClick(event): void {
+    const clickedOnHost: boolean = this.elementRef.nativeElement.contains(event.target);
+    clickedOnHost ? this.handleClickInsideHostElement() : this.handleClickOutsideHostElement(event);
+  }
+
+  private _placement: Placement;
+  private _triggers: Array<TriggerEvent> | TriggerEvent = 'click';
```
RasmusKjeldgaard: What are your thoughts about converting this to always expect an array? 
We would then have to go `[triggers]="['click']"` for single triggers, but it would make this code a bit easier to read and maintain and we can do without isArray checks throughout. 

The input is named triggers with an **_s_** after all 🤔

libs/designsystem/src/lib/directives/floating/floating.directive.ts
```diff
@@ -0,0 +1,262 @@
...
+    this._triggers = eventTriggers;
+    this.tearDownEventHandling();
+    this.setupEventHandling();
+  }
+
+  private get triggers(): Array<TriggerEvent> | TriggerEvent {
+    return this._triggers;
+  }
+
+  @Input() private isDisabled: boolean = false;
```
RasmusKjeldgaard, line 102: Now that this is an internal utility that we are not documenting in the cookbook, I think we should do some extra effort in writing some thorough comments on what each input does.

RasmusKjeldgaard: Maybe how it is done in [shoelace](https://github.com/shoelace-style/shoelace/blob/current/src/components/popup/popup.ts#L52) can be of some inspiration here

libs/designsystem/src/lib/directives/floating/floating.directive.ts
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    if (Array.isArray(this.triggers)) {
+      (this.triggers as Array<TriggerEvent>).forEach((trigger: TriggerEvent) =>
+        this.attachTriggerEventToReferenceElement(trigger)
+      );
+    } else {
+      this.attachTriggerEventToReferenceElement(this.triggers);
+    }
+  }
```
RasmusKjeldgaard: If we always have triggers as an array as mentioned above this could probably be 

```
this.triggers.forEach((trigger: TriggerEvent) => 
        this.attachTriggerEventToReferenceElement(trigger)
);

```

libs/designsystem/src/lib/directives/floating/floating.directive.ts
RasmusKjeldgaard: If we do this, and type it with `Strategy` we should also be able to set the host styling dynamically based on `this.strategy` in `addFloatStylingToHostElement()`: 

`this.renderer.setStyle(this.elementRef.nativeElement, 'position', this.strategy);`

Thoughts?

RasmusKjeldgaard (changes requested):

RasmusKjeldgaard:
Incredibly nice work, I already love this utility! 
I have thrown in my first round of comments so we can start the discussion while I get my hands dirty by trying to use the utility for our dropdown.

libs/designsystem/src/lib/directives/floating/floating.directive.ts
NillerW: Sure we can do that, it just requires some extra testing on specific use case

libs/designsystem/src/lib/directives/floating/floating.directive.ts
NillerW: I suspect that 80% of the cases, the user would need only one trigger. It would then be kind of cumbersome to always have to put in an array, due to still supporting the 20% of the cases where we have two or more triggers.

As I see it, it's a QoL for the user, while the maintainer gets a bit of extra work. 

I'm not married to one or the either and can see pros/cons on both, so if you have a preference we can go with that.

RasmusKjeldgaard: How about a nice compromise and just always convert it to an array in the setter, so we always know what we are working with.

PR #2822, opened 2023-02-02
Fix rare title intersection observer bug

thupi:
## What is the new behavior?

There is a initializeTitle() function in Kirby-page which used to only run when the component enter or leave. However, in rare cases the reference to the title change without unmount the page component. In this case, the new title element won't be observed and there title will no be hidden and shown according to the scroll position.

- [ ] Yes
- [X] No

I made sure to test if running the initializeTitle() more often would cause performance issues. However, I found that assigning the same value to an existing angular propter won't cause and re-renders.

The issues couldn't be reproduced in the flows app. However, we could reproduce it in "relationsbanken". The bug was reported in kirby-guild(https://bankdata.slack.com/archives/CGKJ1RBM4/p1674640659682029)

## Checklist:

### Reminders

### Review  
- [ ] ~~Request that the changes are [UX reviewed](https://github.com/kirbydesign/designsystem/blob/main/.github/CONTRIBUTING.md/#ux-review) (only necessary if your PR introduces visual changes)~~

### DIFF
--- a/libs/designsystem/src/lib/components/page/page.component.ts
+++ b/libs/designsystem/src/lib/components/page/page.component.ts
@@ -287,7 +287,6 @@ export class PageComponent
   }
 
   ngOnChanges(changes: SimpleChanges): void {
-    console.log('changes', changes);
     if (changes.titleMaxLines) {
       this.fitHeadingConfig = {
         ...this.fitHeadingConfig,
--- a/libs/designsystem/src/lib/components/page/page.component.ts
+++ b/libs/designsystem/src/lib/components/page/page.component.ts
@@ -287,6 +287,7 @@ export class PageComponent
   }
 
   ngOnChanges(changes: SimpleChanges): void {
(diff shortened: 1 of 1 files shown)


libs/designsystem/src/lib/components/page/page.component.ts
```diff
@@ -287,6 +287,7 @@ export class PageComponent
   }
 
   ngOnChanges(changes: SimpleChanges): void {
+    console.log('changes', changes);
```
kodeaben: Please delete this :)

kodeaben (changes requested):

kodeaben (approved):

mictro:
This issue was reported by Janusa Vijayathurai (@javi) but it was reported that he is no longer able to reproduce the problem.
Perhaps the new routing concept will make this a real issue - this still waits to be proven.

PR #2876, opened 2023-02-20
Add portal directive

### LINKED ISSUE #2874: [Enhancement] Portal Directive
Add portal directive. This directive should be able to portal the host element it's placed on into another DOM element based on a reference to the desired portal outlet.

This is needed for #2805 but can be used generally. 

The directive has already been created.

NillerW:

This PR closes #2874 and is need to close #2805 

## What is the new behavior?

Adds a portal directive for internal use. Adding portal directive to a host element, and providing a DOM element as outlet, will portal the host element to be a child of the DOM outlet in the DOM.  

- [ ] Yes
- [x] No

Is need for floating directive #2805, to ensure that the floating element will stay on top of other elements. Portalling is sometimes needed due to stacking context issues.

Solution extends CdkPortalOutlet to gain functionality for attaching/detaching the host element, and for proper cleanup.

### Review  
- [ ] Request that the changes are [UX reviewed]


libs/designsystem/src/lib/directives/portal/portal.directive.ts
```diff
@@ -0,0 +1,90 @@
...
+  ViewContainerRef,
+} from '@angular/core';
+
+/**
+ * @summary PortalDirective is a utility that lets you portal the host element into another DOM element.
+ * This is designed to be beneficial when there's issues with the stacking context.
+ *
+ * Extends CdkPortalOutlet to gain functionality for attaching/detaching the host element, and for proper cleanup.
+ *
+ * @status In development
```
RasmusKjeldgaard: I love the summary but maybe we don't need the 'in development' status - it might frighten consumers from using the component.

libs/designsystem/src/lib/directives/portal/portal.directive.ts
```diff
@@ -0,0 +1,90 @@
...
+  public constructor(
+    private elementRef: ElementRef,
+    private componentFactoryResolver: ComponentFactoryResolver,
+    private viewContainerRef: ViewContainerRef
+  ) {
+    super(componentFactoryResolver, viewContainerRef);
+  }
+
+  /**
+   * Nearly identical to base, yet we want to use provided outlet instead of directive host as new content
```
RasmusKjeldgaard: Does `base` here mean the super class? Could we write that instead if that is the case?

RasmusKjeldgaard (commented):
Solid work, and very nice finding with the existing CDK Portal. 
When tests are added I think we are golden!

RasmusKjeldgaard (approved):
Looks good!

PR #2917, opened 2023-03-06
Add floating directive

### LINKED ISSUE #2721: [COMPONENT] Multi select

NillerW:

This PR doesn't directly close any issues, but it is needed to close #2721 and #2744 down the road and provide general functionality for popup/floating components. 

The PR Adds a new and better way to handle floating/popup elements and is meant to replace the current popup component. 

This directive is meant for internal use, to create Kirby components such as Dropdown, Tooltip, Popup. etc. and not to be exposed outside the Kirby design system.

## What is the new behavior?

This PR introduces the KirbyFloating directive. This directive can be added directly to HTML-tags to make the tag float relative to another tag. 

The KirbyFloating directive implementation provides the following benefits:
- Ease-of-use: It's simply to attach two attributes to an existing element.
- Reference element and floating element does not have to be placed directly next to each other.
- Not introducing an unnecessary component tag to make the actual content float.
- Trigger events to toggle visibility of the floating element, like click and hover, are handled automatically by the directive.
- Automatic handling of expand direction, based on available size.
- Automatic positioning of floating content based on available size. 

With below image as reference, it can be seen how the new directive works. 

image

Required attributes:
- KirbyFloating: Provides the functionality.
- Reference: Element for which the floating content should be placed next to.

Optional attributes:
- Triggers: How the floating element should be set to show/hide. "Click" will attach a click event to the reference element, which toggles display of the floating element. Supports single value and array.
- Placement: How the element should be placed. Example top, down (under reference element centered), start-down (Under reference element aligned with start of element) etc.

Further inputs to the floating directive is likely to come, based on use-cases and requirements.

- [ ] Yes
- [X] No

## Checklist:

### Review  


apps/cookbook/src/app/examples/dropdown-example/examples/popover.ts
```diff
@@ -0,0 +1,20 @@
+import { Component, Input } from '@angular/core';
```
RasmusKjeldgaard: Seems unrelated, consider removing as the dropdown popover functionality should be replaced with this floating utility altogether pretty soon.

apps/cookbook/src/app/showcase/dropdown-showcase/dropdown-showcase.component.html
```diff
@@ -12,6 +12,15 @@ <h3>Default</h3>
...
 </cookbook-example-viewer>
+<h3>Popup</h3>
+<cookbook-example-viewer [html]="popupExample.template">
+  <div>
+    <cookbook-dropdown-example-popover
+      [size]="size"
+      #popupExample
+    ></cookbook-dropdown-example-popover>
+  </div>
+</cookbook-example-viewer>
```
RasmusKjeldgaard: Seems unrelated, consider removing as the dropdown popover functionality should be replaced with this floating utility altogether pretty soon.

libs/designsystem/shared/floating/src/floating.directive.ts
```diff
@@ -0,0 +1,368 @@
...
+
+/**
+ * @summary FloatingDirective is a utility that lets you declarative anchor "popup" containers to another element.
+ *
+ * Uses floating-ui, with this directive wraps the functionality: https://floating-ui.com/docs/getting-started
+ *
+ * Uses PortalDirective to enable functionality for portaling the floated content. This should be used when needed
+ * and not as the default option.
+ *
+ * @status In development
```
RasmusKjeldgaard: Around here, everything is in development 😛 Consider removing.

libs/designsystem/shared/floating/src/floating.directive.ts
```diff
@@ -0,0 +1,368 @@
...
+import { from } from 'rxjs';
+import { PortalDirective } from '@kirbydesign/designsystem/shared/portal';
+
+export type TriggerEvent = 'hover' | 'click' | 'focus';
+
+export enum FloatingOffset {
+  none = 0,
+  small = 5,
+  medium = 10,
+  large = 15,
```
RasmusKjeldgaard: Have these values been aligned with designers?
To me it looks like our standard offset for the dropdown is currently 8px, but that needs to be verified.

libs/designsystem/shared/floating/src/floating.directive.spec.ts
```diff
@@ -0,0 +1,463 @@
...
+  @ViewChild(FloatingDirective, { static: true }) public floatingDirective: FloatingDirective;
+
+  /** Start values set to match directive */
+  @Input() public isDisabled: boolean = false;
+  @Input() public closeOnEscapeKey: boolean = true;
+  @Input() public closeOnSelect: boolean = true;
+  @Input() public closeOnBackdrop: boolean = true;
+  @Input() public strategy: Strategy = 'absolute';
+  @Input() public placement: Placement = 'bottom-start';
+}
```
RasmusKjeldgaard, line 45: Cant we just not initialise them, so is uses the default values from the directive, and then you set the input when needed?

libs/designsystem/shared/floating/src/floating.directive.spec.ts
```diff
@@ -0,0 +1,463 @@
...
+  it('should be created', () => {
+    expect(spectator.component).toBeTruthy();
+  });
+
+  describe('input', () => {
+    describe('reference', () => {
+      it('should set', () => {
+        directive.reference = component.floatingElementRef;
+        expect(directive.reference).toEqual(component.floatingElementRef);
+      });
```
RasmusKjeldgaard, line 73: This test seems redundant, as we forcibly assign the value on the first line, before checking what we just did. 

What do we actually want to test here?

RasmusKjeldgaard: Unless there is a really good reason, I would not test the getter/setter functionality of Angular. The code for setting the local variable inside the component is very simple, and it is hard to test anything meaningful. 

This goes for all the 'should set' tests in my opinion.

libs/designsystem/shared/floating/src/floating.directive.ts
```diff
@@ -0,0 +1,368 @@
...
+}
+
+interface EventMethods {
+  event: string;
+  method: () => void;
+}
+
+type EventListener = () => void;
```
RasmusKjeldgaard: The Typescript dom lib has a similar type, so you should be able to remove this line.

libs/designsystem/shared/floating/src/floating.directive.ts
```diff
@@ -0,0 +1,368 @@
...
+    }
+  }
+
+  private tearDownEventHandling(): void {
+    this.eventListeners.forEach((eventListener) => {
+      if (eventListener != null) {
+        eventListener();
+      }
+    });
+    this.eventListeners = [];
```
RasmusKjeldgaard: For performance reasons, to not risk collecting a lot of detached elements, shouldnt we actually remove any attached event-listeners here instead?

libs/designsystem/shared/floating/src/floating.directive.spec.ts
```diff
@@ -0,0 +1,463 @@
...
+        autoUpdateFuncSpy = jasmine.createSpy('autoUpdate').and.returnValue(null);
+        spyOnProperty(floatingUi, 'autoUpdate', 'get').and.returnValue(autoUpdateFuncSpy);
+
+        computePositionFuncSpy = jasmine
+          .createSpy('computePosition')
+          .and.returnValue(new Promise(null));
+        spyOnProperty(floatingUi, 'computePosition', 'get').and.returnValue(computePositionFuncSpy);
+      });
+
+      describe('add styling', () => {
```
RasmusKjeldgaard: I believe this describe is redundant, and could be removed for readablility so it reads:
`FloatingDirective method ngOnInit should add style XYZ` instead of `FloatingDirective method ngOnInit add style should add style XYZ`

RasmusKjeldgaard (changes requested):
I think this looks great! Only some small comments that should not take long to resolve.

We need to see if we can clean up all the unrelated ng-package changes (consider checking each one out from develop and removing the pre-commit hook in package.json on that commit).

libs/designsystem/shared/floating/src/floating.directive.ts
NillerW: I'll check in with Henrik :)

libs/designsystem/shared/floating/src/floating.directive.spec.ts
NillerW: Since I bind them to the directive, setting them to undefined would also give that value to the directive, thus not being able to use the default values. 

I need these inputs to be able to simulate actually changing the values from the component using the directive :)

libs/designsystem/shared/floating/src/floating.directive.spec.ts
NillerW: Generally I agree, I just thought for a component library, it seemed reasonable to be sure. 

Especially for reference, since the code is more than a simply setter:

`@Input() public set reference(ref: ElementRef) {
    this.tearDownEventHandling();
    this._reference = ref as ElementRef;
    this.setupEventHandling();
    this.autoUpdatePosition();
  }`
  
  So I'm making sure that if someone changes the affected code, we still get the right reference set.

libs/designsystem/shared/floating/src/floating.directive.ts
NillerW: Can you elaborate. The if statement should not be necessary, but just in-case so we are sure to be able to execute the function

libs/designsystem/shared/floating/src/floating.directive.spec.ts
RasmusKjeldgaard: I see 👍

libs/designsystem/shared/floating/src/floating.directive.ts
RasmusKjeldgaard: It dawns on me that I misread this, I was not aware that this utilizes the teardown function returned from renderer. 

But that hightlights the actual problem here, I believe: The naming is a bit misleading, because the value returned and pushed to the eventListeners array is not actually an event listener, but rather an `EventListenerDisposeFunction` or whatever we might call that - so my comment from before about removing the EventListener type is actually wrong, it should be replaced with something like that instead. And then we should rename the array to reflect that.

libs/designsystem/shared/floating/src/floating.directive.spec.ts
RasmusKjeldgaard: Okay, fair point!

RasmusKjeldgaard (approved):
🎉

PR #3442, opened 2024-04-15
Progress circle: fix wrong animation direction

### LINKED ISSUE #3441: [BUG] Progress Circle load animation is wrong direction

## Describe the bug
When the progress circle is visible on page load, the animation goes from 100% progress backwards towards the value set by consumers, instead of building up from 0 towards the set value.

This does not happen when the progress circle is below the fold, and the animation is triggered by the intersection observer once the progress circle is scrolled into view. 

1. Go to https://cookbook.kirby.design/#/home/showcase/progress-circle 
2. See that the progress circles immediately visible loads counter-clockwise from full to value
3. Scroll down to see progress circles previously hidden outside the viewport animating correctly, clockwise from 0 to value

9.4.1

Progress circle should animate clockwise from 0 to the set value on page load and when scrolled into view.

Here are the offset and centerCircumference values when it loads the wrong way around
image

<hr />

## Checklist:

### Verification

### Implementation 

### Review

- [ ] Create a pull-request.

RasmusKjeldgaard:
This PR closes #3441

## What is the new behavior?
The issue described here https://github.com/kirbydesign/designsystem/issues/3441#issuecomment-2056620529 happens because the colored part of the circle is rendered before the attribute bindings for `stroke-dashoffset` and `stroke-dasharray` have been evaluated, so the full circle is shown.

A view-initialized class is now added when the components view is rendered, and we then wait until this class is added before we apply the transition.

Additionally, instead of using stroke-dashoffset I have refactored the stroke to use only the dasharray attribute to set a stroke and gap length. 
One downside of this approach is that even with a stroke of length 0, we end up applying a stroke width, so we always draw a little dot when the value is 0. 
image
To work around this I have added a conditional that sets the stroke width to 0, when value is 0.

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  

### DIFF
--- a/libs/designsystem/progress-circle/src/progress-circle-ring.component.scss
+++ b/libs/designsystem/progress-circle/src/progress-circle-ring.component.scss
@@ -8,19 +8,6 @@ svg {
   stroke: utils.get-color('semi-light');
 }
 
-.progress {
-  // Avoid flash of full progress circle if attribute binding in template has not yet been evaluated.
-  &:not([stroke-dasharray]) {
-    stroke-dasharray: 0 999;
-  }
-
-  transition-property: stroke-dasharray;
-  transition-duration: utils.get-transition-duration('extra-long');
-  transition-timing-function: utils.get-transition-easing('enter-exit');
-  transform-origin: 50% 50%;
-  stroke: var(--kirby-progress-circle-stroke-color, #{utils.get-color('success')});
-}
-
 :host {
   @each $color-name, $color-value in utils.$notification-colors {
     &.#{$color-name} {
@@ -31,6 +18,10 @@ svg {
   &.view-initialized {
     .progress {
       transition-property: stroke-dasharray, stroke;
+      transition-duration: utils.get-transition-duration('extra-long');
+      transition-timing-function: utils.get-transition-easing('enter-exit');
+      transform-origin: 50% 50%;
+      stroke: var(--kirby-progress-circle-stroke-color, #{utils.get-color('success')});
     }
   }
 }
--- a/libs/designsystem/progress-circle/src/progress-circle-ring.component.scss
+++ b/libs/designsystem/progress-circle/src/progress-circle-ring.component.scss
@@ -9,20 +9,23 @@ svg {
 }
 
 .progress {
-  transition-property: stroke-dasharray, stroke;
+  transition-property: stroke-dasharray;
   transition-duration: utils.get-transition-duration('extra-long');
   transition-timing-function: utils.get-transition-easing('enter-exit');
   transform-origin: 50% 50%;
   stroke: var(--kirby-progress-circle-stroke-color, #{utils.get-color('success')});
-  // Avoid flash of full progress circle if attribute binding in template has not yet been evaluated.
-  &:not([stroke-dasharray]) {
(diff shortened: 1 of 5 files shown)


libs/designsystem/progress-circle/src/progress-circle-ring.component.ts
```diff
@@ -39,7 +30,18 @@ export class ProgressCircleRingComponent {
...
+
+    return (_value * this._centerCircumference) / 100;
+  }
+
+  get _remainder(): number {
+    return this._centerCircumference - this._progress;
+  }
+
+  get _progressStrokeWidth(): number {
+    return this.value == 0 ? 0 : this.strokeWidth;
```
jakobe: Maybe add a comment as to why this is necessary - and/or implement with a guard to make the intent clearer:
```suggestion
    // Don't render stroke if value is 0, otherwise it will show as a dot:
    if (this.value === 0) return 0;
    return this.strokeWidth;
```

libs/designsystem/progress-circle/src/progress-circle-ring.component.ts
```diff
@@ -39,7 +30,18 @@ export class ProgressCircleRingComponent {
     return this._centerRadius * 2 * Math.PI;
   }
 
-  private calculateOffset(value: number): number {
-    return this._centerCircumference - this._centerCircumference * (value / 100);
+  get _progress(): number {
+    const valueWithinBounds = this.value < this.upperBound || this.value > 99;
+    const _value = valueWithinBounds ? this.value : this.upperBound;
+
+    return (_value * this._centerCircumference) / 100;
```
jakobe: Consider readability, ie. as in the original `calculateOffset` method - the progress is a percentage (value out of 100) times the circumference 
```suggestion
    return this._centerCircumference * (_value / 100);
```
OR even more explicit:
```suggestion
    const progressPercentage = _value / 100;
    return this._centerCircumference * progressPercentage;
```

libs/designsystem/progress-circle/src/progress-circle-ring.component.scss
```diff
@@ -9,13 +9,18 @@ svg {
...
   transition-timing-function: utils.get-transition-easing('enter-exit');
   transform-origin: 50% 50%;
   stroke: var(--kirby-progress-circle-stroke-color, #{utils.get-color('success')});
 }
 
+// Avoid flash of full progress circle if attribute binding in template has not yet been evaluated.
+.progress:not([stroke-dasharray]) {
+  stroke-dasharray: 0 999;
+}
+
```
jakobe: Consider grouping with rest of `.progress` styles:
```suggestion
  stroke: var(--kirby-progress-circle-stroke-color, #{utils.get-color('success')});
  // Avoid flash of full progress circle if attribute binding in template has not yet been evaluated.
  &:not([stroke-dasharray]) {
    stroke-dasharray: 0 999;
  }
}
```

jakobe (commented):
Looks really grreat! Just a few suggestions for minor improvements, mainly around redability 👌🏻

jakobe (dismissed):

jakobe (approved):

PR #3455, opened 2024-04-25
Fix issue when opening modals in quick succession

andershessellund:
This is an issue during automated testing.


RasmusKjeldgaard:
Further info: 
Modals can not be opened in quick succession in automated testing scenarios even when animations are disabled. 
This is an issue when opening a modal, and then quickly opening another nested modal - in this scenario the nested modal will not be opened unless delays are added to tests.

RasmusKjeldgaard:
Closing this, as the proposal does not fix the issue.

PR #3476, opened 2024-05-17
Release v9.5.0

RasmusKjeldgaard:
[Changelog ](https://github.com/kirbydesign/designsystem/releases/tag/untagged-537f36d54bee54d3ca5e)

### DIFF
--- a/libs/core/src/scss/base/_functions.scss
+++ b/libs/core/src/scss/base/_functions.scss
@@ -226,7 +226,7 @@
 @mixin trailing-icon($svg: 'link') {
   $icon-scaling-factor: 1.5em;
 
-  @if str-index($svg, 'http') {
+  @if string.index($svg, 'http') {
     // External icon
     background-image: url('#{$svg}');
   } @else {
--- a/libs/designsystem/accordion/src/accordion-item.component.scss
+++ b/libs/designsystem/accordion/src/accordion-item.component.scss
@@ -41,7 +41,7 @@ $padding: utils.size('s');
 }
 
 .content-body {
-  padding: 0 $padding $padding $padding;
+  padding: 0 $padding $padding;
 }
 
 .disabled {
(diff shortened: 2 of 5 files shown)


libs/core/src/scss/base/_functions.scss
```diff
@@ -226,7 +226,7 @@
 @mixin trailing-icon($svg: 'link') {
   $icon-scaling-factor: 1.5em;
 
-  @if str-index($svg, 'http') {
+  @if string.index($svg, 'http') {
```
RasmusKjeldgaard, line 229: The package-lock file was updated after bumping, and with it I started to see stylelint errors. 
I can see from the package-lock file that our stylelint dep was bumped from 16.3.1 -> 16.5.0, so that might be why these two errors suddenly emerged? 🤷

libs/designsystem/accordion/src/accordion-item.component.scss
```diff
@@ -41,7 +41,7 @@ $padding: utils.size('s');
 }
 
 .content-body {
-  padding: 0 $padding $padding $padding;
+  padding: 0 $padding $padding;
```
RasmusKjeldgaard, line 44: This happens due to stylelint being bumped from 16.3.1 -> 16.5.0 under the hood.
From the stylelint changelog: Fixed: shorthand-property-no-redundant-values false negatives for functions (https://github.com/stylelint/stylelint/pull/7657).

libs/core/src/scss/base/_functions.scss
RasmusKjeldgaard: Yep, thats it. This happens due to stylelint/scss being bumped from 6.2.1 -> 6.3.0 under the hood.

From the stylelint/scss changelog: Fixed: no-global-function-names report errors inside at-rules (https://github.com/stylelint-scss/stylelint-scss/pull/988).

jakobe (approved):

PR #3478, opened 2024-05-23
Docs: update accordion animation note

RasmusKjeldgaard:
A quick docs clarification as there are now two ways to register the browser animations in angular, depending on wether a project is module-based or standalone.


jakobe (approved):
👌🏻

PR #3504, opened 2024-06-19
Repo: Set up core path alias and fix publish script paths

RasmusKjeldgaard:
No issue, this is a quick fix. 
I forgot to change the paths for our publish script, now that we place dist files inside the individual project folders.

- [ ] Yes
- [x] No

### DIFF
--- a/scripts/publish.js
+++ b/scripts/publish.js
@@ -35,7 +35,7 @@ const coreLibDir = `${libsRootDir}/core`;
 const coreLibSrcDir = `${coreLibDir}/src`;
 
 const dist = `dist`;
-const distDesignsystemTarget = `${dist}/${designsystemLibDir}`;
+const distDesignsystemTarget = `${designsystemLibDir}/${dist}`;
 const distDesignsystemPackageJsonPath = `${distDesignsystemTarget}/package.json`;
 const distCoreTarget = `${dist}/${coreLibDir}`;
 const distCorePackageJsonPath = `${distCoreTarget}/package.json`;


libs/core/package.json
```diff
@@ -3,9 +3,9 @@
   "version": "0.0.61",
   "description": "Core styles and helpers for the Kirby Design System.",
   "type": "module",
-  "main": "dist/index.js",
-  "module": "dist/index.js",
-  "types": "dist/types/index.d.ts",
```
RasmusKjeldgaard, line 8: We can simplify this, as there is no reason to have a dist folder inside our published package. In our publish script we copy the package.json to the same level as our index.js files already, so there is no need to nest things inside a dist folder. Now it just looks like this when installed:
image

libs/core/tsconfig.json
```diff
@@ -2,11 +2,10 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "outDir": "./dist",
-    "rootDir": "src",
```
RasmusKjeldgaard, line 5: With this change, the `package.json` (placed in the root folder outside `src`) can be included and `tsc` will copy it to`dist` when building. This is much like how ng-packagr does for our angular packages, which means we can now reuse our package alias strategy and point to the core dist folder during local development. This streamlines the wait-on logic for parallel building too.

libs/designsystem/tsconfig.spec.json
```diff
@@ -1,63 +1,64 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
-    "baseUrl": ".",
-    "rootDir": ".",
```
RasmusKjeldgaard, line 5: To allow setting up an alias for core and include the core files in the designsystem test compilation, we need to remove this to change the rootDir to the base directory of the repo. Otherwise the core files will be "outside the root dir specified for compilation".

RasmusKjeldgaard: This in turn affects all the path aliases, but as an added benefit they now have an identical structure to the base.tsconfig.json, and maybe we can do a similar transformation as we do for the jest library down the line, to get rid of these paths?

scripts/publish.js
```diff
@@ -195,9 +187,7 @@ if (doPublishCore) {
   console.log('--- Publishing core ---');
   cleanDistribution(distCoreTarget)
     .then(() => buildPackage('core'))
-    .then(() => copyCoreDistributionFiles(coreLibDir, distCoreTarget))
```
RasmusKjeldgaard, line 198: The files are now published directly from their location in `libs/dist/core`, so no need to copy them anywhere.

scripts/publish.js
```diff
@@ -195,9 +182,7 @@ if (doPublishCore) {
...
     .then(() => copyScssFiles(coreLibSrcDir, distCoreTarget))
-    .then(() => copyPackageJson(coreLibDir, distCorePackageJsonPath))
```
RasmusKjeldgaard, line 200: Now handled by the typescript compiler

jakobe (approved):

PR #3508, opened 2024-06-21
Modal footer: Apply correct theme variables for nested elements inside footer

Fuzzy3:

This PR fixes an issue where grey buttons (buttons with `attentionLevel="3"`) in an inline footer would not change their color to white, when modal footer has `type="inline"`
An example code snippet of the use would look like this:
```
<kirby-modal-footer type="inline">
  <button kirby-button attentionLevel="3">Click Me</button>
</kirby-modal-footer>
```
And the output of this snippet would show a grey button on a grey background when a modal footer is set as inline, which makes the button not stand out:

image

## What is the new behavior?

When the `type="inline"` is added to a modal footer, the correct theme (light-theme) will be applied causing the grey buttons within the modal footer to properly turn white. 

The output will now look like this:

image

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  

### DIFF
--- a/libs/designsystem/modal/src/modal/footer/modal-footer.component.scss
+++ b/libs/designsystem/modal/src/modal/footer/modal-footer.component.scss
@@ -43,6 +43,7 @@ ion-footer {
 :host(.inline) ion-footer {
   background: transparent;
   box-shadow: none;
+  @include themes.apply-light-theme;
 }
 
 :host-context(.keyboard-visible) {
--- a/libs/designsystem/modal/src/modal/footer/modal-footer.component.scss
+++ b/libs/designsystem/modal/src/modal/footer/modal-footer.component.scss
@@ -37,13 +37,16 @@ ion-footer {
 }
 
 :host(.light) ion-footer {
+  @include themes.apply-light-theme;
+
   background-color: utils.get-color('background-color');
 }
 
 :host(.inline) ion-footer {
+  @include themes.apply-light-theme;
+
   background: transparent;
   box-shadow: none;
-  @include themes.apply-light-theme;
 }
 
 :host-context(.keyboard-visible) {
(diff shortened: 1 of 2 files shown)


libs/designsystem/modal/src/modal/footer/modal-footer.component.scss
```diff
@@ -43,6 +43,7 @@ ion-footer {
 :host(.inline) ion-footer {
   background: transparent;
   box-shadow: none;
+  @include themes.apply-light-theme;
```
jkaltoft: Just a FYI-comment 😊:

A common best practice using Sass is to apply mixins before "local" declarations. It can make it easier to spot if some properties are overwritten. 

However, in this case it doesn't matter, since `themes.apply-light-theme` doesn't apply declarations for `background` or `box-shadow`

RasmusKjeldgaard, line 46: This only fixes it for the inline modal but would be good to fix it for the `light` themeColor footer, by also applying the mixin to the `:host(.light)` selector 👍

RasmusKjeldgaard: This only fixes it for the inline modal but would be good to fix it for the `light` themeColor footer, by also applying the mixin to the `:host(.light)` selector 👍

RasmusKjeldgaard (changes requested):
I think we should either add an issue and reference it in this PR, or put a description of the bug with some screenshots in the PR description for future reference.

Remember to also add a fitting label to the PR by following the checklist in the PR description.

At last it would be good to change the PR title to be a bit more descriptive, as it will end up on the changelog. Maybe something like: **Modal footer: apply correct theme variables for nested elements inside footer**

libs/designsystem/modal/src/modal/footer/modal-footer.component.scss
Fuzzy3: Makes great sense ! I have rearranged

Fuzzy3: Aha of course ! done now.

RasmusKjeldgaard (approved):
Great stuff! 🎉

PR #3566, opened 2024-09-02
Upgrade to Angular 18

### LINKED ISSUE #3547: [Housekeeping] Angular 18 upgrade

We should upgrade all our Angular dependencies to v18. 
Buildable version was made here, but som tests are failing, and any changes have not been thoroughly considered: https://github.com/kirbydesign/designsystem/pull/3503.
In addition, Ionic upgrade (#3548) should be done before this to be fully compatible with Angular 18.

Fuzzy3:

This PR closes #3547 

## What is the new behavior?

This will upgrade our repo to angular 18, including other dependency version like storybook, nx typescript and more. 

- [x] Yes
- [ ] No

## Checklist:

### Reminders

### Review  


libs/designsystem/form-field/src/input/input.component.stories.ts
```diff
@@ -30,7 +30,7 @@ export const Input: Story = {
     inputmode: '',
     placeholder: '',
   },
-  render: (args: InputComponent) => ({
+  render: (args) => ({
```
RasmusKjeldgaard: Initially I was a bit confused by this change, but tracked down [this storybook issue](https://github.com/storybookjs/storybook/issues/26278#issuecomment-1999383800), that seems to argue that type annotations should generally be removed as storybook **_might_** do some internal mapping of the types. 

So, I guess it would be better to be consistent here, and do it for all our stories, not just the ones failing at the moment. 

But seeing as this would clutter the PR quite a bit, I would argue that we should create a pre-burner for this with the storybook 8 changes to keep the scope nicely separated.

RasmusKjeldgaard: Would be good to first verify that the Storybook 8 upgrade is at all necessary for the Angular 18 upgrade.

package.json
```diff
@@ -33,18 +33,18 @@
...
+    "@nx/angular": "19.6.5",
+    "@nx/jest": "19.6.5",
+    "@nx/js": "19.6.5",
+    "@nx/storybook": "19.6.5",
+    "@nx/web": "19.6.5",
+    "@nx/workspace": "19.6.5",
+    "@storybook/addon-essentials": "8.2.9",
+    "@storybook/addon-interactions": "8.2.9",
+    "@storybook/angular": "8.2.9",
+    "@storybook/core-server": "8.2.9",
```
RasmusKjeldgaard: I think it would be good to add ^ to these to automatically get new improvements and bugfixes, at least for dev dependencies like nx and storybook with a good reputation for adhering to semver.

RasmusKjeldgaard: We should specifically not add it for the typescript dependency down below, as typescript does not follow semver: https://github.com/microsoft/TypeScript/issues/14116

RasmusKjeldgaard (changes requested):
Great work, seems very close to done!

libs/designsystem/form-field/src/input/input.component.stories.ts
Fuzzy3: image
These are the peerDeps of the @storybook/angular version 7.6.13 that we currently use - and it looks like it only supports up to angular version 18 (not including that version) - so I believe we have to upgrade storybook

RasmusKjeldgaard: I see, then I think we should do the storybook upgrade in a separate PR before we merge this one, and remove the args type annotation throughout 👍 

Then we can focus on storybook first, and fix any issues. For instance, I can see that `VisualTestingReadme.mdx` is currently failing to load inside storybook because of the upgrade.

RasmusKjeldgaard (approved):

PR #3601, opened 2024-09-12
Icon: Customizable icon asset path

andershessellund:
This adds the possibility of customizing the asset path used to fetch icons. The default remains unchanged `assets/kirby/icons/svg` .

The path is customized by providing an injection token `BUILT_IN_ICONS_URL_TOKEN` with the path to the icons.


libs/designsystem/icon/src/icon-registry.service.ts
```diff
@@ -32,7 +40,12 @@ export class IconRegistryService {
   }
 
   private addDefaultIcons(): void {
-    this.addIcons(kirbyIconSettings.icons);
+    const baseUrl = this.builtInIconsUrl ?? DEFAULT_BUILT_IN_ICONS_URL;
```
Fuzzy3, line 44: Instead of this, you can provide a "factory"-method as an option to the injection token, which will return a default value if not overridden. Check this stackblitz as an example: https://stackblitz.com/edit/angular-default-value-token?file=src%2Fapp%2Fapp.component.ts

Fuzzy3: Then you can also get rid of the "optional: true" in the inject itself

libs/designsystem/icon/src/icon-registry.service.ts
```diff
@@ -32,7 +40,12 @@ export class IconRegistryService {
...
+    const baseUrlWithSlash = baseUrl.endsWith('/') ? baseUrl : baseUrl + '/';
+    for (const { name, svg } of kirbyIconSettings.icons) {
```
Fuzzy3:
```suggestion
    kirbyIconSettings.icons.forEach((icon) => {
```

libs/designsystem/icon/src/icon-registry.service.ts
```diff
@@ -1,13 +1,21 @@
...
 
 import { Icon } from './icon-settings';
 import { kirbyIconSettings } from './kirby-icon-settings';
 
+/**
+ * Token used to configure base URL for path to built-in icons. Default is 'assets/kirby/icons/svg'.
+ */
+export const BUILT_IN_ICONS_URL = new InjectionToken<string>('BUILT_IN_ICONS_URL');
+
+export const DEFAULT_BUILT_IN_ICONS_URL = 'assets/kirby/icons/svg';
```
Fuzzy3:
```suggestion
export const DEFAULT_BUILT_IN_ICONS_URL = 'assets/kirby/icons/svg/';
```

libs/designsystem/icon/src/icon-registry.service.ts
```diff
@@ -1,13 +1,21 @@
...
+
 @Injectable({
   providedIn: 'root',
 })
 export class IconRegistryService {
   private iconRegistry = new Map<string, string>();
+  private builtInIconsUrl = inject(BUILT_IN_ICONS_URL, { optional: true });
```
Fuzzy3: Move this to the constructor, then its similar to what we do other places in Kirby 😄  I know it doesn't change anything, but its just nice to have to the same way.

Fuzzy3 (changes requested):
Really nice PR ! Just a few changes needed in the code, and then its good to go. 
Besides this it would also be nice if you could update the Icon page in the cookbook and also the README on how to override this icon-path. 
For inspiration we just made a short "setup" section for the DataTable component, maybe this could be written somewhat similarly. [https://cookbook.kirby.design/branch/kby-develop/#/home/showcase/data-table](https://cookbook.kirby.design/branch/kby-develop/#/home/showcase/data-table)

libs/designsystem/icon/src/icon-registry.service.ts
```diff
@@ -32,7 +40,12 @@ export class IconRegistryService {
   }
 
   private addDefaultIcons(): void {
-    this.addIcons(kirbyIconSettings.icons);
+    const baseUrl = this.builtInIconsUrl ?? DEFAULT_BUILT_IN_ICONS_URL;
+    const baseUrlWithSlash = baseUrl.endsWith('/') ? baseUrl : baseUrl + '/';
+    for (const { name, svg } of kirbyIconSettings.icons) {
+      const url = baseUrlWithSlash + svg;
```
Fuzzy3, line 47:
```suggestion
      const url = baseUrlWithSlash + icon.svg;
```

libs/designsystem/icon/src/icon-registry.service.ts
```diff
@@ -32,7 +40,12 @@ export class IconRegistryService {
...
+      this.addIcon(name, url);
```
Fuzzy3, line 48:
```suggestion
      this.addIcon(icon.name, url);
```

libs/designsystem/icon/src/icon-registry.service.ts
```diff
@@ -32,7 +40,12 @@ export class IconRegistryService {
...
+    }
```
Fuzzy3:
```suggestion
    });
```

libs/designsystem/icon/src/icon-registry.service.ts
andershessellund: IMO constructor injection is the legacy way to do it, but it is entirely up to you.

apps/cookbook/src/app/showcase/icon-showcase/icon-showcase.component.html
```diff
@@ -21,6 +21,26 @@ <h2>Sizes</h2>
...
+      explained in our readme file
+    </a>
+  </p>
+  <p>
+    If the icons exists at another location, it is possible to configure the path that the icons are
+    loaded from by providing the
+    <code>BUILT_IN_ICONS_URL</code>
+    injection token.
+  </p>
+
```
jakobe, line 44: Consider adding that the `BUILT_IN_ICONS_URL` token can be set both to a relative and an absolute path.

jakobe, line 44: Consider adding an inline code snippet that shows how to add the token to app providers

jakobe (approved):
👌🏻

Fuzzy3 (approved):

PR #3644, opened 2024-09-27
Icons: Sort by name

jkaltoft:

This PR closes NO ISSUE

## What is the new behavior?

Sort array of icons by name (alphabetically) to improve cookbook Icon showcase.

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  

### DIFF
--- a/libs/designsystem/icon/src/kirby-icon-settings.ts
+++ b/libs/designsystem/icon/src/kirby-icon-settings.ts
@@ -1,8 +1,13 @@
-import { IconSettings } from './icon-settings';
+import { Icon, IconSettings } from './icon-settings';
+
+const sortIconsByName = (icons: Icon[]) => {
+  const collator = new Intl.Collator();
+  return icons.sort((a, b) => collator.compare(a.name, b.name));
+};
 
 /* eslint-disable */
 export const kirbyIconSettings: IconSettings = {
-  icons: [
+  icons: sortIconsByName([
     { name: 'accounts-outline', svg: 'accounts-outline.svg' },
     { name: 'accounts', svg: 'accounts.svg' },
     { name: 'add', svg: 'add.svg' },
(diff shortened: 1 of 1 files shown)


libs/designsystem/icon/src/kirby-icon-settings.ts
```diff
@@ -1,8 +1,13 @@
-import { IconSettings } from './icon-settings';
+import { Icon, IconSettings } from './icon-settings';
+
+const sortIconsByName = (icons: Icon[]) => {
+  const collator = new Intl.Collator();
+  return icons.sort((a, b) => collator.compare(a.name, b.name));
```
jakobe: Wondering about the use of `Intl.Collator`: all our icons are named in English and we don't construct the collator with any locales, so I guess it always runs the comparison in English anyhoo? 🤔

Consider simplifiying to `icons.sort()`

jkaltoft: I don't think we can use just the `.sort()` without providing a compare function, because `icons` is an array of *objects*, not numeric values.

At least we need to specify which properties the `sort()` method should compare, e.g.:

```
.sort((a, b) => a.name - b.name))
```

but that doesn't seem to work.

The best/safest way to sort an array of objects by a string property I could find is to use `localeCompare()` 

```
.sort((a, b) => a.name.localeCompare(b.name))
```

MDN states that 

> In implementations with [Intl.Collator API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) support, this method simply calls Intl.Collator

We could change it to use `localeCompare()` since that must be equivalent to the current implementation using `Intl.Collator`

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

jkaltoft: Another option is to use the "sort by name" example from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#sorting_array_of_objects

It's more verbose, but quite straight-forward

```
// sort by name
items.sort((a, b) => {
  const nameA = a.name.toUpperCase(); // ignore upper and lowercase
  const nameB = b.name.toUpperCase(); // ignore upper and lowercase
  if (nameA < nameB) {
    return -1;
  }
  if (nameA > nameB) {
    return 1;
  }

  // names must be equal
  return 0;
});
```

jakobe (approved):

PR #3689, opened 2024-10-25
Calendar: enable keyboard interaction and improve semantics

### LINKED ISSUE #3681: [Enhancement] A11y improvements for Calendar

## Describe the enhancement
Based on the [date picker dialog pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/), but without the input field functionality, the following improvements should be considered to improve the accessibility of the calendar: 

- [x] Keyboard support for moving around the grid of dates - as detailed under **_Date Picker Dialog: Date Grid_** in the above link.
- [x] (Probably optional) Shortcut keys for changing the month and year displayed in the calendar.
- [x] The calendar heading displaying the month and year should be marked as a live region so screen reader users get feedback from the buttons and keyboard commands that change the month and year.
- [ ] Possibly keyboard help to convey arrow-key functionality when moving aroind the grid. A live region can be used to announce it to screen reader users when focus moves into the calendar grid.
- [x] Day names and column headers are abbreviated to single characters in the design. However, this makes it more difficult for screen reader users to understand the day names. So, full day names should be provided to assistive technologies in the HTML abbr attribute on the column headers, enabling screen readers to announce the full names when users navigate the grid. 

- Some calendars, like [ion-datetime](https://ionicframework.com/docs/api/datetime#basic-usage) use a grid of buttons. 
- [Cally](https://wicky.nillia.ms/cally/) use a table, but with the addition of buttons for each selectable day. This approach is outlined here: https://wicky.nillia.ms/cally/accessibility/#approach

<hr />

## Checklist:

### Refinement

- [ ] Request that the issue is [UX refined](https://github.com/kirbydesign/designsystem/blob/develop/.github/CONTRIBUTING.md#ux-refinement)
- [ ] Request that the issue is [tech refined](https://github.com/kirbydesign/designsystem/blob/develop/.github/CONTRIBUTING.md#tech-refinement); do not proceed until this is done.

### Implementation 

### Review

RasmusKjeldgaard:
This PR closes #3681

### Semantics
The existing table structure is reused, to allow screen reader users to navigate the calendar using standard table navigation concepts.

Most of these concepts are borrowed from the [WAI ARIA date picker dialog pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/), but without the date input field specifics.

- The table of dates has a grid role, and an aria-describedby attribute that links it with the date and year in the header.
The date and year is marked as a live-region, so that screen readers will announce when the month or year is changed.
- A full name of the weekday is visually hidden in each table header, to be announced by screen readers, while the single letter shorthand shown on screen is hidden for screen readers.
- Weekday names and shorthands are translated according to locale to be announced correctly
- When selected the cell is marked with `aria-selected` attribute to be announced as the selected option.

### Keyboard support
Again based on [WAI ARIA date picker dialog pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/):

| Key Combination     | Action                                                                                                                                     |
|---------------------|--------------------------------------------------------------------------------------------------------------------------------------------|
| Enter               | Select the date, ~~close the dialog, and move focus to the "Choose Date" button. Update the value of the "Date" input with the selected date. Update the accessible name of the "Choose Date" button to include the selected date.~~ |
| Up Arrow            | Moves focus to the same day of the previous week.                                                                                           |
| Down Arrow          | Moves focus to the same day of the next week.                                                                                               |
| Right Arrow         | Moves focus to the next day.                                                                                                                |
| Left Arrow          | Moves focus to the previous day.                                                                                                            |
| Home                | Moves focus to the first day (e.g Sunday) of the current week.                                                                              |
| End                 | Moves focus to the last day (e.g. Saturday) of the current week.                                                                            |
| Page Up             | Changes the grid of dates to the previous month. Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month. |
| Shift + Page Up     | Changes the grid of dates to the same month in the previous year. Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month. |
| Page Down           | Changes the grid of dates to the next month. Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month. |
| Shift + Page Down   | Changes the grid of dates to the same month in the next year. Moves focus to the day of the month that has the same number. If that day does not exist, moves focus to the last day of the month. |

### Other stuff
- To simplify adding keyboard support as much as possible I removed the `CalendarHelper` and related functionality as it seems to be unused by the current implementation of the calendar.

#### TODO
- [x] Find out how we want to provide translations - tracked in #3694 

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


libs/designsystem/calendar/src/calendar.component.html
```diff
@@ -24,7 +24,7 @@
       [noDecoration]="true"
       class="no-margin"
     >
-      <kirby-icon name="arrow-more"></kirby-icon>
+      <kirby-icon name="arrow-more" aria-label="Next month"></kirby-icon>
```
RasmusKjeldgaard: Opted to add the aria-label (that also needs translation!) on the icon, not the button, as I think that makes most sense. It is the icon we want to describe, as it is added in place of text content.

apps/cookbook/src/app/showcase/_showcase.shared.scss
```diff
@@ -12,3 +12,18 @@ cookbook-example-viewer,
...
+// Style the Keyboard Input element. Borrowed from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd
+kbd {
+  border-radius: 3px;
+  border: 1px solid var(--kirby-semi-dark);
+  box-shadow:
+    0 2px 1px var(--kirby-dark-overlay-20),
+    0 2px 1px 0 var(--kirby-white) inset;
+  color: var(--kirby-black);
+  font-size: 0.85em;
+  font-weight: 700;
```
jakobe: Consider using built in presets: 
```suggestion
  font-weight: var(--kirby-font-weight-bold);
```

libs/designsystem/calendar/src/calendar.component.html
jakobe: I can totally follow this argument - on the other hand, the illustration/icon is purely presentational and it's the button that performs the action, so you could also argue that the button should "own" the label 🤔 
(In practice it doesn't matter as the button will get it's accessible label from it's content either way...)
Another argument though could be, that the label describes what the button does and not what the icon depicts (arrow back/forward) and hence it should sit on the button.
[Shoelace btw puts it on the button and `aria-hidden` on the icon](https://shoelace.style/components/icon-button)

libs/designsystem/calendar/src/calendar.component.spec.ts
```diff
@@ -652,6 +640,213 @@ describe('CalendarComponent', () => {
...
+    });
+
+    it('should set focussed date to equal selectedDate when selectedDate is set', () => {
+      const selectedDate = localMidnightDate('1997-08-29');
+
+      spectator.setInput('selectedDate', selectedDate);
+      expect(spectator.component['focussedDate']).toEqual(selectedDate);
+    });
+
+    it('should set focussed date to previously selectedDate when selectedDate is set to null (deselected)', () => {
```
jakobe: Not sure I understand this requirement? 🤔

libs/designsystem/calendar/src/calendar.component.html
```diff
@@ -37,27 +37,35 @@
...
-          class="{{ day.cssClasses }} contain-state-layer"
+          (keydown)="_onDateKeydown($event)"
+          class="{{ day.cssClasses }}"
           [class.selected]="day.isSelected"
+          [class.focussed]="day.isFocussed"
+          [noDecoration]="true"
+          [tabIndex]="day.isFocussed ? 0 : -1"
+          [attr.aria-label]="day.ariaLabel"
+          [attr.aria-disabled]="day.isSelectable ? null : true"
+          [disabled]="day.isFocusable ? null : true"
```
jakobe, line 71: You've probably researched this, but is there any reason why `aria-disabled` and `disabled` differs?
Is it best practice to allow an otherwise disabled button to be focusable? 🤔

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -304,18 +302,17 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
...
       end: endOfWeek(now, { locale: this.locale }),
     });
 
-    return week.map((date) => this.getFirstLetterOfWeekDayCapitalized(date));
-  }
-
-  private getFirstLetterOfWeekDayCapitalized(date: Date) {
-    return this.formatWithLocale(date, 'EEEEE');
+    return week.map((date) => ({
+      firstLetterCapitalized: this.formatWithLocale(date, 'EEEEE'), // EEEEE returns the first letter of the day capitalized
```
jakobe: The comment you've added was the exact reason it was wrapped in a function with a telling name 😆

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -304,18 +302,17 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
...
+      fullName: this.formatWithLocale(date, 'EEEE'), // EEEE returns the full name of the day
```
jakobe: Consider naming the formats instead of using comments:
```suggestion
      const weekdayNarrowFormat = 'EEEEE';
      const weekdayWideFormat = 'EEEE';
      firstLetterCapitalized: this.formatWithLocale(date, weekdayNarrowFormat),
      fullName: this.formatWithLocale(date, weekdayWideFormat),
```

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -354,19 +351,27 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
...
         date: cellDate.getDate(),
         monthIndex: cellDate.getMonth(),
         year: cellDate.getFullYear(),
         isCurrentMonth: day.isCurrentMonth,
         isSelectable,
+        isFocussed,
         isSelected,
-        cssClasses: this.getCssClasses(day, isSelectable, isSelected),
+        isFocusable,
+        ariaLabel: this.formatWithLocale(cellDate, 'd MMMM'),
```
jakobe: Should the format string (order of day and month) be the same regardless of locale? And should it contain punctuation? 🤔 
I.e.:
* en-GB: "4 November
* en-US: ""November 4"
* da: "4. november"

https://englishlessonsbrighton.co.uk/how-to-write-dates-british-american-english/

jakobe (commented):
Really nice work!!! 👌🏻

libs/designsystem/calendar/src/calendar.component.ts
RasmusKjeldgaard: Hah, good point!!

libs/designsystem/calendar/src/calendar.component.spec.ts
RasmusKjeldgaard: I can definitely see why 😂  Improved readability in f9bafc02f (and 3ce12cb)🤞

libs/designsystem/calendar/src/calendar.component.html
RasmusKjeldgaard: This is to enhance the "explorability" of the calendar. The behavior is that dates and ranges of dates that are disabled via `disabledDates` or `enabledDates` are still focusable, but marked with `aria-disabled`. Screen readers will announce these dates as `dimmed`.

When dates are 'disabled' with min/max or `disableFutureDates`/`disablePastDates` we completely disable the button so it cannot be focussed with keyboard either, and is not available for users to find them via tabbing.

I'll try and make it more readable. 

Also, this got me thinking that the month-navigation buttons should probably always be available, but be announced as dimmed when min/max or disablePast/FutureDates are set.

libs/designsystem/calendar/src/calendar.component.ts
RasmusKjeldgaard: Good point, I added functionality for this in ea547b6

libs/designsystem/calendar/src/calendar.component.scss
```diff
@@ -48,17 +52,17 @@ td {
   align-items: center;
   justify-content: space-between;
 
-  button:disabled {
+  button[aria-disabled] {
```
jakobe: I know we only apply this attribute when it's `true`, but we probably should include this in the selector as to not make it brittle for future changes: 
```suggestion
  button[aria-disabled="true"]  {
```

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -240,20 +240,23 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
...
   }
 
+  private formatDateLabel(): string {
+    const localeDateFormats = {
+      da: 'd. MMMM',
+      'en-GB': 'd MMMM',
+      'en-US': 'MMMM d',
+    };
+
+    const dateFormat = localeDateFormats[this.locale.code] || 'MMMM d';
```
jakobe: Consider making it explicit that we default to `en-US`:
```suggestion
    const defaultDateFormat = localeDateFormats['en-US'];
    const dateFormat = localeDateFormats[this.locale.code] || defaultDateFormat;
```

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -240,20 +240,23 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
...
+    return dateFormat;
+  }
+
   ngOnInit() {
+    this.focussedDate = this.todayDate ? startOfDay(this.todayDate) : startOfDay(new Date());
```
jakobe: Consider simplifiying:
```suggestion
    this.focussedDate = startOfDay(this.todayDate ?? new Date());
```

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -479,13 +466,14 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
...
     this.activeMonth = add(this.activeMonth, { [unit]: index });
+    this.focussedDate = add(this.focussedDate, { [unit]: index });
 
     this.refreshActiveMonth();
   }
 
   get _canNavigateBack(): boolean {
-    const reachedPastDatesLimit =
-      this.disablePastDates && isSameMonth(this.activeMonth, this.todayDate);
+    const today = this.todayDate ? startOfDay(this.todayDate) : startOfDay(new Date());
```
jakobe: Consider simplifying:
```suggestion
    const today = startOfDay(this.todayDate ?? new Date());
```

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -495,8 +483,8 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
   }
 
   get _canNavigateForward(): boolean {
-    const reachedFutureDatesLimit =
-      this.disableFutureDates && isSameMonth(this.activeMonth, this.todayDate);
+    const today = this.todayDate ? startOfDay(this.todayDate) : startOfDay(new Date());
```
jakobe: Consider simplifying:
```suggestion
    const today = startOfDay(this.todayDate ?? new Date());
```

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -548,4 +525,76 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
...
+    }
+  }
+
+  private focusDate(newDate: Date | null) {
+    if (!newDate) return;
+    if (this.timezone === 'UTC') {
+      newDate = zonedTimeToUtc(this.subtractTimezoneOffset(newDate), this.timeZoneName);
+    }
+
+    const today = this.todayDate ? startOfDay(this.todayDate) : startOfDay(new Date());
```
jakobe: Consider simplifying:
```suggestion
    const today = startOfDay(this.todayDate ?? new Date());
```

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -351,67 +354,57 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
...
-    );
+  private isSelectable(date: Date, today: Date): boolean {
+    if (this.alwaysEnableToday && isSameDay(today, date)) return true;
+
+    if (this.isDisabledDate(date)) return false;
+    if (!this.isEnabledDate(date)) return false;
+    if (this.disableWeekends && isWeekend(date)) return false;
+    if (!this.isFocusable(date, today)) return false;
+
+    return true;
```
jakobe: Very nice refactoring! 👌🏻

libs/designsystem/calendar/src/helpers/calendar-cell.model.ts
```diff
@@ -1,9 +1,18 @@
-export interface CalendarCell {
+export interface CalendarCell extends CalendarCellMetadata {
```
jakobe: Consider renaming - it feels like this interface represents a day now more than a cell, also given that `cssClasses` has been removed:
```suggestion
export interface CalendarDay extends CalendarDayMetadata {
```

libs/designsystem/calendar/src/helpers/calendar-cell.model.ts
```diff
@@ -1,9 +1,18 @@
...
   date: number;
   monthIndex: number;
   year: number;
+  ariaLabel: string;
+}
+
+export interface CalendarCellMetadata {
```
jakobe: Consider renaming - it feels like this interface represents metadata for a day more than a cell:
```suggestion
export interface CalendarDayMetadata {
```

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -218,8 +217,9 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
   }
 
   constructor(
-    private calendarHelper: CalendarHelper,
-    @Inject(LOCALE_ID) locale: string
+    @Inject(LOCALE_ID) locale: string,
+    private elementRef: ElementRef,
+    private cdr: ChangeDetectorRef
```
jakobe, line 226: If we are handling changedetection ourselves, maybe reflect this in the component metadata, i.e. `changeDetection: ChangeDetectionStrategy.OnPush` 🤔

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -548,4 +525,76 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
...
+
+  private setFocussedCell(newDate: Date) {
+    const newDay = this.getCell(newDate);
+    if (newDay) {
+      if (this.focussedDay) {
+        this.focussedDay.isFocussed = false;
+      }
+      newDay.isFocussed = true;
+      this.focussedDay = newDay;
+    }
```
jakobe: Consider guarding same was as `focusDate` method below:
```suggestion
    if (!newDay) return;
    if (this.focussedDay) {
        this.focussedDay.isFocussed = false;
     }
     newDay.isFocussed = true;
     this.focussedDay = newDay;
```

libs/designsystem/calendar/src/calendar.component.ts
```diff
@@ -351,67 +354,57 @@ export class CalendarComponent implements OnInit, AfterViewInit, OnChanges {
...
-      disabled: day.isDisabled,
-    };
-    let cssClassString = 'day';
-    for (const key in cssClasses) {
-      if (cssClasses[key]) {
-        cssClassString += ' ' + key;
-      }
-    }
-    return cssClassString;
+  private isFocusable(date: Date, today: Date) {
```
jakobe: I see we reuse this in the `isSelectable` method, which made me wonder "why does the ability to focus control whether it's selectable"? 🤔 
So maybe rename or extract to new method:
```suggestion
  private isWithinAllowedRange(date: Date, today: Date) {
```

jakobe (commented):
I see a lot of places, where we define `today` from either `this.todayDate` or "now" (i.e. `new Date()` and use it as parameter for various methods. It smells like this could be a member (field or method) on the component class itself? 🤔

jakobe (approved):
📆 👌🏻

PR #3707, opened 2024-11-14
Segmented Control: a11y improvements 

### LINKED ISSUE #3687: [Enhancement] A11y improvements for Segmented Control

## Describe the enhancement
To make Segmented Control support proper announcing to screen readers we need to:
- [x] ~~Announce total number of Segment tabs (1/1 -> 1/3)~~
   [This is a known bug in Chrome with VoiceOver on macOS](https://bugs.chromium.org/p/chromium/issues/detail?id=1405462)
- [x] Announce selected tab
- [x] Implement [Tabs pattern from ARIA Authoring Practices Guide](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/)

Also filed a bug issue in the Ionic repo: ionic-team/ionic-framework#30022 

<hr />

## Checklist:

### Refinement

- [x] Request that the issue is [UX refined](https://github.com/kirbydesign/designsystem/blob/develop/.github/CONTRIBUTING.md#ux-refinement)
- [x] Request that the issue is [tech refined](https://github.com/kirbydesign/designsystem/blob/develop/.github/CONTRIBUTING.md#tech-refinement); do not proceed until this is done.

### Implementation 

- [x] Make sure you have read: "[Before you get coding](../CONTRIBUTING.md/#before-you-get-coding)".
- [x] Create a branch from the [develop branch](https://github.com/kirbydesign/designsystem/tree/develop) following our [branch naming convention](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Branch). 
- [x] Publish a WIP implementation to Github as a draft PR and ask for feedback. 
- [x] Update the [cookbook](https://cookbook.kirby.design) with examples and showcases.

### Review

jakobe:

This PR closes #3687

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  

### DIFF
--- a/apps/cookbook/src/app/showcase/segmented-control-showcase/segmented-control-showcase.component.html
+++ b/apps/cookbook/src/app/showcase/segmented-control-showcase/segmented-control-showcase.component.html
@@ -146,9 +146,9 @@
 <!-- prettier-ignore -->
 <ul>
   <li>
-    <kbd>Tab</kbd>:
+    <kbd>Tab</kbd>
     <ul>
-      <li>When focus moves into the segmented control, places focus on the selected segment button. If there is no selected segment button moves to the first segment button</li>
+      <li>When focus moves into the segmented control, places focus on the selected segment button. If none is selected, places focus on the first segment button</li>
       <li>When the segmented control contains the focus, moves focus to the next element in the page tab sequence outside the control</li>
     </ul>
   </li>
--- a/apps/cookbook/src/app/showcase/segmented-control-showcase/segmented-control-showcase.component.html
+++ b/apps/cookbook/src/app/showcase/segmented-control-showcase/segmented-control-showcase.component.html
@@ -5,185 +5,177 @@
 
 <kirby-divider [hasMargin]="true"></kirby-divider>
 
-<div class="example">
-  <h2>Examples</h2>
-  <h3>Default</h3>
-  <cookbook-example-viewer [html]="example.template" [ts]="example.codeSnippet">
-    <cookbook-segmented-control-example-default
-      #example
-    ></cookbook-segmented-control-example-default>
-  </cookbook-example-viewer>
+<h2>Examples</h2>
+<h3>Default</h3>
+<cookbook-example-viewer [html]="example.template" [ts]="example.codeSnippet">
+  <cookbook-segmented-control-example-default #example></cookbook-segmented-control-example-default>
+</cookbook-example-viewer>
 
-  <h3>Size</h3>
-  <p>
-    Segmented Control can be rendered in two different sizes when in default mode:
-    <code>sm</code>
-    and
-    <code>md</code>
-    .
-  </p>
-  <cookbook-example-viewer [html]="sizeExample.template">
-    <cookbook-segmented-control-example-sizes
-      #sizeExample
-    ></cookbook-segmented-control-example-sizes>
-  </cookbook-example-viewer>
+<h3>Size</h3>
+<p>
+  Segmented Control can be rendered in two different sizes when in default mode:
+  <code>sm</code>
+  and
+  <code>md</code>
+  .
+</p>
+<cookbook-example-viewer [html]="sizeExample.template">
+  <cookbook-segmented-control-example-sizes #sizeExample></cookbook-segmented-control-example-sizes>
+</cookbook-example-viewer>
 
-  <h3>Mode</h3>
-  <p>
-    Use the
-    <code>mode</code>
-    property to control the rendering of each segment. In addiditon to the
-    <code>default</code>
-    mode, Segmented Control supports
-    <code>chip</code>
-    and
-    <code>compactChip</code>
-    mode.
-  </p>
-  <cookbook-example-viewer [html]="modeExample.template" [ts]="modeExample.codeSnippet">
-    <cookbook-segmented-control-example-modes
-      #modeExample
-    ></cookbook-segmented-control-example-modes>
-  </cookbook-example-viewer>
+<h3>Mode</h3>
+<p>
+  Use the
+  <code>mode</code>
+  property to control the rendering of each segment. In addiditon to the
+  <code>default</code>
+  mode, Segmented Control supports
+  <code>chip</code>
+  and
+  <code>compactChip</code>
+  mode.
+</p>
+<cookbook-example-viewer [html]="modeExample.template" [ts]="modeExample.codeSnippet">
+  <cookbook-segmented-control-example-modes #modeExample></cookbook-segmented-control-example-modes>
+</cookbook-example-viewer>
 
-  <h3>Grouped</h3>
-  <p>
-    Segmented Control can be combined with a Button to create a related group of choices with one
-    option separate from the others—all mutually exclusive. Use this pattern to create a filter as
-    in the example below.
-  </p>
-  <p>
-    <em>
-      Note: If changing the
-      <code>size</code>
-      of the Segmented Control be sure to set the same
-      <code>size</code>
-      for the Button.
-    </em>
-  </p>
-  <cookbook-example-viewer
-    [html]="groupedExample.template"
-    [ts]="groupedExample.codeSnippet"
-    [css]="groupedExample.styles"
+<h3>Grouped</h3>
+<p>
+  Segmented Control can be combined with a Button to create a related group of choices with one
+  option separate from the others—all mutually exclusive. Use this pattern to create a filter as in
+  the example below.
+</p>
+<p>
+  <em>
+    Note: If changing the
+    <code>size</code>
+    of the Segmented Control be sure to set the same
+    <code>size</code>
+    for the Button.
+  </em>
+</p>
+<cookbook-example-viewer
+  [html]="groupedExample.template"
+  [ts]="groupedExample.codeSnippet"
+  [css]="groupedExample.styles"
+>
+  <cookbook-segmented-control-example-grouped
+    #groupedExample
+  ></cookbook-segmented-control-example-grouped>
+</cookbook-example-viewer>
+
+<h3>Badge</h3>
+<p>
+  Badges can be applied to Segmented Control in
+  <code>default</code>
+  mode. A badge can be added to any
+  <code>SegmentItem</code>
+  using the
+  <code>badge</code>
+  property.
+</p>
+<p>
+  <a
+    href="https://github.com/kirbydesign/designsystem/blob/main/libs/designsystem/src/lib/components/segmented-control/segment-item.ts"
+    class="kirby-external-icon"
+    target="_blank"
   >
-    <cookbook-segmented-control-example-grouped
-      #groupedExample
-    ></cookbook-segmented-control-example-grouped>
-  </cookbook-example-viewer>
+    See SegmentItem on Github
+  </a>
+</p>
+<p>
+  The badge can contain either a plain text string through
+  <code>badge.content</code>
+  or an icon provided as the name of an icon in the
+  <code>badge.icon</code>
+  property.
+  <br />
+  When using an icon the
+  <code>badge.description</code>
+  property should be set which will be used as the
+  <code>aria-label</code>
+  of the badge for accessibility.
+  <br />
+  The background color of the badge can be controlled through
+  <code>badge.themeColor</code>
+  .
+</p>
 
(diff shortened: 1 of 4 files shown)


libs/designsystem/src/lib/components/segmented-control/segmented-control.component.ts
```diff
@@ -150,4 +151,25 @@ export class SegmentedControlComponent<TItem extends SegmentItem = SegmentItem>
...
+    if (item.id === this.value?.id) return null;
+    // Allow tab stop on first item if no value is set:
+    if (!this.value && index === 0) return null;
+    // Prevent tab stop on all other items:
+    return -1;
+  }
+
+  @HostListener('focusin', ['$event'])
+  @HostListener('focusout', ['$event'])
+  _noop() {
```
RasmusKjeldgaard, line 172: Consider making this private so it does not show up in the component type
```suggestion
  private noop() {
```

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.ts
```diff
@@ -150,4 +151,25 @@ export class SegmentedControlComponent<TItem extends SegmentItem = SegmentItem>
...
     }
   }
+
+  focusNativeButton(event: UIEvent) {
+    (event.target as HTMLIonSegmentButtonElement)?.setFocus();
+  }
+
+  getTabIndex(item: TItem, index: number) {
+    // When focused prevent tab stop from inner native button to outer ion-segment-button:
+    if (this.ionSegmentElement.nativeElement.matches(':focus-within')) return -1;
```
RasmusKjeldgaard: Consider extracting to const `segmentElementHasFocusWithin`, `isFocusWithinSegmentElement` or similar

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.ts
RasmusKjeldgaard: Consider making this explicitly private so it does not show up in the component type
```suggestion
  private noop() {
```

RasmusKjeldgaard (commented):
👏 👏

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.ts
jakobe: Actually a really good point - but:

1. We don't use that pattern anywhere else - instead we prefix with `_` to denote as private/internal
2. Marking the method as private results in a `_noop is declared but its value is never read.` ts warning

As I recall there used to be a time where Angular insisted on the method being public 🤔  It doesn't seem that's a requirement anymore, but maybe that's why all methods decorated with `@HostListener` in our codebase are public? 🤷🏻

RasmusKjeldgaard: For me the warning is acceptable in this case, but I see your point about  consistency!

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.ts
```diff
@@ -150,4 +151,26 @@ export class SegmentedControlComponent<TItem extends SegmentItem = SegmentItem>
...
+    if (segmentElementHasFocus) return -1;
+    // Allow tab stop on selected item:
+    if (item.id === this.value?.id) return null;
+    // Allow tab stop on first item if no value is set:
+    if (!this.value && index === 0) return null;
+    // Prevent tab stop on all other items:
+    return -1;
+  }
+
+  @HostListener('focusin', ['$event'])
```
Fuzzy3: One way to get rid of your function call from the template could be that you call the getTabIndex() for each item in your "focusin/focusout" hostlistener and then add the tabindex result to the item model. 
But i'm fine with leaving it as is, since your performance tests were fine ! -> I guess it will never really become a problem since it limited how many item there will ever be ;-)

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.spec.ts
```diff
@@ -68,6 +69,167 @@ describe('SegmentedControlComponent', () => {
...
+        initSelectedSegmentButton = spectator.queryHost<HTMLIonSegmentButtonElement>(
+          'ion-segment-button.segment-button-checked'
+        );
+        await initSelectedSegmentButton.setFocus();
+      });
+
+      it('should focus previous item when ArrowLeft is pressed', () => {
+        spectator.keyboard.pressKey('ArrowLeft', ionSegmentElement, 'keydown');
+
+        expect(document.activeElement).toEqual(initSelectedSegmentButton.previousElementSibling);
```
Fuzzy3, line 85: i never knew something like "nextElementSibling/previousElementSibling" existed 😮 SO COOL !

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.spec.ts
```diff
@@ -68,6 +69,167 @@ describe('SegmentedControlComponent', () => {
...
+
+      it('should focus next item when ArrowRight is pressed', () => {
+        spectator.keyboard.pressKey('ArrowRight', ionSegmentElement, 'keydown');
+
+        expect(document.activeElement).toEqual(initSelectedSegmentButton.nextElementSibling);
+      });
+
+      it('should focus first item when Home is pressed', async () => {
+        const lastSegmentButton = spectator.queryHost<HTMLIonSegmentButtonElement>(
+          'ion-segment-button:last-of-type'
```
Fuzzy3, line 96: last-of-type is also new for me !! 😮

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.spec.ts
```diff
@@ -68,6 +69,167 @@ describe('SegmentedControlComponent', () => {
...
+        const segmentButtons =
+          spectator.queryHostAll<HTMLIonSegmentButtonElement>('ion-segment-button');
+
+        const segmentButtonsWithTabStop = segmentButtons.filter(
+          (button) => !button.hasAttribute('tabindex')
+        );
+        const segmentButtonsWithoutTabStop = segmentButtons.filter(
+          (button) => button.getAttribute('tabindex') === '-1'
+        );
+        expect(segmentButtonsWithTabStop).toHaveLength(1);
```
Fuzzy3, line 209: Would it be an idea to check that its the first segmentButton in the array that is without tabindex?

Fuzzy3 (commented):
❤️

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.spec.ts
jakobe: Good point! I've added 2 additional tests (with/without selected segment) that covers this 👍🏻

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.ts
jakobe: Good point 👌🏻 

In practice the method is only called on init and on focusin/out (i.e. tab into component, navigatin with arrow keys and selecting an item) so as far as I can see it would result in the exact same number of methods calls anyhoo 🤷🏻

Fuzzy3: I think it might resolve in more method calls than you're suggesting, since angular doesn't know which value the function will return and thereby it has to rerun it a lot of times, until it has convinced itself that the value will not change. 
**Example with segmented control having 3 segmented buttons:** 
When angular reaches the first button that needs to be rendered it calls the getTabIndex() function and gets a value for the attr.tabindex for the first button. Then it reaches the second button and calls the same function. However now angular doesn't know if second time the function was called, changed some things that could result in a different value for the first button - and now it has to rerun that again for the first button. It will keep rerunning this function for both buttons until it is convinced that they won't change. And then angular reaches the third button - calls the function again but now it needs to check with the first button and second button again.

All in all a lot of change detection cycles run - but again even 100cycles doesn't increase performance so I dont think we should do anything with this.

apps/cookbook/src/app/showcase/segmented-control-showcase/segmented-control-showcase.component.html
```diff
@@ -2,135 +2,180 @@
...
+    class="kirby-external-icon"
+    target="_blank"
+    href="https://www.w3.org/WAI/ARIA/apg/patterns/tabs/"
+  >
+    Tabs Pattern from ARIA Authoring Practices Guide
+  </a>
+</p>
+<h3>Keyboard support</h3>
+<p>
+  The component has full keyboard support for navigating between and selecting segment buttons .
```
Fuzzy3:
```suggestion
  The component has full keyboard support for navigating between and selecting segment buttons.
```

apps/cookbook/src/app/showcase/segmented-control-showcase/segmented-control-showcase.component.html
```diff
@@ -2,135 +2,180 @@
...
+  <kbd>Space</kbd>
+  to select the element that is currently focused.
+</p>
+<p>The following keys can be used to navigate within the control:</p>
+<!-- prettier-ignore -->
+<ul>
+  <li>
+    <kbd>Tab</kbd>:
+    <ul>
+      <li>When focus moves into the segmented control, places focus on the selected segment button. If there is no selected segment button moves to the first segment button</li>
```
Fuzzy3:
```suggestion
      <li>When tabbing into into the segmented control, the focus is placed on the selected segment button. If there is no selected segment button, the focus is placed on the first</li>
```

apps/cookbook/src/app/showcase/segmented-control-showcase/segmented-control-showcase.component.html
```diff
@@ -2,135 +2,180 @@
...
+      <li>When the segmented control contains the focus, moves focus to the next element in the page tab sequence outside the control</li>
```
Fuzzy3, line 152:
```suggestion
      <li>When tabbing while the focus is within the segmented control, the focus is placed on the next element in the page tab sequence outside the control</li>
```

Fuzzy3 (changes requested):
Great documentation on keyboard support !

apps/cookbook/src/app/showcase/segmented-control-showcase/segmented-control-showcase.component.html
```diff
@@ -2,135 +2,180 @@
...
+</p>
+<h3>Keyboard support</h3>
+<p>
+  The component has full keyboard support for navigating between and selecting segment buttons .
+  Keyboard navigation will only focus segment buttons, use
+  <kbd>Enter &#8629;</kbd>
+  or
+  <kbd>Space</kbd>
+  to select the element that is currently focused.
+</p>
```
RasmusKjeldgaard: Consider simplifying a bit
```suggestion
  Keyboard navigation moves focus between segment buttons, and
  <kbd>Enter &#8629;</kbd>
  or
  <kbd>Space</kbd>
  selects the element that is currently focused.
</p>
```

libs/designsystem/src/lib/components/segmented-control/segmented-control.component.ts
jakobe: @Fuzzy3 as we talked about at demo last week this is not the case 🤓

apps/cookbook/src/app/showcase/segmented-control-showcase/segmented-control-showcase.component.html
jakobe: @Fuzzy3 I'd like to keep the wording as it's taken directly from the [Tabs Pattern in the ARIA Authoring Practices Guide (APG)](https://www.w3.org/WAI/ARIA/apg/)

apps/cookbook/src/app/showcase/segmented-control-showcase/segmented-control-showcase.component.html
jakobe: @Fuzzy3 I'd like to keep the wording as it's taken directly from the [Tabs Pattern in the ARIA Authoring Practices Guide (APG)](https://www.w3.org/WAI/ARIA/apg/)

Fuzzy3 (approved):

PR #3898, opened 2025-04-10
Checkbox, Radio, Toggle: fix screen reader announcement of both visible and aria label

### LINKED ISSUE #3883: [BUG] Setting aria-label on some elements that have a text label makes the screen reader read both

## Describe the bug
On some Kirby elements, setting the `aria-label` will not override the accessible text if the element also has visible text. In stead, the value of the aria-label is prefixed to the visible text and the screen reader will read both.

The issue can be reproduced with a kirby-radio:

```
<kirby-radio
      [text]="'Visible text'"
      [attr.aria-label]="'Accessible text'"
></kirby-radio>
```

10.3.0

Only the value in the aria-label is read.

- OS: MacOS
- Browser Chrome

<hr />

## Checklist:

### Verification

### Implementation 

### Review

- [ ] Create a pull-request.

RasmusKjeldgaard:

This PR closes #3883

## What is the new behavior?
Whenever a hidden label (which is a consumer-defined aria-label or a fallback label inferred from any parent item) is present, we make sure to hide the visible label from screen readers, so it it possible to define both a visible label and an alternate aria-label without screen readers announcing both. 

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  

### DIFF
--- a/libs/designsystem/checkbox/src/checkbox.component.html
+++ b/libs/designsystem/checkbox/src/checkbox.component.html
@@ -9,8 +9,8 @@
   [justify]="_justify"
 >
   <span class="hidden-label" *ngIf="_labelText">{{ _labelText }}</span>
-  <span *ngIf="text">{{ text }}</span>
-  <span class="default-content">
+  <span *ngIf="text" [attr.aria-hidden]="_labelText ? true : undefined">{{ text }}</span>
+  <span class="default-content" [attr.aria-hidden]="_labelText ? true : undefined">
     <ng-content></ng-content>
   </span>
 </ion-checkbox>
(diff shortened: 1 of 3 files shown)


troelslenda (approved):
LGTM.
I'm curious though. why not simply use aria-label on ion-checkbox, ion-radio and ion-toggle  instead of introducing hidden-label and the aria-hidden property? Wouldn't it be the same?

RasmusKjeldgaard:
> LGTM. I'm curious though. why not simply use aria-label on ion-checkbox, ion-radio and ion-toggle instead of introducing hidden-label and the aria-hidden property? Wouldn't it be the same?

Definitely the right question to ask! Unfortunately, the behavior of ion-checkbox etc. is different when theres no label applied between the elements tags. So when the label comes from e.g. a `kirby-label` sibling element or from an aria-label, the ion element will not expand itself to make the whole item clickable. 

Therefore we go through a number of hoops to ensure that the ionic element thinks there is a visible label inside it, so it will position itself correctly in item.

PR #3914, opened 2025-04-30
Chart: Add cookbook stories to storybook

Fuzzy3:

This PR is a preburner for upgrading the chart.js version.


libs/designsystem/chart/src/stock-chart/stock-chart.component.stories.ts
```diff
@@ -68,3 +72,15 @@ export const StockChart: Story = {
     dataLabelOptions: chartDataLabelOptions,
   },
 };
+
+export const CookbookStockChartExample: Story = {
```
RasmusKjeldgaard: There is an underlying issue with our cookbook structure here, that I think would be great to fix now that we are at it, instead of introducing multiple cookbook example stories in storybook.

Since the stock chart already is a separate component with its own showcase, my suggestion is to move the two stock chart examples into another folder, so the structure is changed to something that resembles our regular structure:

```
...
charts-example/
stock-charts-example/
    examples/
        single-dataset.ts
        comparison.ts
    stock-chart-example.component.ts
...
```

Then the two stock examples can be added inside the stock-chart-example component, and we can snapshot that.

One additional reason for doing this change, is that it fixes the dead 'Pop out' link on the stock chart cookbook page.

RasmusKjeldgaard (commented):
Looks really good, got one suggestion for the structure of the stock chart stuff.

RasmusKjeldgaard (approved):

PR #3977, opened 2025-06-23
Flaky test: make focus heading test in router outlet more robust 

RasmusKjeldgaard:
This attempts to fix flakyness of test, as comparing the elements is not always successful. Ionic is adding a tabindex attribute asynchronously on page enter that seems to cause some noise when comparing.


Fuzzy3 (approved):
Fut fut 🚋 🚋 🚋

PR #4010, opened 2025-07-21
Dropdown: A11y improvements for dropdown

### LINKED ISSUE #3683: [BUG] A11y improvements for Dropdown
## Describe the bug

Please ensure that the dropdown component announces the item the focusfield is hovering over, when using keyboard. 
it only announces the item, when it is selected. 

Tech-refinement @loujuu

Following improvements should be considered to improve accessibility of the dropdown:

Considerations regarding best suited pattern for dropdown component:

When to Use the [Select-Only Combobox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/) 
Use this pattern if your dropdown:
* Is triggered by a button or input
* Allows the user to select one item from a list
* Does not allow text input
* Behaves like a native <select> element, but with custom styling or templating. Users can type characters to select matching options.
* Uses a button to open a popover
* like an HTML <select>, users can type characters to select matching options.

When to Use the [Menu button](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/) Pattern
* A button that opens a menu 
* Offers a list of choices to the user, such as a set of actions or functions.
* Common use case: Context menus, toolbars, dropdowns with actions (e.g., “Edit”, “Delete”)

The Menu Button pattern is for triggering actions, not selecting values. It uses role="menu" and role="menuitem", which would mislead screen readers into thinking your dropdown is a command menu.

- OS: [e.g. Windows, MacOS, iOS]
- Browser [e.g. Chrome, Safari]
- Browser version [e.g. 22]

<hr />

## Checklist:

### Verification

### Implementation 

### Review

- [ ] Create a pull-request.

Loujuu:

This PR closes #3683

## What is the new behavior?

Improvement regarding accessibility for dropdown according to select-only combobox pattern. Remaining keboard improvements will be fixed in another task.

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


libs/designsystem/dropdown/src/dropdown.component.html
```diff
@@ -26,7 +31,7 @@
 </ng-template>
 
 <ng-template #itemWrapperTemplate>
-  <kirby-card>
+  <kirby-card role="listbox" [attr.id]="_listboxId" [attr.aria-labelledby]="_comboboxId + '-label'">
```
Fuzzy3: aria-labelledby should point to the combo-box, which has the label, so I think the '-label' part should be removed such that it matches the id of the combobox
```suggestion
  <kirby-card role="listbox" [attr.id]="_listboxId" [attr.aria-labelledby]="_comboboxId">
```

libs/designsystem/dropdown/src/dropdown.component.html
```diff
@@ -56,9 +61,12 @@
     (click)="onItemSelect(index)"
     [class.focused]="focused"
     role="option"
+    [attr.id]="_listboxId + '-item-' + index"
+    [attr.aria-selected]="selected"
+    tabindex="0"
```
Fuzzy3: I don't think you need the tabindex here. When we have selectable="true" on an item, ionic will automatically render a focusable button behind the scenes, so I suggest to remove the tabindex="0" here. 
```suggestion
```

libs/designsystem/dropdown/src/dropdown.component.spec.ts
```diff
@@ -126,19 +126,51 @@ describe('DropdownComponent', () => {
...
+
+    it('should set aria-activedescendant when focusedIndex is set', () => {
+      spectator.component.focusedIndex = 1;
+      spectator.detectChanges();
+      expect(buttonElement.getAttribute('aria-activedescendant')).toBe(
+        listboxId + '-item-' + spectator.component.focusedIndex
+      );
+    });
+
+    it('should set correct aria attributes on listbox and options when open', () => {
```
Fuzzy3: I suggest to split this into two unit tests. One for the listbox and one for the options. A good rule of thumb is to not have any logic after your `expect()`

apps/cookbook/src/app/showcase/dropdown-showcase/dropdown-showcase.component.html
```diff
@@ -85,6 +85,39 @@ <h3>Forms</h3>
...
+    target="_blank"
+    href="https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/"
+  >
+    Select-only Combobox Pattern
+  </a>
+  from ARIA Authoring Practices Guide.
+</p>
+<p>
+  The Dropdown is a button that toggles a list of items. It uses the
+  <code>aria-haspopup="listbox"</code>
```
Fuzzy3: There is no reason to mention `aria-haspopup`, `aria-expanded` or `aria-activedescendant` as these are not something for the consumer to consider, but merely internal logic.

apps/cookbook/src/app/showcase/dropdown-showcase/dropdown-showcase.component.html
```diff
@@ -85,6 +85,39 @@ <h3>Forms</h3>
     ></cookbook-dropdown-example-ng-forms>
   </div>
 </cookbook-example-viewer>
+<h2>Accessibility</h2>
+<p>
+  The Dropdown implements the
```
Fuzzy3, line 90: Yes and no 😄 The dropdown doesn't fully implement that pattern before you also have the next issue done, but lets just leave this for now.

apps/cookbook/src/app/showcase/dropdown-showcase/dropdown-showcase.component.html
```diff
@@ -85,6 +85,39 @@ <h3>Forms</h3>
...
+  to reflect the dropdown's state and the currently focused item. To ensure screen readers announce
+  the dropdown correctly, it is required to provide either an
+  <code>aria-label</code>
+  or an
+  <code>aria-labelledby</code>
+  attribute to give it an accessible name. The dropdown component will apply this attribute to the
+  internal button element.
+</p>
+<ul>
+  <li>Use aria-label if you do not have a visible label.</li>
```
Fuzzy3: I dont think we need this explanation of when to use `aria-label` or `aria-labelledby`. I believe we should only mention this if there are special cases to it 🤔

Fuzzy3 (changes requested):
Looks really nice ! I have a few comments on the documentation and unit test part.
 
Functionality wise everything looks right ! However I think there is still something missing. When opening the dropdown in the example here: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/ it directly focusses the first element, but in our current implementation we need to first select the listbox manually with a weird shift+control+option+right-arrow combination (or something like that)before we can navigate between the elements. I found out that this is due the fact, that when we open our dropdown by a click-event (which is the same event triggered when doing control+option+space with VO) then the focussedIndex is not set to the first element, and thus the aria-activedescendant is not set to point at the first element, which somehow causes this "middle" state.

I got it working by updating line 403 in the dropdown.component.ts file to check if selectedItem is -1 and if so, then setting focussedIndex to 0. See if that also works for you 😄

apps/cookbook/src/app/showcase/dropdown-showcase/dropdown-showcase.component.html
Loujuu: Keyboard documentation is related to part two of the dropdown task (#4015) with keyboard support implementation.

apps/cookbook/src/app/showcase/dropdown-showcase/dropdown-showcase.component.html
Loujuu: It is required for the consumer to add a aria-label or a aria-labelledby in the kirby-dropdown component. I do not think this should be deleted.

libs/designsystem/dropdown/src/dropdown.component.html
Loujuu: I will delete -label.

Fuzzy3 (changes requested):
Almost there ! looks like you missed the "focus the first element on open" comment from last review 😸

apps/cookbook/src/app/showcase/dropdown-showcase/dropdown-showcase.component.html
Fuzzy3: Then maybe just remove the three bullets 😸 it is already described in the sentence before that aria-label or aria-labelledby is required

Fuzzy3 (changes requested):
I have a single comment to the documentation and then a revert of the last to commits needs to be done, since some accidental code came in 😛

Loujuu:
We have agreed on adding focus on first item in dropdown, this.focusedIndex = this.selectedIndex > -1 ? this.selectedIndex : 0; even though it does not follow the select-only-keyboard pattern, regarding keys with no visible focus.

libs/designsystem/dropdown/src/dropdown.component.spec.ts
Fuzzy3: I would still like to see this one split into two :)

libs/designsystem/dropdown/src/dropdown.component.spec.ts
```diff
@@ -46,17 +46,21 @@ describe('DropdownComponent', () => {
...
-        `<kirby-dropdown [selectedIndex]="selectedIndex" [items]="items" [size]="size"></kirby-dropdown>`,
+        `<kirby-dropdown [selectedIndex]="selectedIndex" aria-labelledby="labelId" [items]="items" [size]="size"></kirby-dropdown>`,
         {
           hostProps: {
             items: items,
           },
         }
       );
       buttonElement = spectator.query('button[kirby-button]');
+      listboxId = spectator.component._listboxId;
```
Fuzzy3: Since listboxId and comboboxId is only something we use for a few tests, I suggest to not have them defined here, but in each individual test where needed

libs/designsystem/dropdown/src/dropdown-popover.component.spec.ts
```diff
@@ -106,6 +107,66 @@ describe('DropdownComponent (popover version)', () => {
...
+    describe('listbox', () => {
+      it('should set correct ARIA attributes on listbox when open', () => {
+        spectator.component.open();
+        spectator.detectChanges();
+        const listbox = spectator.query('[role="listbox"]');
+        expect(listbox).toBeTruthy();
+        expect(listbox.getAttribute('id')).toBe(listboxId);
+      });
+
+      it('should have correct item size', fakeAsync(() => {
```
Fuzzy3: This test about items snuck inside the listbox describe :) I'm not sure we need a whole describe for the single listbox test.

libs/designsystem/dropdown/src/dropdown-popover.component.spec.ts
```diff
@@ -106,6 +107,66 @@ describe('DropdownComponent (popover version)', () => {
...
+    });
+
+    it('should have correct ID on button', () => {
+      expect(buttonElement.getAttribute('id')).toBe(comboboxId);
+    });
+
+    it('should update aria-expanded when opened', () => {
+      spectator.component.open();
+      spectator.detectChanges();
+      expect(buttonElement.getAttribute('aria-expanded')).toBeTruthy();
```
Fuzzy3:
```suggestion
      expect(buttonElement.getAttribute('aria-expanded')).toBeTrue();
```

libs/designsystem/dropdown/src/dropdown-popover.component.spec.ts
```diff
@@ -106,6 +107,66 @@ describe('DropdownComponent (popover version)', () => {
       expect(buttonElement.attributes['disabled']).toBeUndefined();
     });
 
+    it('should have correct ARIA attributes on button', () => {
+      expect(buttonElement.getAttribute('role')).toBe('combobox');
+      expect(buttonElement.getAttribute('aria-haspopup')).toBe('listbox');
+      expect(buttonElement.getAttribute('aria-controls')).toBe(listboxId);
+      expect(buttonElement.getAttribute('aria-expanded')).toBe('false');
```
Fuzzy3:
```suggestion
      expect(buttonElement.getAttribute('aria-expanded')).toBeFalse();
```

libs/designsystem/dropdown/src/dropdown.component.spec.ts
```diff
@@ -126,19 +126,51 @@ describe('DropdownComponent', () => {
...
+    });
+
+    it('should update aria-expanded when opened', () => {
+      spectator.component.open();
+      spectator.detectChanges();
+      expect(buttonElement.getAttribute('aria-expanded')).toBeTruthy();
+    });
+
+    it('should set aria-activedescendant when focusedIndex is set', () => {
+      spectator.component.focusedIndex = 1;
```
Fuzzy3: By default focusIndex is set to 1 when opened, so I think a better test would be to set focusInded to 2 here, since the focusIndex might already be set to 1.

libs/designsystem/dropdown/src/dropdown.component.stories.ts
```diff
@@ -85,7 +85,7 @@ export const DropdownOpened: Story = {
   play: async ({ canvasElement }) => {
     const canvas = within(canvasElement);
 
-    const dropdownToOpen = canvas.getByRole('button', {
+    const dropdownToOpen = canvas.getByRole('combobox', {
       name: 'Item 1',
     });
```
Fuzzy3:
```suggestion
    const dropdownToOpen = canvas.getByRole('combobox');
```

Fuzzy3: This should fix the error in the storybook interactive test

libs/designsystem/dropdown/src/dropdown.component.ts
```diff
@@ -385,7 +400,8 @@ export class DropdownComponent implements AfterViewInit, OnDestroy, ControlValue
       );
 
       // Move focus to selected item (if any)
-      this.focusedIndex = this.selectedIndex;
+      // If no item is selected, focus the first item
+      this.focusedIndex = this.selectedIndex > -1 ? this.selectedIndex : 0;
```
Fuzzy3, line 404: Lets make a test that checks that seletedIndex is set to 0 when the dropdown is opened

libs/designsystem/dropdown/src/dropdown.component.spec.ts
```diff
@@ -126,19 +126,51 @@ describe('DropdownComponent', () => {
...
     });
 
-    it('should receive focus on button click', () => {
-      spectator.click('button');
-      expect(spectator.element).toBeFocused();
+    it('should have correct aria attributes on button', () => {
+      expect(buttonElement.getAttribute('role')).toBe('combobox');
+      expect(buttonElement.getAttribute('aria-haspopup')).toBe('listbox');
+      expect(buttonElement.getAttribute('aria-controls')).toBe(listboxId);
+      expect(buttonElement.getAttribute('aria-expanded')).toBe('false');
```
Fuzzy3:
```suggestion
      expect(buttonElement.getAttribute('aria-expanded')).toBeFalse();
```

Fuzzy3 (changes requested):
I have a few comments mainly about the unit tests. An important thing I'm seing is that the tests changed and the new tests introduced differ a little bit in the popover spec vs the regular dropdown spec. I think they should have the exact same unit tests, such that we test that dropdown works exactly the same no matter if we use dropover or not. I suggest that you make all unit tests correct in one of the spec files and the copy those over to the other one :) 
Keep up the good work !

libs/designsystem/dropdown/src/dropdown-popover.component.spec.ts
Loujuu: toBeFalse() is not appropriate here because getAttribute('aria-expanded') returns a string, not a boolean. ARIA attributes are always represented as strings in the DOM, so 'false' ≠ false. I suggest to keep expect(buttonElement.getAttribute('aria-expanded')).toBe('false');

libs/designsystem/dropdown/src/dropdown-popover.component.spec.ts
Loujuu: toBeTrue() is not suitable here because getAttribute('aria-expanded') returns a string, not a boolean. expect(buttonElement.getAttribute('aria-expanded')).toBeTrue(); Also expects a boolean. I suggest to change to 
expect(buttonElement.getAttribute('aria-expanded')).toBe('true');

libs/designsystem/dropdown/src/dropdown.component.spec.ts
Loujuu: Good idea :-)

libs/designsystem/dropdown/src/dropdown.component.spec.ts
Loujuu: toBeFalse() is not appropriate here because getAttribute('aria-expanded') returns a string, not a boolean. ARIA attributes are always represented as strings in the DOM, so 'false' ≠ false. I suggest to keep expect(buttonElement.getAttribute('aria-expanded')).toBe('false');

libs/designsystem/dropdown/src/dropdown.component.spec.ts
Loujuu: I do think that focusIndex is set to 0 when opened. But I have changes test to 2.

libs/designsystem/dropdown/src/dropdown.component.spec.ts
Loujuu: I did split into two, but it was a late re-push. I forgot to inform you to refresh.

libs/designsystem/dropdown/src/dropdown.component.stories.ts
Loujuu: Great :-)

libs/designsystem/dropdown/src/dropdown.component.ts
Loujuu: I do not think it is expected for the selectedIndex to be 0, when dropdown is opened. It is not intended to have a default selected item. But it is intended to have a focused item, if an item is selected. If no item is selected, then we should expect focusIndex to be 0, but not the selectedIndex. I have added two tests to check for should set focusedIndex to selectedIndex if selectedIndex is set and should set focusedIndex to 0 if no selectedIndex is set instead.

Fuzzy3: you are right :)

Fuzzy3 (approved):

PR #4079, opened 2025-10-10
Sidebar: Add new Sidebar component

Jonaythaz:
Hi Kirby,

I have tried to add the sidebar menu to kirby extensions with as few changes as possible, only adding some tests and a storybook, but let me know if there is anything you want changed before it can be merged in.

~~Also it seems like storybooks don't work that well with the new input/output API that the sidebar component uses, especially `model`, but maybe there is a way to mitigate that?~~


libs/extensions/angular/sidebar-menu/src/animations/drop-down.animation.ts
```diff
@@ -0,0 +1,13 @@
+import { animate, AUTO_STYLE, style, transition, trigger } from '@angular/animations';
```
RasmusKjeldgaard, line 1: Consider adding @angular/animations as a peerDependency in `libs/extensions/angular/package.json`

libs/extensions/angular/sidebar-menu/src/components/menu-item-content/menu-item-content.component.scss
```diff
@@ -0,0 +1,32 @@
...
+
+.menu-item-end {
+  display: inline-flex;
+  align-items: center;
+  height: utils.icon-font-size('sm');
+}
+
+.expanded-icon {
+  transform: rotate(90deg) !important;
+}
```
RasmusKjeldgaard: Now that we are at it, consider aligning this with current icon animation in kirby icon (and get rid of the unneeded !important statement):

```
kirby-icon {
  transition: transform utils.get-transition-duration('quick');
  &.expanded-icon {
    transform: rotate(90deg);
  }
}

```

libs/extensions/angular/sidebar-menu/src/components/menu-item-list/menu-item-list.component.scss
```diff
@@ -0,0 +1,66 @@
...
+    font-weight: utils.font-weight('bold');
+  }
+}
+
+kirby-divider {
+  display: block;
+  width: 100%;
+  opacity: 0.2;
+  margin: 8px -8px;
+}
```
RasmusKjeldgaard, line 67: If replaced with this, it will actually match both correct color from the the design spec and the expected 8px margin all around the divider (current divider spacing seems like a bug):

```suggestion
kirby-divider {
  --kirby-divider-color: var(--kirby-white-overlay-20);
}
```

libs/extensions/angular/sidebar-menu/src/public-components/sidebar-header/sidebar-header.component.ts
```diff
@@ -0,0 +1,17 @@
+import { Component } from '@angular/core';
```
RasmusKjeldgaard, line 1: Consider just doing `.component.ts` for these file names since they are already in a public-components folder.

RasmusKjeldgaard (commented):
Excellent work! 👏

libs/extensions/angular/sidebar-menu/src/animations/drop-down.animation.ts
Jonaythaz: done

libs/extensions/angular/sidebar-menu/src/components/menu-item-content/menu-item-content.component.scss
Jonaythaz: done

libs/extensions/angular/sidebar-menu/src/components/menu-item-list/menu-item-list.component.scss
Jonaythaz: I don't know why it was styled like this, but I have a branch ready for a follow-up PR with the new design where this is fixed, so I would prefer to address this problem there instead

libs/extensions/angular/sidebar-menu/src/public-components/sidebar-header/sidebar-header.component.ts
Jonaythaz: done

RasmusKjeldgaard (approved):

PR #4215, opened 2026-01-13
Accordion: Convert accordion to web component

### LINKED ISSUE #4148: [WebComponent] Convert Accordion to web component

Convert Accordion component to web component

<hr />

## Checklist:

### Refinement

- [ ] Request that the issue is [UX refined](https://github.com/kirbydesign/designsystem/blob/develop/.github/CONTRIBUTING.md#ux-refinement); do not proceed until this is done.
- [ ] Request that the issue is [tech refined](https://github.com/kirbydesign/designsystem/blob/develop/.github/CONTRIBUTING.md#tech-refinement); do not proceed until this is done.

### Implementation 

### Review

Loujuu:

This PR closes #4148

## What is the new behavior?

The main challenge in converting the Accordion component into a Lit web component was ensuring it could still work seamlessly with the existing Angular components. In this case Icon and List.

To solve this, I used two approaches:
1. Angular Proxy Components
Angular proxy components were generated to wrap the new Lit Accordion component. These proxies allow Angular applications to continue using the Accordion component through standard Angular imports, while the underlying implementation is now handled by Lit.
2. Angular Elements for Icon
For the Angular Icon component, I used Angular Elements to wrap it and expose it as a real web component.
Registering it as a web component in provide-kirby.ts so it can be used anywhere.
3. Base Class for List Support 
The List component is implemented as a base component to the Angular proxy component.  I have updated the generate-angular-proxies script so it can automatically generate these base files when needed.

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


libs/designsystem/config/src/provide-kirby.ts
```diff
@@ -70,6 +73,18 @@ export function provideKirby(
     CanDismissHelper,
     ...patchIonicProviders(),
     features,
+    {
```
RasmusKjeldgaard: This initializer is only run if the token is actually injected, so effectively it is only registered, not executed, if the consuming application does not inject it.

Instead, I believe this should be an environment initializer that is executed when the application is bootstrapped.

```
provideEnvironmentInitializer(() => {
  const injector = inject(Injector);
  if (!customElements.get('kirby-icon')) {
    const iconElement = createCustomElement(IconComponent, { injector });
    customElements.define('kirby-icon', iconElement);
  }
}),
```

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,106 @@
...
+    return html`
+      <div class="content-layer" aria-disabled=${this.isDisabled ? 'true' : 'false'}>
+        <div
+          aria-level=${this.headingLevel ?? nothing}
+          role=${this.headingLevel ? 'heading' : nothing}
+        >
+          <button
+            type="button"
+            class="header"
+            aria-expanded=${String(this.isExpanded)}
```
Fuzzy3: aria-expanded expects boolean, so remove the string casting 
```suggestion
            aria-expanded=${this.isExpanded}
```

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,106 @@
+import { CSSResultGroup, html, nothing } from 'lit';
+import { property } from 'lit/decorators.js';
+import { classMap } from 'lit/directives/class-map.js';
+import KirbyElement from '../internal/kirby-element';
+import baseStyles from '../styles/element.base.styles';
+import styles from './accordion-item.element.styles';
+
+// Counter for generating unique element ids
+let uniqueId = 0;
```
Fuzzy3, line 9: Probably should use the unique-id-generator.helper.ts, but since this is not available in the core package. this is fine for now. Consider adding a comment that this should be using the id-generator when its available.

libs/designsystem/config/src/provide-kirby.ts
```diff
@@ -70,6 +73,13 @@ export function provideKirby(
     CanDismissHelper,
     ...patchIonicProviders(),
     features,
+    provideEnvironmentInitializer(() => {
+      const injector = inject(Injector);
+      if (!customElements.get('kirby-icon-element')) {
+        const iconElement = createCustomElement(IconComponent, { injector });
+        customElements.define('kirby-icon-element', iconElement);
```
Fuzzy3, line 80: is there a reason we add the "...-element" part to the kirby-icon ?

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,106 @@
...
+            type="button"
+            class="header"
+            aria-expanded=${String(this.isExpanded)}
+            aria-controls=${this._contentId}
+            id=${this._titleId}
+            ?disabled=${this.isDisabled}
+            @click=${this._onToggleExpanded}
+          >
+            <span class="state-layer" aria-hidden="true"></span>
+            <div class="title" ?bold=${this.isExpanded}>${this.getTitle()}</div>
```
Fuzzy3: I can't figure out where this bold-attribute is used. Maybe it can be removed?

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,106 @@
...
+          </button>
+        </div>
+        <div
+          class="content"
+          role="region"
+          aria-labelledby=${this._titleId}
+          id=${this._contentId}
+          ?expanded=${this.isExpanded}
+        >
+          <div class=${classMap({ 'content-body': true, padding: this.hasPadding })}>
```
Fuzzy3: no need for content-body in a classMap if its always true. Just put it in as a static css class.

Fuzzy3: Also instead of using classMap for css classes, use the attribute hasPadding which is applied when true, to style. I believe this is the way we wanna style LitElements.
In your css you probably then need to change the hasPadding styling to something like this:
```
:host([hasPadding]) .content-body {
  ...the padding style here
}
```

libs/angular/accordion-item/src/accordion-item-base.component.ts
```diff
@@ -0,0 +1,47 @@
+import { AfterContentInit, ContentChildren, Directive, OnChanges, QueryList } from '@angular/core';
+import { ListComponent } from '@kirbydesign/designsystem/list';
+
+@Directive()
```
Fuzzy3, line 5: Add comment about this component being temporary until the list component is also converted to a lit element

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  updated(changedProps: Map<string, unknown>) {
+    if (changedProps.has('isDisabled') && this.isDisabled) {
+      this.isExpanded = false;
+    }
+  }
+
+  render() {
+    return html`
+      <div class="content-layer" aria-disabled=${this.isDisabled ? 'true' : 'false'}>
```
Fuzzy3:
```suggestion
      <div class="content-layer" aria-disabled=${this.isDisabled}>
```

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,115 @@
...
+              <kirby-icon-element name="arrow-down"></kirby-icon-element>
+            </span>
+          </button>
+        </div>
+        <div
+          class="content"
+          role="region"
+          aria-labelledby=${this._titleId}
+          id=${this._contentId}
+          ?expanded=${this.isExpanded}
```
Fuzzy3: This expanded attribute seems redundant. The isExpanded attribute is already applied to the hostelement when this is true, so I believe we should use that instead of applying this "expanded" attribute to a content element.

Fuzzy3 (changes requested):
Really nice work on this so far ! One thing I think we should do when converting to LitElements, is style based on the reflected attributes on the host element instead of styling using dynamic classes. 

Also I can see that the accordion-item was converted to a LitElement, but what about the accordion directive? That directive has some heading logic which applies the same heading level to all of the underlying accordion-items. Is that part of a later PR ?

I don't have a strong opinion on the generate-proxies script, so I haven't looked too deeply into that. But I think its important that we make some code comments about the temporary code this PR contains - specifically the accordion base class

libs/designsystem/config/src/provide-kirby.ts
Loujuu: The reason is, that it is important to not use the components selector as custom element tag name, since it can lead to unexpected behaviour. https://angular.dev/guide/elements#transforming-components-to-custom-elements

libs/core/src/accordion-item/accordion-item.element.ts
Loujuu: You are right, it can be removed. The css uses button[aria-expanded='true'] .title to apply bold.

libs/core/src/accordion-item/accordion-item.element.ts
Loujuu: I do need classmap for has padding. But I have changed to static content-body.  <div class="content-body ${classMap({ padding: this.hasPadding })}">

Loujuu: Makes sense to remove classMap and add haspadding in styling instead

libs/angular/accordion-item/src/accordion-item-base.component.ts
```diff
@@ -0,0 +1,48 @@
+// This file is temporary until the list component is also converted to a lit element and this base component can be removed.
```
Fuzzy3: Move this comment down to right before @directive()

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,109 @@
...
+  @property({ type: Boolean, reflect: true }) isDisabled = false;
+  @property({ type: String }) disabledTitle = '';
+  @property({ type: Boolean, reflect: true }) hasPadding = true;
+  @property({ type: Number }) headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
+  @property({ type: String }) shape = 'none';
+  @property({ type: Boolean, reflect: true }) inCard = false;
+  private _titleId = `kirby-accordion-item-title-${++uniqueId}`;
+  private _contentId = `kirby-accordion-item-content-${uniqueId}`;
+
+  getTitle() {
```
Fuzzy3:
```suggestion
  private getTitle() {
```

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (this.isDisabled && !!this.disabledTitle) {
+      return this.disabledTitle;
+    } else {
+      return this.title;
+    }
+  }
+
+  _onToggleExpanded(event: UIEvent) {
```
Fuzzy3:
```suggestion
  private onToggleExpanded(event: UIEvent) {
```

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,109 @@
...
+          role=${this.headingLevel ? 'heading' : nothing}
+        >
+          <button
+            type="button"
+            class="header"
+            aria-expanded=${this.isExpanded}
+            aria-controls=${this._contentId}
+            id=${this._titleId}
+            ?disabled=${this.isDisabled}
+            @click=${this._onToggleExpanded}
```
Fuzzy3:
```suggestion
            @click=${this.onToggleExpanded}
```

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,109 @@
...
+  getTitle() {
+    if (this.isDisabled && !!this.disabledTitle) {
+      return this.disabledTitle;
+    } else {
+      return this.title;
+    }
+  }
+
+  _onToggleExpanded(event: UIEvent) {
+    event.preventDefault();
```
Fuzzy3: Probably doesn't do anything since a button doesn't have a "default action"

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,109 @@
...
+  static styles: CSSResultGroup = [baseStyles, styles];
+
+  @property({ type: String }) title = '';
+  @property({ type: Boolean, reflect: true }) isExpanded = false;
+  @property({ type: Boolean, reflect: true }) isDisabled = false;
+  @property({ type: String }) disabledTitle = '';
+  @property({ type: Boolean, reflect: true }) hasPadding = true;
+  @property({ type: Number }) headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
+  @property({ type: String }) shape = 'none';
+  @property({ type: Boolean, reflect: true }) inCard = false;
```
Fuzzy3: The inCard property was introduced here since "host-context" was no longer an option in css and outside of angular. Instead of introducing this as a property and using `closest()`in the `connectedCallback`. 
As a general strategy for all places where we use host-context, I think the cleanest solution is to use CSS custom properties. So in this case we need to introduce some 
css custom properties that may then be set from outside (in this case, the card component)

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,109 @@
...
+      new CustomEvent('toggle', {
+        detail: this.isExpanded,
+        bubbles: true,
+        composed: true,
+      })
+    );
+  }
+
+  //detect card and add class to prevent double borders when used inside card
+  connectedCallback() {
```
Fuzzy3: Remove and use CSS custom properties

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  firstUpdated() {
+    const slot = this.renderRoot.querySelector('slot');
+    if (slot) {
+      slot.addEventListener('slotchange', () => {
+        const assigned = slot.assignedElements({ flatten: true });
+        assigned.forEach((el) => {
+          if (el.tagName === 'KIRBY-LIST') {
```
Fuzzy3: Remember to also set `hasPadding` to false if there is a list. Also probably you need to introduce a hasList property and style using that.

libs/core/src/accordion-item/accordion-item.element.styles.ts
```diff
@@ -0,0 +1,170 @@
...
+    transition: transform var(--kirby-transition-quick);
+    font-size: var(--kirby-icon-font-size, 24px);
+    width: 24px;
+    height: 24px;
+    display: block;
+  }
+
+  .content {
+    overflow: hidden;
+    transition:
```
Fuzzy3: I guess this transition is introduced to solve that we can no longer depend on angular doing the animation. However this doesn't seem to work as the animation is no longer there.

libs/core/src/accordion-item/accordion-item.element.styles.ts
```diff
@@ -0,0 +1,170 @@
...
+    background: none;
+    box-shadow: none;
+    font-family: var(--kirby-font-family);
+    font-size: inherit;
+    line-height: normal;
+    text-align: start;
+  }
+
+  .title {
+    flex-grow: 2;
```
Fuzzy3: why flex-grow: 2;? could it just be 1?

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  static styles: CSSResultGroup = [baseStyles, styles];
+
+  @property({ type: String }) title = '';
+  @property({ type: Boolean, reflect: true }) isExpanded = false;
+  @property({ type: Boolean, reflect: true }) isDisabled = false;
+  @property({ type: String }) disabledTitle = '';
+  @property({ type: Boolean, reflect: true }) hasPadding = true;
+  @property({ type: Number }) headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
+  @property({ type: String }) shape = 'none';
```
Fuzzy3: Why was this introduced? Accident ?

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -0,0 +1,109 @@
...
+            el.setAttribute('shape', 'none');
+          }
+        });
+      });
+    }
+  }
+
+  updated(changedProps: Map<string, unknown>) {
+    if (changedProps.has('isDisabled') && this.isDisabled) {
+      this.isExpanded = false;
```
Fuzzy3, line 45: updating a property in the updated cycle, will trigger a second render cycle. I think its possible to use the `willUpdate()` callback instead to avoid this.

Fuzzy3 (changes requested):
Looking better and better 😸 
I just found that the animation broke, so we need to fix that ! also i dont think that the list in accordion logic is completely like the old one. - see my comments in the code. 
We are close !

libs/core/src/accordion-item/accordion-item.element.ts
Loujuu: This makes sense. I have removed the InCard property and replaced it with css custom properties inn the card component.

libs/core/src/accordion-item/accordion-item.element.styles.ts
Loujuu: For a smooth effect like Angular’s, you need to animate between numeric heights using JavaScript. I have applied this instead and removed the transition in css.

Loujuu: Update: Not using JS instead using interpolate-size for smooth expand/colapse in Chrome and a fallback with grid in Safari.

libs/core/src/accordion-item/accordion-item.element.styles.ts
```diff
@@ -94,7 +90,44 @@ export default css`
...
+    }
+
+    .content-body {
+      overflow: hidden;
+      min-height: 0;
+    }
+  }
+
+  /* Chrome / Edge */
+  @supports (interpolate-size: allow-keywords) {
```
Fuzzy3: Maybe this "@supports" -line may be removed, so this is part of the base styling?

libs/designsystem/card/src/card.component.scss
```diff
@@ -9,6 +9,9 @@ $default-border-outline: 1px solid utils.get-color('medium');
   --kirby-card-main-color: #{utils.get-color('white-contrast')};
   --kirby-card-footer-background-color: #{utils.get-color('white')};
   --kirby-card-footer-color: #{utils.get-color('white-contrast')};
+  --kirby-accordion-item-border-color: #{utils.get-color('background-color')};
```
Fuzzy3, line 12: Could these three lines be reduced to just --kirby-accordion-item-border?

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -46,29 +45,22 @@ export class KirbyAccordionItemElement extends KirbyElement {
...
-        const assigned = slot.assignedElements({ flatten: true });
-        assigned.forEach((el) => {
-          if (el.tagName === 'KIRBY-LIST') {
-            el.setAttribute('shape', 'none');
-          }
-        });
+      const assigned = slot.assignedElements({ flatten: true });
+      const hasList = assigned.some((el) => el.tagName === 'KIRBY-LIST');
+      this.hasList = hasList;
+      this.hasPadding = !hasList;
```
Fuzzy3: Only set hasPadding when a list has been found, otherwise it will always overwrite the value even if the consumer set this.

libs/designsystem/card/src/card.component.scss
Fuzzy3: or maybe two vars?

Fuzzy3 (changes requested):
Almost there ! I just have a very few nitty grittys, and then its ready !!

libs/core/src/accordion-item/accordion-item.element.styles.ts
Loujuu: You are right. Not needed :-)

libs/core/src/accordion-item/accordion-item.element.ts
Loujuu: You are right :-)

libs/core/src/accordion-item/accordion-item.element.ts
```diff
@@ -51,7 +50,9 @@ export class KirbyAccordionItemElement extends KirbyElement {
       const assigned = slot.assignedElements({ flatten: true });
       const hasList = assigned.some((el) => el.tagName === 'KIRBY-LIST');
       this.hasList = hasList;
-      this.hasPadding = !hasList;
+      if (hasList) {
```
Fuzzy3: also put this.hasList and the forEach loop inside the if

Fuzzy3 (changes requested):

PR #4302, opened 2026-02-23
Combobox: Component added

### LINKED ISSUE #4359: [COMPONENT] Add Multi Select Autocomplete Component
Kirby lacks a component that enables the user to search in a dropdown

A new component that supports search / filtering in a dropdown.
This way a user does not need to scroll if the list is very long.

[Link to sketch](https://www.sketch.com/s/e2ab5f35-e31d-4b01-b6ed-c609fac87e27#Inspect)

We have looked at the already existing core components, and they do not support this feature.

<hr />

## Checklist:

### Refinement

- [x] Request that the issue is [UX refined](https://github.com/kirbydesign/designsystem/blob/develop/.github/CONTRIBUTING.md#ux-refinement); do not proceed until this is done.
- [x] Request that the issue is [tech refined](https://github.com/kirbydesign/designsystem/blob/develop/.github/CONTRIBUTING.md#tech-refinement); do not proceed until this is done.

### Implementation 

- [x] Make sure you have read: "[Before you get coding](../CONTRIBUTING.md/#before-you-get-coding)".
- [x] Create a branch from the [develop branch](https://github.com/kirbydesign/designsystem/tree/develop) following our [branch naming convention](https://github.com/kirbydesign/designsystem/wiki/The-Good%3A-Branch). 
- [x] Update the [cookbook](https://cookbook.kirby.design) with examples and showcases.

### Review

HootERasmus:

This PR closes #4359 

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


apps/cookbook/src/app/examples/item-example/examples/sizes.ts
Fuzzy3, line 1: We need to revert all the files that were only changed because of the "file mode" that changed. Could this be a classic mac vs windows problem, where shifting os changes something in the files?

libs/designsystem/form-field/src/form-field.component.html
```diff
@@ -50,6 +50,7 @@
     <ng-content select="textarea[kirby-textarea]"></ng-content>
     <ng-content select="kirby-radio-group"></ng-content>
     <ng-content select="kirby-dropdown"></ng-content>
+    <ng-content select="kirby-multi-select-autocomplete"></ng-content>
```
Fuzzy3: If we want the multi-select working in a kirby form field, there is more work to be done in the form-field.component.ts codebehind. Consider doing this as part of a second PR, so this PR focusses on the multi select component only.

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+    There is a bug that causes the dropdown to get a focus ring on click, if it is the first element that is interacted with
+    after the page is loaded. If the user interacts with any other element before, then the dropdown won't get a focus ring.
+    See issue: https://github.com/kirbydesign/designsystem/issues/2477.
+
+    This solution can potentially be refactored, when popover is not experimental anymore. Then it could be possible
+    to close the dropdown when the popover backdrop is clicked, instead of relying on the blur event, which is utilized
+    by this line below: this.elementRef.nativeElement.focus(). Right now this forces the blur event to be triggered, when
+    clicking outside of the dropdown.
+    */
+  @HostBinding('class.clicked')
```
Fuzzy3: Have you investigated if this is still needed?

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.ts
```diff
@@ -0,0 +1,752 @@
...
+    InputComponent,
+    NgTemplateOutlet,
+    PopoverComponent,
+    CardComponent,
+    ItemComponent,
+    IconComponent,
+    AffixDirective,
+  ],
+})
+export class MultiSelectAutocompleteComponent
```
Fuzzy3: I guess the "Autocomplete" is just a "feature" to this component. Consider renaming to MultiSelectComponent or ComboBoxComponent

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+  @Input()
+  public set focusedItem(item: unknown) {
+    if (this._focusedItem !== item) {
+      this._focusedItem = item;
+      this.scrollFocusedItemIntoView();
+    }
+  }
+
+  @Input()
+  public placeholder = 'Please search...';
```
Fuzzy3, line 290: Be carefull with "default text"/"harded coded text". This could end up showing english text in other language-scenarios (danish, german)

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+  @ViewChild(PopoverComponent)
+  public popover?: PopoverComponent;
+
+  @ViewChild('rootElement', { static: true, read: ElementRef })
+  public rootElement!: ElementRef<HTMLElement>;
+
+  @ViewChild(InputComponent, { read: ElementRef })
+  private textInput?: ElementRef<HTMLInputElement>;
+
+  private forwardAriaLabelToDropdownButton() {
```
Fuzzy3: Rename function so it matches the components name (this is not a dropdown component 😉 )

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+      value,
+      RendererStyleFlags2.DashCase
+    );
+  }
+
+  public ngOnDestroy(): void {
+    this.unlistenAllSlottedItems();
+  }
+
+  private ensureComponents(): void {
```
Fuzzy3: In general we would like all controls to be working both in the context of a form-field and "outside" the context of a form-field

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+    const currentIndex = this.searchItems.indexOf(this.focusedItem);
+    let newIndex;
+
+    if (numberOfItems > 0) {
+      newIndex = Math.min(this.searchItems.length - 1, currentIndex + numberOfItems);
+    } else {
+      newIndex = Math.max(0, currentIndex + numberOfItems);
+    }
+
+    console.log('newIndex', newIndex);
```
Fuzzy3: Remove

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+    // object is complex, and the user uses custom template, so we rely on itemIdProperty
+    if (item && typeof item === 'object' && this.itemIdProperty in item) {
+      return String((item as Record<string, unknown>)[this.itemIdProperty]);
+    }
+
+    // object is complex, and the user uses custom template, but itemIdProperty is missing
+    throw new Error(
+      'Each item must have an id property for scroll to work. Ensure that the itemIdProperty input is set correctly, and that each item has a unique id value.'
+    );
+  }
```
Fuzzy3, line 198: `itemIdProperty` and `getitemId()` seems unused except for in the `scrollFocusedItemIntoView()`-function. Consider removing this al-together and in the scroll-function just find the item by using the index of the item that needs to be viewed.

libs/extensions/angular/combobox/src/combobox.component.ts
Fuzzy3: Never mind - we do the same in our dropdown component..

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.ts
Fuzzy3: I think we should name this component a "ComboboxComponent"

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.ts
```diff
@@ -0,0 +1,752 @@
...
+import { AffixDirective, InputSize } from '@kirbydesign/designsystem/form-field';
+import { EventListenerDisposeFn } from '@kirbydesign/designsystem/types';
+import { forwardAttributes, ResizeObserverService } from '@kirbydesign/designsystem/shared';
+import { InputComponent } from '@kirbydesign/designsystem/form-field';
+import { IconComponent } from '@kirbydesign/designsystem/icon';
+import { NgTemplateOutlet } from '@angular/common';
+import { OpenState } from './multi-select-autocomplete.types';
+
+@Component({
+  selector: 'kirby-multi-select-autocomplete',
```
Fuzzy3: all extension components have "-x" in the selector.
```suggestion
  selector: 'kirby-x-combobox',
```

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.ts
```diff
@@ -0,0 +1,752 @@
...
+  templateUrl: './multi-select-autocomplete.component.html',
```
Fuzzy3: rename all files to "combobox" instead of "multi-select-autocomplete.

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.ts
```diff
@@ -0,0 +1,752 @@
...
+})
+export class MultiSelectAutocompleteComponent
+  implements OnInit, AfterViewInit, OnDestroy, ControlValueAccessor
+{
+  static readonly OPEN_DELAY_IN_MS = 100;
+  private state = OpenState.closed;
+  private _popout: HorizontalDirection | `${HorizontalDirection}` = HorizontalDirection.right;
+  private _attributesToForward = ['aria-label', 'aria-labelledby'];
+
+  public dropdownId: string = UniqueIdGenerator.scopedTo('kirby-dropdown').next();
```
Fuzzy3:
```suggestion
  public comboboxId: string = UniqueIdGenerator.scopedTo('kirby-x-combobox').next();
```

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+    }
+
+    const lowerSearchTerm = searchTerm.toLowerCase();
+    return this.items.filter((item) =>
+      this.getItemText(item).toLowerCase().includes(lowerSearchTerm)
+    );
+  };
+
+  @Input()
+  public noSearchResultsText = 'Ingen resultater';
```
Fuzzy3: Should be english

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+
+  @Input()
+  public itemTextProperty: string = 'text';
+  protected getItemText(item: unknown): string {
```
Fuzzy3, line 124: This function seems more complicated than its equal in dropdown. I understand that it probably has something to do with the need to type-cast due to "unknown"-type, however consider if theres a more "readable" way of doing this.

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+
+  // _focusedItem keeps track of which element has focus and will be selected
+  // if it is activated (by pressing ENTER)
+  private _focusedItem: unknown = undefined;
+  get focusedItem(): unknown {
+    return this._focusedItem;
+  }
+
+  @Input()
+  public set focusedItem(item: unknown) {
```
Fuzzy3, line 252: focusedItem should probablty not be an Input()-property as it doesn't really make sense to allow integrators to decide where the focus should be.

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+    if (this._hasError !== value) {
+      this._hasError = value;
+      this.hasErrorChange.emit(this._hasError);
+    }
+  }
+
+  @Input()
+  public size: InputSize = InputSize.medium;
+
+  // Tabindex defaults to 0 instead of no value to align tabbing behavior in Safari
```
Fuzzy3: I dont think this is true though, since you are using an input and not a button in the html. I know we do this in the dropdown, but that is to make sure the dropdown is tab-able in safari because it uses a button. However I think safari does always make inputs tab-able by default. Please remove this tab-index and check if the combobox is still tab-able in safari

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.html
```diff
@@ -0,0 +1,63 @@
+<div (click)="onToggle($event)" #rootElement>
+  <input
+    kirby-input
+    type="text"
+    [attr.id]="inputId"
+    aria-haspopup="listbox"
+    [attr.aria-controls]="dropdownId"
```
Fuzzy3: change to comboboxId

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.html
```diff
@@ -0,0 +1,63 @@
...
+    [attr.aria-expanded]="isOpen"
+    [attr.aria-activedescendant]="isOpen ? dropdownId + '-item-' + getItemText(focusedItem) : null"
```
Fuzzy3:
```suggestion
    [attr.aria-activedescendant]="isOpen ? comboboxId + '-item-' + getItemText(focusedItem) : null"
```

Fuzzy3: Be careful with having function calls in an attibute: `getItemText(focusedItem)` it might affect performance as we don't know how many times this will be called. Might not be a problem, but consider using index instead

libs/extensions/angular/combobox/src/combobox.component.html
```diff
@@ -0,0 +1,63 @@
...
+    type="text"
+    [attr.id]="inputId"
+    aria-haspopup="listbox"
+    [attr.aria-controls]="dropdownId"
+    [attr.aria-expanded]="isOpen"
+    [attr.aria-activedescendant]="isOpen ? dropdownId + '-item-' + getItemText(focusedItem) : null"
+    [hasError]="hasError"
+    [size]="size"
+    [disabled]="disabled"
+    [tabindex]="_tabindex"
```
Fuzzy3: might not be relevant - see other comment about tabindex

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.html
```diff
@@ -0,0 +1,63 @@
...
+    [hasError]="hasError"
+    [size]="size"
+    [disabled]="disabled"
+    [tabindex]="_tabindex"
+    [placeholder]="placeholder"
+    (input)="onInput($event)"
```
Fuzzy3: consider more readable naming of function. maybe "updateSearchResults($event)" instead of just "onInput" so its more descriptive about what happens on input

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.html
```diff
@@ -0,0 +1,63 @@
...
+  <kirby-icon [name]="isOpen ? 'arrow-up' : 'arrow-down'" kirby-affix="suffix"></kirby-icon>
+</div>
+
+<kirby-popover
+  [target]="rootElement"
+  [popout]="popout"
+  (click)="onPopoverClick()"
+  (willHide)="onPopoverWillHide()"
+>
+  <kirby-card role="listbox" [attr.id]="dropdownId" [attr.aria-labelledby]="inputId">
```
Fuzzy3: I think we should just use the same naming as we use in the dropdown. So the card is the listbox and has id "_listboxId" while the trigger button (in here the trigger-input) should have the id "_comboboxId" and then the aria-labelledBy should point at "_comboboxId"

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.html
```diff
@@ -0,0 +1,63 @@
...
+</kirby-popover>
+
+<ng-template #defaultItemTemplate let-item let-selected="selected" let-focused="focused">
+  <kirby-item
+    [selectable]="true"
+    [selected]="selected"
+    (click)="onItemSelect(item)"
+    [class.focused]="focused"
+    role="option"
+    [attr.id]="dropdownId + '-item-' + getItemText(item)"
```
Fuzzy3: same comments as the attr.id in the input

libs/extensions/angular/combobox/src/combobox.component.stories.ts
```diff
@@ -0,0 +1,270 @@
...
+    placeholder: 'Select currencies but very much longer',
+    disabled: false,
+    hasError: false,
+  },
+};
+
+export const WithTextProperty: Story = {
+  args: {
+    items: currencyItems,
+    itemTextProperty: 'code',
```
Fuzzy3, line 121: consider using 'name' instead of 'code' here, so it doesn't look like the default example

libs/extensions/angular/combobox/src/combobox.component.stories.ts
Fuzzy3, line 1: We need an overall docs page to be explaining the usage cases and different attributes. Consider looking at the other components in extensions to get inspired 😄 https://extensions.kirby.design/

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+    }
+
+    return this.items.find((it) => this.getItemText(it) === input) ?? input;
+  }
+
+  private setInputDisplayValue(value: string): void {
+    // We intentionally update the DOM input imperatively ("hard way")
+    // to ensure the visible value is cleared/updated immediately.
+    if (this.textInput?.nativeElement) {
+      this.textInput.nativeElement.value = value;
```
Fuzzy3: consider using angular's renderer2 for dom manipulation

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,752 @@
...
+  public ngOnDestroy(): void {
+    this.unlistenAllSlottedItems();
+  }
+
+  private ensureComponents(): void {
+    if (!this.rootElement) {
+      throw new Error('requires <kirby-form-field> element to function properly');
+    }
+
+    // if (!this.itemIdProperty) {
```
Fuzzy3: Commented out

Fuzzy3 (changes requested):
Really nice initial work on the combobox (as I would call it). 😄 
I like that you took a lot of inspiration from the dropdown, which makes it easier to understand. However I've also noticed that the word "dropdown" snuck in a lot of places (i guess due to the copy/paste of the dropdown code). I suggest to read through the code and replace the dropdown-word with "combobox" (same goes for variations of the "multiselect"-word). 

I have smaller and larger comments, some are more conceptual than others - please reach out if you are in doubt. 🤓 

I'd like if you spent some time on the documentation in storybook such that integrators have a better understanding of the intention of this component and how to use it. 📖 

One thing I noticed is that this should exist within the context of a form field, which I like. But consider having a later task for making that part work. In general our controls should work both inside and outside of a form field context. 
I see that you have decided to put this component in a folder named "forms", in storybook. Let's for now, just have it at root level and name it "combobox".

One thing we need to be sure of is of proper quality is the "input"-fields as these are VERY hard to change later on, since changes to those after the first release of this component, would be breaking changes. (API-changes) 

Overall its a really well work and looking forward to this component into our extension library 🔥 🚀

apps/cookbook/src/app/examples/item-example/examples/sizes.ts
HootERasmus: yeap that was a mac thing

libs/designsystem/form-field/src/form-field.component.html
HootERasmus: Lets do it in two then, if there are more to it

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.html
HootERasmus: yeah copy paste mistake

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.html
HootERasmus: same as above

HootERasmus: we do though see some issues by using index, as when you search the list shown in the popover changes, and the index range just decreases instead of being filtered

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.html
HootERasmus: I dig that

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.html
HootERasmus: sure, we like consistency

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.html
HootERasmus: sure, we like consistency

libs/extensions/angular/combobox/src/combobox.component.stories.ts
HootERasmus: Added documentation and an example for the search function

libs/extensions/angular/combobox/src/combobox.component.stories.ts
HootERasmus: this is intentional, i like to show only one new feature on each example

libs/extensions/angular/multi-select-autocomplete/src/multi-select-autocomplete.component.ts
HootERasmus: yeah combobox it is

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: Yeap

libs/extensions/angular/combobox/src/combobox.component.html
HootERasmus: yeap not needed

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: yeah this is because of the type 'unknown', it forces us to handle the scenarios instead of exploding.
I have tried to refactor it a bit

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: Yeah i get that, so far i have not gotten a way to filter on the 'kirbyItems' as they are the dom elements we are looking at.

So i am all open to suggestions

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: true

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: looks like it is not need any more now that we are using input instead

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: yeah consistency

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: This was for some older code, now rootElement is the component itself, so this is just for safety because of the use of ! (bang) on rootElement on initialization

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: removed old debug logs

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: yeah good idea here

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: still tabable

libs/extensions/angular/combobox/src/combobox.component.html
```diff
@@ -0,0 +1,81 @@
...
+  <kirby-icon [name]="isOpen ? 'arrow-up' : 'arrow-down'" kirby-affix="suffix"></kirby-icon>
+</div>
+
+<kirby-popover
+  [target]="rootElement"
+  [popout]="popout"
+  (click)="onPopoverClick()"
+  (willHide)="onPopoverWillHide()"
+>
+  <kirby-card role="listbox" [attr.id]="_listboxId" [attr.aria-labelledby]="_comboboxId">
```
Fuzzy3: role="listbox" should probably be moved to cdk-virtual-scroll-viewport since this is now wrapping the kirby-items

libs/extensions/angular/combobox/src/combobox.component.html
```diff
@@ -0,0 +1,81 @@
...
+    [attr.id]="_comboboxId"
+    aria-haspopup="listbox"
+    [attr.aria-controls]="_listboxId"
+    [attr.aria-expanded]="isOpen"
+    [attr.aria-activedescendant]="isOpen && focusedItem ? getItemId(focusedItem) : null"
+    [hasError]="hasError"
+    [size]="size"
+    [disabled]="disabled"
+    [placeholder]="placeholder"
+    (input)="updateSearchResults($event)"
```
Fuzzy3, line 15: Should probably have `role="combobox"`

libs/extensions/angular/combobox/src/combobox.module.ts
```diff
@@ -0,0 +1,30 @@
...
+import { CardModule } from '@kirbydesign/designsystem/card';
+import { IconModule } from '@kirbydesign/designsystem/icon';
+import { ItemModule } from '@kirbydesign/designsystem/item';
+import { PopoverComponent } from '@kirbydesign/designsystem/popover';
+
+import { ListItemTemplateDirective, ListModule } from '@kirbydesign/designsystem/list';
+import { FormFieldModule, InputComponent } from '@kirbydesign/designsystem/form-field';
+import { ComboboxComponent } from './combobox.component';
+
+@NgModule({
```
Fuzzy3: Probably no need for an NgModule - combobox is standalone

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,852 @@
...
+    NgTemplateOutlet,
+    PopoverComponent,
+    CardComponent,
+    ItemComponent,
+    IconComponent,
+    AffixDirective,
+    CdkFixedSizeVirtualScroll,
+    CdkVirtualScrollViewport,
+    CdkVirtualForOf,
+  ],
```
Fuzzy3, line 64: Consider OnPush changedetection strategy

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,852 @@
...
+      value,
+      RendererStyleFlags2.DashCase
+    );
+  }
+
+  public ngOnDestroy(): void {
+    this.unlistenAllSlottedItems();
+  }
+
+  private ensureComponents(): void {
```
Fuzzy3: `ensureComponents()` may be removed entirely since the rootElement is decorated with `{static: true}` which is resolved before ngOnInit, so this function does nothing.

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,852 @@
...
+    if (!this.focusedItem) {
+      return;
+    }
+
+    const id = this.getItemId(this.focusedItem);
+    if (!id) return;
+
+    const match = kirbyItems.toArray().find((el) => {
+      if (el.nativeElement.id === undefined) {
+        throw new Error(
```
Fuzzy3: Throwing errors will crash the whole view. Consider using console.error instead

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,852 @@
...
+    this.renderer.setStyle(
+      this.cardElement?.nativeElement,
+      style,
+      value,
+      RendererStyleFlags2.DashCase
+    );
+  }
+
+  public ngOnDestroy(): void {
+    this.unlistenAllSlottedItems();
```
Fuzzy3, line 482: missing function call to unobserve resizeObserver, to avoid memory leaks

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,852 @@
...
+
+  @HostListener('keydown.tab')
+  public onTab() {
+    if (this.isOpen) {
+      this.selectItem(this.focusedItem);
+      this.close();
+    }
+  }
+
+  @HostListener('keydown.enter')
```
Fuzzy3: duplicate "keydown.enter"

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,852 @@
...
+  @HostListener('keydown.escape')
+  public onEnterOrEscape() {
+    this.close();
+  }
+
+  @HostListener('keydown.enter', ['$event'])
+  public onEnterOrSpace(event: Event) {
```
Fuzzy3: duplicate "keydown.enter" and name of function suggests that this is also triggered by space, which it is not

Fuzzy3 (changes requested):
Really nice work ! I think we are almost there - however lets see if we can improve the screen reader experience even more, maybe try with an aria-live that announces how many results have been found based on the search. 
Also it would be nice to add the necessary aria-attributes as a safe guard, to the custom item template, if consumers use their own template.

I see that an NgModule was introduced but since the component is actually stand-alone, I believe we can delete that entirely. 

Also change the changedetection strategy to be OnPush, since this component does handle quite a large amount of items. 

There are a few more inline code comments that needs to be fixed, and then this is good to go!

libs/extensions/angular/combobox/src/combobox.component.html
HootERasmus: you are right

libs/extensions/angular/combobox/src/combobox.component.html
HootERasmus: added

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: added

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: only had it to throw nice errors. Removing it

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: changed to console.error

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: good spottet

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: changed to only onEscape

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: changed to only onEnter

libs/extensions/angular/combobox/src/combobox.module.ts
HootERasmus: nope there is not

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,882 @@
...
+    return this._focusedItem;
+  }
+
+  public set focusedItem(item: unknown) {
+    if (this._focusedItem === item) {
+      return;
+    }
+    this._focusedItem = item;
+
+    const htmlElement = this.getKirbyHtmlElementForItem(item);
```
Fuzzy3:
```suggestion
    const itemElement = this.getKirbyItemElement(item);
```

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,882 @@
...
+    const index = this.getIndexOfItem(item);
+    const setsize = this.searchItems.length;
+    this.setAriaAttributeOnHtmlElement(htmlElement, 'aria-setsize', setsize.toString());
+    this.setAriaAttributeOnHtmlElement(htmlElement, 'aria-posinset', `${index + 1}`);
```
Fuzzy3: Consider calling the renderer instead, as this is currently just forwarding parameters to that

Fuzzy3: setsize and posinset needs to be calculated whenever the searchItems-list changes.

libs/extensions/angular/combobox/src/combobox.component.ts
```diff
@@ -0,0 +1,882 @@
...
+  }
+
+  private getIndexOfItem(item: unknown): number {
+    return this.searchItems.indexOf(item);
+  }
+
+  private getKirbyHtmlElementForItem(
```
Fuzzy3:
```suggestion
  private getKirbyItemElement(
```

Fuzzy3 (changes requested):

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: now only uses the name 'element'

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: has been renamed

libs/extensions/angular/combobox/src/combobox.component.ts
HootERasmus: Method has been removed, and all the logic been put into it's own method

HootERasmus: Moved setAriaPosinsetOnElement it the begning of focusedItem, to make sure the focused item always gets its updated aria-attributes

Fuzzy3 (approved):
Really awesome work on making this accessible !! this will be such a great addition to the designsystem 🚀 🧑‍🔬

PR #4371, opened 2026-03-19
fix(ci): update reference to renamed validator script

troelslenda:
Follow-up fix for https://github.com/kirbydesign/designsystem/pull/4368

During the migration of the workflow from a test repository, I renamed the validator file to align with the naming conventions used across the project. However, I missed updating the corresponding reference in the workflow configuration.

This PR corrects that oversight.

With the introduction of SVG validation, we also need to ensure that all SVGs pass validation. As a result, two SVGs were identified and corrected.


RasmusKjeldgaard (approved):
Very nice! LGTM 🦾

PR #4466, opened 2026-05-19
Workflow: update dependencies in actions

martinkolle:
Nodejs v.20 is deprecated in the Github Pipelines. This updates dependencies to versions that uses nodejs v.24

All versions is pinned to sha to prevent supply chain attach 

## Changes
- Updated GitHub Actions dependencies to latest versions:
  - `actions/checkout` v2 → v6.0.2
  - `actions/setup-node` v4 → v6.4.0
  - `actions/cache` v4 → v5.0.5
  - `dorny/paths-filter` v2.9.2 → v4.0.1
  - `nrwl/nx-set-shas` v4 → v5.0.1
- Replaced deprecated `octokit/request-action` with native `gh api` CLI calls for deployment management
- Fixed deprecated `::set-output` syntax → `$GITHUB_OUTPUT`
- Fixed cache key using wrong input variable (`inputs.node-version` → `env.NODE_VERSION`)

## Be aware of
- Added the `permissions` scope to limit the `gh_token` permissions in the workflow

### DIFF
--- a/.github/actions/kirby-setup/action.yml
+++ b/.github/actions/kirby-setup/action.yml
@@ -14,7 +14,7 @@ runs:
       run: jq 'del(.version)' package.json > hash-file1.json && jq 'del(.version)' package-lock.json > hash-file2.json
       shell: bash
     - name: Fetch Node Modules Cache
-      uses: actions/cache@v5
+      uses: actions/cache@v5.0.5
       id: cache
       with:
         path: | 
--- a/.github/actions/kirby-setup/action.yml
+++ b/.github/actions/kirby-setup/action.yml
@@ -7,19 +7,19 @@ runs:
       run: echo "NODE_VERSION=$(jq -r .engines.node ./package.json)" >> $GITHUB_ENV
       shell: bash
     - name: Install Node.js
-      uses: actions/setup-node@v2
+      uses: actions/setup-node@v4
       with:
         node-version: ${{env.NODE_VERSION}}
     - name: Prepare hash-files
       run: jq 'del(.version)' package.json > hash-file1.json && jq 'del(.version)' package-lock.json > hash-file2.json
       shell: bash
     - name: Fetch Node Modules Cache
-      uses: actions/cache@v4
+      uses: actions/cache@v5
       id: cache
       with:
         path: | 
           **/node_modules
-        key: node-modules-${{ runner.os }}-${{ inputs.node-version }}-${{hashFiles('**/hash-file1.json', '**/hash-file2.json')}}
+        key: node-modules-${{ runner.os }}-${{ env.NODE_VERSION }}-${{hashFiles('**/hash-file1.json', '**/hash-file2.json')}}
     - name: Remove hash-files
       run: rm hash-file1.json && rm hash-file2.json
       shell: bash
--- a/.github/actions/kirby-setup/action.yml
+++ b/.github/actions/kirby-setup/action.yml
@@ -7,14 +7,14 @@ runs:
       run: echo "NODE_VERSION=$(jq -r .engines.node ./package.json)" >> $GITHUB_ENV
       shell: bash
     - name: Install Node.js
-      uses: actions/setup-node@v6.4.0
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
       with:
         node-version: ${{env.NODE_VERSION}}
     - name: Prepare hash-files
       run: jq 'del(.version)' package.json > hash-file1.json && jq 'del(.version)' package-lock.json > hash-file2.json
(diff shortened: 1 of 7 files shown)


.github/actions/kirby-setup/action.yml
```diff
@@ -7,19 +7,19 @@ runs:
       run: echo "NODE_VERSION=$(jq -r .engines.node ./package.json)" >> $GITHUB_ENV
       shell: bash
     - name: Install Node.js
-      uses: actions/setup-node@v2
+      uses: actions/setup-node@v6.4.0
```
RasmusKjeldgaard: To be robust towards supply chain attacks, consider pinning to specific SHA's: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

RasmusKjeldgaard: So something like this: 
```suggestion
      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0

```

.github/workflows/build.yml
```diff
@@ -132,31 +136,30 @@ jobs:
...
-          mediaType: '{"previews": ["flash", "ant-man"]}' # some parameters need preview: https://docs.github.com/en/rest/reference/repos#list-deployment-statuses-preview-notices
+        run: |
+          payload=$(jq -n \
+            --arg ref "${{ github.event_name == 'pull_request' && format('refs/heads/{0}', github.head_ref) || github.ref }}" \
+            --arg environment "${{ env.GH_DEPLOY_ENVIRONMENT }}" \
+            --argjson production_environment ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v') }} \
+            '{ref: $ref, environment: $environment, production_environment: $production_environment, auto_merge: false, required_contexts: []}')
+          deployment_id=$(echo "$payload" | gh api /repos/${{ github.repository }}/deployments \
+            --method POST \
+            --input - \
```
RasmusKjeldgaard, line 160: Consider extracting the user-controlled github.head_ref to an intermediate environment variable as described [here](https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable) to mitigate script injection risk.

apps/cookbook/src/styles.scss
```diff
@@ -8,7 +8,7 @@ body,
 html {
   margin: 0;
   padding: 0;
-  background-color: var(--kirby-white);
+  background-color: #df6da9;
```
RasmusKjeldgaard: You already know, but remember to revert.

.github/actions/kirby-setup/action.yml
martinkolle: Good idea! I have updated these as well

.github/workflows/build.yml
martinkolle: That's a no brainer. Thanks. I have extracted some extra values as well. The readability is better then

.github/workflows/chromatic.yml
```diff
@@ -28,15 +31,15 @@ jobs:
...
             echo "designsystem=true" >> $GITHUB_OUTPUT
           fi
-          if [ -d ${{ github.workspace }}/dist/storybook/extensions/angular ]; then
+          if [ -d "$GITHUB_WORKSPACE/dist/storybook/extensions/angular" ]; then
             echo "extensions_angular=true" >> $GITHUB_OUTPUT
           fi
 
       - name: Publish Designsystem  to Chromatic
-        uses: chromaui/action@latest  
+        uses: chromaui/action@7aca53e7aa87489020a97f633c1e7e82c12e5973 # 1.0.0
```
RasmusKjeldgaard: Isn't this v17? Based on this: https://github.com/chromaui/action/commit/7aca53e7aa87489020a97f633c1e7e82c12e5973

.github/workflows/chromatic.yml
```diff
@@ -49,7 +52,7 @@ jobs:
       - name: Publish Extensions to Chromatic
       # If extensions storybook dist has been built, it means project might have been affected and should be tested
         if: steps.check_storybook_folders.outputs.extensions_angular == 'true'
-        uses: chromaui/action@latest
+        uses: chromaui/action@7aca53e7aa87489020a97f633c1e7e82c12e5973 # 1.0.0
```
RasmusKjeldgaard: Same question as above.

RasmusKjeldgaard (commented):
Damn nice, only one question left!

.github/workflows/chromatic.yml
martinkolle: It is true! I have updated to `17.0.0`

RasmusKjeldgaard (approved):

PR #4609, opened 2026-07-31
Modal: drawers now offset by bottom safe area inset when not having footer

### LINKED ISSUE #4579: [BUG] Modal drawers is not offset by the bottom safe-area inset
### Describe the bug and what you expected to happen

On devices with a bottom safe-area inset (e.g. iPhones with a home indicator / gesture bar or Android with the three action buttons), the content at the bottom of a modal drawer is not offset by the safe-area inset.

I expected that the modal drawer content is always completely visible on all types of physical devices.

### Describe how to reproduce the bug

Hard to reproduce. Need an app using kirby, on a phone that have have some safe area inset in the botton. Open a modal drawer and see that the bottom is cut off. 

### Severity

(3) Medium — usable with workaround

### Which Kirby version was used?

11.9.0

### Add screenshots or videos

Image

### Environment (OS / Browser)

Android / IOS

### Additional context

_No response_

Fuzzy3:

This PR closes #4579 

## What is the new behavior?

Adds the bottom safe-area inset to modal drawers and action sheets so their content and footers aren't hidden behind Android's navigation bar.

- [ ] Yes
- [x] No

## Checklist:

### Reminders

### Review  


libs/designsystem/modal/src/modal-wrapper/modal-wrapper.component.scss
```diff
@@ -157,6 +157,15 @@ ion-content {
   }
 }
 
+// On mobile, offset the bottom-most content by the safe-area inset (iOS home
+// indicator). When a footer is present it applies the inset itself, so the
+// content must not add it again:
+@include utils.media('<medium') {
+  :host(:not(.has-footer)) ion-content {
+    --padding-bottom: calc(#{utils.size('m')} + var(--kirby-safe-area-bottom, 0px));
+  }
```
RasmusKjeldgaard: Consider making this pure css with :has selector, and possibly simplifying by making the "no footer" case the baseline styling, and then adding styles for the "has footer" case.

libs/designsystem/modal/src/modal-wrapper/modal-wrapper.component.ts
```diff
@@ -534,10 +535,12 @@ export class ModalWrapperComponent
...
     if (keyboardHeight <= 0 || !element) return 0;
-    const distanceFromViewportBottomToElement = Math.floor(
-      this.windowRef.nativeWindow.innerHeight - element.getBoundingClientRect().bottom
+    const preKeyboardViewportHeight =
+      this.initialViewportHeight || this.windowRef.nativeWindow.innerHeight;
+    const elementDistanceAboveBottomOfScreen = Math.floor(
+      preKeyboardViewportHeight - element.getBoundingClientRect().bottom
     );
-    return Math.max(keyboardHeight - distanceFromViewportBottomToElement, 0);
+    return Math.max(keyboardHeight - elementDistanceAboveBottomOfScreen, 0);
```
RasmusKjeldgaard: Consider extracting this into its own PR and adding test coverage for this change.

libs/designsystem/modal/src/modal-wrapper/modal-wrapper.component.scss
```diff
@@ -157,6 +157,15 @@ ion-content {
   }
 }
 
+// On mobile, offset the bottom-most content by the safe-area inset (iOS home
+// indicator). When a footer is present it applies the inset itself, so the
+// content must not add it again:
+@include utils.media('<medium') {
+  :host(:not(:has(kirby-modal-footer))) ion-content {
+    --padding-bottom: calc(#{utils.size('m')} + var(--kirby-safe-area-bottom, 0px));
+  }
```
RasmusKjeldgaard: Is it possible to set this as default for --padding-bottom, and then override it with the proper value for when the footer is present? Would be easier to follow, I think.

libs/designsystem/modal/src/modal-wrapper/modal-wrapper.component.scss
```diff
@@ -157,6 +157,15 @@ ion-content {
   }
 }
 
+// On mobile, offset the bottom-most content by the safe-area inset (iOS home
+// indicator). When a footer is present it applies the inset itself, so the
+// content must not add it again:
+@include utils.media('<medium') {
+  :host(:not(:has(kirby-modal-footer))) ion-content {
+    --padding-bottom: calc(#{utils.size('m')} + var(--kirby-safe-area-bottom, 0px));
```
RasmusKjeldgaard: Consider using css custom property here instead of utils.size

libs/designsystem/modal/src/modal-wrapper/modal-wrapper.component.ts
Fuzzy3: Doing it !

libs/designsystem/modal/src/modal-wrapper/modal-wrapper.component.scss
Fuzzy3: Good catch ! Done now

libs/designsystem/modal/src/modal-wrapper/modal-wrapper.component.scss
RasmusKjeldgaard: Okay on second thought, why are we using a media query here? 

Safe area variables are intended to only have a value when in a context with a notch, bottom menu etc., otherwise they are 0.

So instead we can maybe do something like 

```
ion-content {
  --padding-bottom: calc(
    var(--kirby-spacing-m) + var(--kirby-safe-area-bottom)
  );
  ...
}

// A footer applies the safe-area inset itself
// so content must not add it again:
:host(:has(kirby-modal-footer)) ion-content {
  --padding-bottom: var(--kirby-spacing-m);
}

```

Fuzzy3: the media query is there to guard against tablets i believe. If we didnt have the media query we could have a centered modal on an ipad being pushed up because the ipad has a bottom home indicator. 

Maybe it might be written more clearly, but we need to have something that "toggles" between giving the footer or the content the padding pending on whether there is a footer or not in the modal.

libs/designsystem/modal/src/modal/footer/modal-footer.component.scss
```diff
@@ -19,15 +19,11 @@ ion-footer {
...
-  :host {
-    --kirby-modal-footer-safe-area-bottom: var(--kirby-safe-area-bottom);
-  }
-}
-
 :host {
   box-sizing: border-box;
   display: block;
+
+  --kirby-modal-footer-safe-area-bottom: var(--kirby-safe-area-bottom);
```
RasmusKjeldgaard: Is this still consumed anywhere? If so, could it consume the --kirby-safe-area-bottom directly?

RasmusKjeldgaard (commented):
One final question. And with a thorough manual test we should be good to go!

RasmusKjeldgaard (approved):
Approving, with a couple of nits:
1. In the changes, there are three different approaches to fallback values (0, 0px and no fallback). I think it would be nice to fix, and I don't think a fallback is necessary, as the env variables that --kirby-safe-area-bottom points to in the end should be 0px, but please verify
2. Would be good with a couple of tests to verify this behavior on both narrow and wide screens.

PR #4696, opened 2026-09-13
Tooling: Bundle dependabot package versions bump

martinkolle:
Bundle the Angular and Nx packages to support Dependabot; otherwise, builds fail due to dependency conflicts


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+    directory: /
+    schedule:
+      interval: weekly
+    groups:
+      angular:
+        patterns:
+          - '@angular/*'
+          - '@angular-devkit/*'
+          - '@angular-eslint/*'
+          - '@nx/angular'
```
RasmusKjeldgaard, line 13: I see your PR comment, but I'd still argue for something like this instead, since major versions are never really relevant for dependabot, and will create a bit of noise. 

```
 angular:
        applies-to: version-updates
        patterns:
          - "@angular*"
        update-types:
          - "minor"
          - "patch"
      minor-and-patch:
        applies-to: security-updates
        patterns:
          - "@angular*"
        update-types:
          - "patch"
          - "minor"
    ignore:
      - dependency-name: "@angular*"
        update-types: ["version-update:semver-major"]
```

RasmusKjeldgaard, line 13: Do we also need to group nx stuff? If so, should it be 
```
      - '@nx/*'
      - 'nx'
```

instead?

martinkolle: Agree. There will be pr's for next major version of angular for a long period. I have changed it

martinkolle: We should group with nx. I have added this as well

RasmusKjeldgaard (approved):

